Issue #20, opened 2026-09-21
Show the app table (level 1)


xamcross:
**Goal.** The route `/apps` shows one row for each app.

**Implementation steps.**
1. Replace the `/apps` placeholder with a component that reads `GET /api/apps` through the poll store.
2. Write one flat `<table>` with the columns app, clicks, unique users, unique sessions, status, and data time.
3. Write the name cell as `<th scope="row"><a routerLink>`, a link to level 2.
4. Show the data time as D31 states.
5. Add the states of D30.

**Acceptance criteria.**
- [x] `NEVER_POLLED` shows "–", not 0.
- [x] A failed app keeps its numbers and shows the status as an icon plus text, "Data from HH:mm:ss", and the visible `lastError`.
- [x] Without an app, the view shows "No app is registered." and a link to `/manage`.
- [x] The table has a `<caption>` and `<th scope>`.
- [x] A click on a row cell opens the link of the name cell.
- [x] The numbers use `Intl.NumberFormat`, right alignment, and `tabular-nums`.
- [x] A refresh keeps the focus and the scroll position.
- [x] No table element has a CSS `display` value, and each target is a minimum of 24 by 24 CSS px.
- [x] `@for` tracks the stable key of D29.
- [x] The view keeps `RefreshBar` of #92 between the `<h1>` and the `<table>`.
- [x] `npm test` passes in `monitor/frontend`.
- [x] A comment on this issue records the test output.

**Related.**
- #92 gives the refresh bar.
- #93 gives the main navigation and the skip link.
- #18 gives the API.
- #114 adds a link cell "First pages" to each app row of this table.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, D28, D30, D31, D33.

xamcross:
Note on 2026-09-21 from the accessibility review of #90 (https://github.com/xamcross/octometer/pull/90#issuecomment-5762487570). This issue mounts the poll store and the button "Pause refresh" in the level 1 view, thus these points belong here:

- The focus rule of D29 stops the refresh only while the DOM focus is inside a `<tbody>`. A screen reader in browse mode (NVDA) moves a virtual cursor and does not move the DOM focus, thus the rows can change while a user reads them, each 5 seconds in dev mode.
- Put the button "Pause refresh" before the `<table>` in the DOM, thus a keyboard user and a screen reader user reach it before the data.
- State the refresh interval as plain text near the button.
- Track each row with its stable key (D29 demands it already).
[...]

Issue #19, opened 2026-09-21
Build the UI shell


xamcross:
**Goal.** The Angular app has the routes, the breadcrumb, the focus rule, the first-load state, the API banner, and the status region, without table content.

**Implementation steps.**
1. Add the routes `/manage`, `/apps/:appId/users`, and `/apps/:appId/elements` of D28 next to `/apps`, each with a placeholder component.
2. Add the breadcrumb `<nav aria-label="Breadcrumb"><ol>` with `aria-current="page"`.
3. Move the focus to `<h1 tabindex="-1">` after each navigation.
4. Add the first-load state (D30).
5. Add the banner for a failed monitor API, with a reserved height (D30).
6. Add one permanent `role="status"` region for user actions only (D33).

**Acceptance criteria.**
- [x] Each of the four routes of D28 opens its placeholder component.
- [x] After a navigation the focus is on the `<h1>`.
- [x] The banner has a reserved height.
- [x] `npm test` passes in `monitor/frontend`.
- [x] A comment on this issue records the test output.

**Related.**
- #25 builds the poll store and the button "Pause refresh".

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, D28, D30, D33.

xamcross:
Body edit on 2026-09-21: #6 closed. No open blocker remains, thus the first line and the label `blocked` are removed. The work can start.

xamcross:
The pull request #91 adds the UI shell.

`npm ci`, `npm test`, `npm run lint`, `npm run format:check`, and
`npm run build` each ran in `monitor/frontend` with Node v24.21.0 and
npm 11.19.0.

`npm test` output:

```
> frontend@0.0.0 test
> ng test --watch=false

Test Files  10 passed (10)
     Tests  32 passed (32)
```

`npm run lint`: "All files pass linting."

`npm run format:check`: "All matched files use Prettier code style!"

`npm run build`: "Application bundle generation complete." The output
went to `dist/frontend`.

xamcross:
Closed on 2026-09-21 by the merge of #91 (commit c5fb5c3). Each reviewer approved:
- Accessibility specialist, second review: https://github.com/xamcross/octometer/pull/91#issuecomment-5763348702
- Angular engineer, fourth review: https://github.com/xamcross/octometer/pull/91#issuecomment-5764268843

Evidence for the criteria:
[...]

Issue #67, opened 2026-09-21
Do the NVDA pass of the UI (no cost, 1 hour)


xamcross:
**Blocked by #65.** The pass follows the NVDA script that #65 writes into `docs/a11y-check.md`. The owner can install NVDA now.

**Goal.** A screen reader user can use each view.

**Owner steps.**
1. Install NVDA.
2. Start NVDA.
3. Open `http://127.0.0.1:7431` in Chrome.
4. Do the NVDA script of `docs/a11y-check.md` for each view.
5. Write the result of each view into a comment on this issue.

**Implementation steps.**
1. File one issue with the labels `bug` and `accessibility` for each defect in the comment of the owner.

**Acceptance criteria.**
- [ ] Each defect is a new issue with the labels `bug` and `accessibility`.
- [ ] A comment on this issue records the owner's result of each view.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, D28, D33, and section 8.

Issue #65, opened 2026-09-21
Verify the UI with the keyboard, the zoom, and a narrow width


xamcross:
**Blocked by #53 and #54.** The checks need each view with data on the installed monitor. #32 and #53 add the last views, #50 and #51 serve the data, and #54 installs the monitor. Steps 1 to 3 can start now.

**Goal.** A user can operate each view with only the keyboard, at 200 percent zoom, and at a width of 320 px.

**Implementation steps.**
1. Create `docs/a11y-check.md` with a keyboard-only procedure for each view: `/apps`, `/apps/:appId/users`, `/apps/:appId/elements`, and `/manage`. The document names the data source: one registered app with events, for example the demo app of #14.
2. Add a check at 200 percent zoom and at 320 px width to the document.
3. Add a script for NVDA with Chrome to the document.
4. Run the keyboard, zoom, and width checks in the browser on `http://127.0.0.1:7431`.
5. Write the results in a comment on this issue.
6. File one new issue for each defect, with the labels `bug` and `accessibility`.

**Acceptance criteria.**
- [ ] `docs/a11y-check.md` holds the data source, the keyboard-only procedure for each view, the zoom and width check, and the NVDA script.
- [ ] Each defect is a new issue with the labels `bug` and `accessibility`.
- [ ] A comment on this issue records the results of the keyboard, zoom, and width checks.

**Related.** #67 does the NVDA pass.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, D28 to D33.

xamcross:
Body edit on 2026-09-22: #51 closed. The first line now names only the open blockers: #32, #50, #53, and #54.

xamcross:
Body edit on 2026-09-22: #50 closed. The first line now names only the open blockers: #32, #53, and #54.

xamcross:
Body edit on 2026-09-22: #32 closed. The first line now names only the open blockers: #53 and #54.

PR #90, opened 2026-09-21
feat(monitor): add the poll store of the frontend

### LINKED ISSUE #25: Build the poll store of the frontend
**Goal.** One poll store refreshes the data of each view on an interval. The user can stop the refresh (WCAG 2.2.2).

**Implementation steps.**
1. Write the poll store of D29 with `timer(0, ms)` plus `exhaustMap`, with `catchError` inside. It has the signals `data`, `lastSuccessAt`, and `error`.
2. Read the interval from `refreshSeconds` of `GET /api/health`.
3. Add the button "Pause refresh" with `aria-pressed`.
4. Stop the refresh while `document.hidden` is true.
5. Stop the refresh while the focus is in `<tbody>`.

**Acceptance criteria.** The tests use Vitest with fake timers.
- [x] The poll store keeps the last good data after a failed request and sets `error`.
- [x] The next tick does not cancel a slow request.
- [x] The refresh stops while `document.hidden` is true and runs one time when the tab shows again.
- [x] The refresh stops while the focus is in `<tbody>`.
- [x] "Pause refresh" has `aria-pressed` and stops the requests.

**Related.**
- #4 gives `refreshSeconds`.
- #19 builds the UI shell.
- #20 uses the store.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, D13, D29.

xamcross:
## Summary

This change adds the poll store of D29 to `monitor/frontend`. The store
polls with `timer(0, ms)` and `exhaustMap`, and it reads `ms` from
`refreshSeconds` of `GET /api/health`. It adds the "Pause refresh" button
with `aria-pressed` (WCAG 2.2.2). All new files stay inside
`monitor/frontend/src/app/poll/`.

Two reviews requested changes. This update fixes each finding; see
"Review round 2" and "Review round 3" below.

## Acceptance criteria

- [x] The poll store keeps the last good data after a failed request and
  sets `error`. Evidence: `poll-store.spec.ts`, test "keeps the last good
  data after a failed request and sets error".
- [x] The next tick does not cancel a slow request. Evidence:
  `poll-store.spec.ts`, test "does not cancel a slow request on the next
  tick".
- [x] The refresh stops while `document.hidden` is true and runs one time
  when the tab shows again. Evidence: `poll-store.spec.ts`, test "stops
  the refresh while the tab is hidden and runs once when it shows again".
- [x] The refresh stops while the focus is in `<tbody>`. Evidence:
  `poll-store.spec.ts`, test "stops the refresh while the focus is inside
  a tbody, and starts again on focusout".
- [x] "Pause refresh" has `aria-pressed` and stops the requests. Evidence:
  `pause-refresh-button.spec.ts` (the `aria-pressed` behavior) and
  `pause-refresh-integration.spec.ts` (the stop of the requests, with the
  button bound to a poll store).
- [x] `npm test` passes in `monitor/frontend`. Evidence: 6 test files, 31
  tests, all green (`npm test` output below).
- [x] A comment on this issue records the test output. The comment of
  2026-09-21 on issue #25 holds the full output.

## Test output

```
Test Files  6 passed (6)
     Tests  31 passed (31)
```

Full output: `npm test` (via `ng test --watch=false`), run in
`monitor/frontend`, 2026-09-21.

## Other checks

- `npm ci`: installs 497 packages, 0 vulnerabilities.
- `npm run lint`: "All files pass linting."
- `npm run format:check`: "All matched files use Prettier code style!"
- `npm run build`: builds `dist/frontend` without an error.

## Review round 2 — findings and fixes

**MAJOR (Angular engineer) — the health route read once, a fixed 60 s
fallback, no `error` on a failed health request.** Fixed. The store now
tries `GET /api/health` again at each tick until it gets a valid
`refreshSeconds`. A failed health request goes to the `error` signal, and
a `refreshSeconds` value that is not a number above 0 falls back to
60 000 ms. Evidence: `poll-store.spec.ts`, tests "tries the health route
again until it gets a valid interval" and "rejects a refreshSeconds value
that is not a number above 0, and uses the fallback".

**MAJOR (both reviewers) — `filter(canPoll)` only dropped one tick; an
unpause or a focusout did not send a request until the next full
interval.** Fixed. The gate is now a stream (`visibilitychange`,
`focusin`, `focusout`, and the `paused` signal, through `auditTime(0)`),
and it restarts a fresh `timer(0, ms)` on each open. An unpause, a focus
that leaves a `<tbody>`, or a tab that shows again now sends one request
at once, and the store never sends two requests close to one due tick.
Evidence: `poll-store.spec.ts`, tests "stops the refresh while the focus
is inside a tbody, and starts again on focusout" (the request comes
before the next tick), "sends one request at once on an unpause, before
the next tick", and "does not send two requests when the gate reopens
close to a due tick".

**MAJOR (accessibility) — the pressed state of "Pause refresh" had no
visible form.** Fixed. `pause-refresh-button.css` gives the pressed state
a border form and a check-mark icon, not colour alone, plus a visible
focus ring and a 24 by 24 CSS px target.

**MINOR (both reviewers) — `isFocusInTbody()` read a `null`
`document.activeElement` as focus inside a `<tbody>`.** Fixed:
`document.activeElement?.closest('tbody') != null`. Evidence:
`poll-store.spec.ts`, test "treats a null document.activeElement as focus
outside a tbody".

**MINOR (both reviewers) — a test name over-claimed a `focusout`
mechanism.** The gate fix above makes the name true: the request now
comes at once on a `focusout`, not on the next tick. The test keeps its
name and now checks the immediate request.

**MINOR (accessibility) — no test proved the button semantics.** Fixed.
`pause-refresh-button.spec.ts` adds checks for the native `<button
type="button">` element, the fixed accessible name "Pause refresh" in
both states, and `aria-pressed` as the text `"false"` or `"true"`.

**MINOR (Angular engineer) — no `refresh()` method, no first-load
signal.** Fixed. The store adds `refresh(): void` for one request now
(a caller uses it after it changes data on the server), and
`firstLoadPending: Signal<boolean>`, true until the store gets its first
answer, good or bad.

**MINOR (both reviewers) — two comments broke the language rules.**
Fixed: the "and"-joined instruction in `poll-store.spec.ts`, and the
metaphor "keeps a slow request alive" in `poll-store.ts`.

**MINOR (both reviewers) — the stale check box.** Fixed above.

## Review round 3 — findings and fixes

**MAJOR (Angular engineer) — `poll-store.ts:121-122`, `startWith(canPoll())`
read the gate at the pipe build, before the store subscribed.** A pause or
a hidden tab inside that window let one request pass a closed gate.
Fixed: `startWith(null)` now stands before `map(() => canPoll())`, so the
first gate value reads `canPoll()` at the subscription. Evidence:
`poll-store.spec.ts`, two new tests, "sends no request when paused
becomes true while the health request is open" and "sends no request
when document.hidden becomes true while the health request is open".
Both tests failed against the old code and pass now.

**MAJOR (accessibility) — `pause-refresh-button.css:25`, the focus ring
took its colour from the button text, and vanished in the pressed
state.** Fixed: `button:focus-visible { outline: 2px solid CanvasText; }`.
`CanvasText` contrasts with the page in both palettes.

**MAJOR (accessibility) — `pause-refresh-button.css:19-21`, the check
mark of `::before` entered the accessible name.** Fixed:
`content: '\2713\00a0' / '';`. The glyph stays visible to the eye, and
the pseudo-element gives nothing to the accessible name. The test at
`pause-refresh-button.spec.ts:57` is now named "keeps the text 'Pause
refresh' in both states", because it reads `textContent`.

**MINOR (accessibility) — the button changed size when pressed.** Fixed:
`button::before` now holds the check mark and `visibility: hidden` in
both states; `button[aria-pressed='true']::before` sets `visibility:
visible`. The button keeps one size.

**MINOR (accessibility) — the `Highlight` pair held no contrast
margin.** Fixed: the normal mode now uses project colours for the
pressed state, `#ffffff` text on a `#003366` background, a contrast of
12.6:1. `Highlight` and `HighlightText` stay inside
`@media (forced-colors: active)` only.

**MINOR (accessibility and Angular engineer) — the doc comment of
`refresh()` did not state the rule against a timer call.** Fixed: the
doc comment now reads "Call `refresh()` only from a user action. A pause
must stop each automatic update."

**MINOR (Angular engineer) — a test left `document.activeElement` as
`null` for the rest of the file.** Fixed: the shared `afterEach` in
`poll-store.spec.ts` now deletes the property.

**MINOR (Angular engineer) — the doc comment of `settleGate` held one
26-word sentence.** Fixed: the comment now holds three shorter
sentences, each at or under 20 words.

**MINOR (Angular engineer) — the idiom at `poll-store.ts:112`.** Fixed:
"so the focus change lands first" now reads "so the browser moves the
focus first".

**MINOR (Angular engineer) — no doc comment recorded that the store
sends no data poll while the health route fails.** Fixed: the doc
comment of `createPollStore` now states this behaviour, for issue #20.

## Design notes

- The store is a factory function, `createPollStore<T>(request)`. A
  caller runs it inside an injection context, for example a component
  field. This keeps the store generic: a view gives it the request
  function.
- The store tries `GET /api/health` again at each tick until it gets a
  valid interval. A failed health request sets `error`.
- The stop gate is a stream: `visibilitychange`, `focusin`, `focusout`,
  and the `paused` signal, through `auditTime(0)`, mapped to the live
  `canPoll()` state. The gate feeds a `switchMap` that runs
  `timer(0, ms)` while open and `EMPTY` while closed, so each open starts
  one request at once.
- `PauseRefreshButton` is a standalone component with one `pressed`
  model. A view binds it with `[(pressed)]="store.paused"`. This pull
  request does not do that binding; issue #20 does.

## What this pull request does not do

- It does not mount `PauseRefreshButton` or the poll store in the app
  shell. Issue #20 does that.
- It does not change `app.ts`, `app.html`, or `app.routes.ts`.
- It adds one line to `app.config.ts`: `provideHttpClient()`. The poll
  store needs `HttpClient`, and no other code in the app provides it yet.
  This was the one permitted change outside `monitor/frontend/src/app/poll/`.
- It does not add a live region. Issue #19 keeps the one region of D33.

Closes #25

Generated with Claude Code


xamcross:
**Review (accessibility): CHANGES REQUESTED**

I checked the pull request in my own clone. I ran `gh pr checkout 90`, then `npm ci` and
`npm test` in `monitor/frontend`. I read the design decisions D29, D30, and D33. A second
reviewer checks the RxJS and the signal code. This review covers the accessibility.

The button is close to correct. It has one gap that a user meets, and the focus rule of D29
protects a keyboard user but not a screen reader user in browse mode.

## Acceptance criteria

| # | Criterion | Result | Evidence |
| - | --------- | ------ | -------- |
| 1 | The store keeps the last good data after a failed request and sets `error`. | PASS | `npm test`: "keeps the last good data after a failed request and sets error". |
| 2 | The next tick does not cancel a slow request. | PASS | `npm test`: "does not cancel a slow request on the next tick". |
| 3 | The refresh stops while `document.hidden` is true. It runs one time when the tab shows again. | PASS | `npm test`: "stops the refresh while the tab is hidden and runs once when it shows again". |
| 4 | The refresh stops while the focus is in `<tbody>`. | PASS | `npm test`: "stops the refresh while the focus is inside a tbody...". See MAJOR 2 for the limit of the rule. |
| 5 | "Pause refresh" has `aria-pressed`. It stops the requests. | PASS | `pause-refresh-button.html:1` and `npm test`: the 4 `aria-pressed` tests plus the 2 integration tests. |
| 6 | `npm test` passes in `monitor/frontend`. | PASS | My run: `Test Files 6 passed (6)`, `Tests 18 passed (18)`, duration 41.00 s. |
| 7 | A comment on the issue records the test output. | PASS | The comment on #25 holds the full `npm test` output. The check box in the pull request body is out of date. |

## What is correct

The toggle button follows the ARIA Authoring Practices "Button" pattern. I confirmed each
point in my clone:

- The control is a real `<button type="button">`. It is not a `<div>` with a role.
- The accessible name is the text "Pause refresh". The name is the same in both states. No
  `aria-label` and no `aria-labelledby` change it.
- The template writes `[attr.aria-pressed]="pressed()"`. Angular renders the strings "false"
  and "true". The attribute is always present. An absent attribute would make the control a
  plain button for a screen reader.
- The visible text does not change. The component adds the state only through `aria-pressed`.

The Authoring Practices state the rule for the label:

> Toggle button: A two-state button that can be either off (not pressed) or on (pressed). To
> tell assistive technologies that a button is a toggle button, specify a value for the
> attribute aria-pressed.

> Important: it is critical the label on a toggle does not change when its state changes.

A text that changes to "Resume refresh" plus `aria-pressed` is the known defect of this
pattern. The two signals then contradict: NVDA reads "Resume refresh toggle button pressed",
and the user cannot tell which state is active. This pull request avoids the defect. Keep it
that way.

The keyboard works, because the control is a native button. The Authoring Practices state
that a user activates a button with the Space key or the Enter key. The component adds no key
handler that replaces the native behaviour.

On the live regions the pull request is correct. The component template and the store add no
`aria-live`, no `role="status"`, no `role="alert"`, and no `role="log"`. I ran
`grep -rn 'aria-live|role="status"|role="alert"' monitor/frontend/src` and it found nothing.
D33 keeps one permanent region in the shell (#19). A refresh each 5 seconds in dev mode must
never announce itself. This pull request holds that line.

The diff holds 7 source files only. It holds no secret, no database file, and no build output.
The language of each comment and each commit body follows ASD-STE100.

## Findings

### MAJOR 1 — the pressed state has no visible form

`monitor/frontend/src/app/poll/pause-refresh-button.ts:7-10`, and
`monitor/frontend/src/app/poll/pause-refresh-button.html:1`

The component has no `styleUrl`. `monitor/frontend/src/styles.css` holds one comment and no
rule. Thus the pressed state reaches a screen reader through `aria-pressed`, and it reaches
nobody else. A sighted user presses "Pause refresh", the label stays the same, and the button
looks the same. The only other feedback is the stop of the table updates. In prod mode the
interval is 60 seconds, so that feedback takes up to 60 seconds. The user presses the button
a second time, and starts the refresh again by mistake.

This is not a failure of WCAG 1.4.1 alone, because no colour carries the state today. It is
the state with no visual form at all. When a style arrives later, 1.4.1 applies: the pressed
state needs a second cue next to the colour.

No issue owns this today. #20 holds the criterion "each target is a minimum of 24 by 24 CSS
px", and it holds no criterion for the visible state. #65 and #67 are verification issues.
They would find the gap late.

**The fix.** Take one of the two:

1. Add `pause-refresh-button.css` to the component with a rule for the pressed state. Carry
   the state with two cues, not with a colour alone. An example:

   ```css
   button {
     min-block-size: 24px;
     min-inline-size: 24px;
   }
   button[aria-pressed='true'] {
     background: Highlight;
     color: HighlightText;
     border-style: inset;
   }
   button[aria-pressed='true']::before {
     content: '\2713\00a0'; /* a check mark, a cue next to the colour */
   }
   ```

2. Add an acceptance criterion to #20: "The pressed state of the Pause refresh button is
   visible without a colour alone, and the target is a minimum of 24 by 24 CSS px." Then
   record the criterion, so no later step loses it.

### MAJOR 2 — the focus rule of D29 does not protect a screen reader user (owner: #20, not a blocker for this pull request)

`monitor/frontend/src/app/poll/poll-store.ts:37-40` and `:66`

The rule reads `document.activeElement?.closest('tbody')`. It stops the refresh only while
the DOM focus sits inside a table body.

- A keyboard user is safe. The user tabs to a link in a row, the DOM focus moves into the
  `<tbody>`, and the refresh stops. This works.
- A screen reader user in browse mode is not safe. NVDA in Chrome moves a virtual cursor
  through the virtual buffer. An arrow key does not move the DOM focus. `document.activeElement`
  stays on `<body>` or on the last focused element. `closest('tbody')` returns null, the filter
  passes, and the store replaces the rows while the user reads them. NVDA rebuilds the changed
  part of the buffer. The reading position can move, or the text can change with no notice.
  In dev mode this happens each 5 seconds.
- A low-vision user with a magnifier and a mouse user have the same problem. Neither one moves
  the DOM focus while they read.

So the rule is necessary and it is not sufficient. The pull request follows D29 exactly, thus
this is a gap in the design and not a defect of this code. The fix belongs to the issues that
mount the button and the table.

**The fix, for #20 (with the shell of #19).** Add these criteria:

1. The "Pause refresh" button stands before the `<table>` in the DOM order, in the same
   region. A browse-mode user then meets the control before the data, and can stop the
   refresh with one action.
2. A short text before the table states the rule, for example "The table refreshes each 60
   seconds." A user then knows why the content moves.
3. `@for` tracks the stable key of D29. A refresh then changes only the changed cells, and
   NVDA rebuilds a small part of the buffer.
4. The paused state holds across a navigation between the three levels. A user who needs the
   pause needs it on each view.

Point 1 is the important one. Do not rely on the focus rule alone.

### MINOR 1 — `isFocusInTbody` returns true when `document.activeElement` is null

`monitor/frontend/src/app/poll/poll-store.ts:38-40`

```ts
return document.activeElement?.closest('tbody') !== null;
```

`?.` gives `undefined` when `activeElement` is null. `undefined !== null` is true. The store
then reads "the focus is in a table body", and it stops every request.

I proved this in my clone with a temporary test: I set `document.activeElement` to null, ran
the store for 30 seconds of fake time, and the request count stayed at 0. The store made no
request at all.

Chrome keeps `activeElement` on `<body>` in a normal page, so the case is rare. The code
still says the opposite of the intent, and the failure is silent.

**The fix.** Use `!= null` or a boolean cast:

```ts
return document.activeElement?.closest('tbody') != null;
```

### MINOR 2 — a test name claims a `focusout` mechanism that the code does not hold

`monitor/frontend/src/app/poll/poll-store.spec.ts:134`

The name is "stops the refresh while the focus is inside a tbody, and starts again on
focusout". The store holds no `focusout` listener. It re-reads the live DOM state on the next
`timer` tick. The test passes because it advances 10 seconds after `button.blur()`.

The behaviour agrees with D29. The name over-claims the mechanism. A later reader can look
for a listener that does not exist.

**The fix.** Rename the test to "stops the refresh while the focus is inside a tbody, and
starts again on the next tick after the focus leaves it".

### MINOR 3 — no test proves the keyboard use or the button semantics

`monitor/frontend/src/app/poll/pause-refresh-button.spec.ts:29-41`

Each state test uses `button().click()`. A `click()` call passes for a `<div>` with a click
handler too. No test proves that the control is a native button, that it sits in the tab
order, or that the accessible name stays the same.

I wrote the missing tests in my clone and ran them. **The result: 7 test files, 26 tests, all
green.** The 6 new checks pass against the current code, so they add cover and they change
nothing.

One point on the test environment: `ng test` runs on jsdom. jsdom runs no default activation
for a key on a native button. I proved it with a probe: I dispatched Enter and Space on a
plain `<button>`, and the click count stayed at 0; a `click()` call raised it to 1. So a test
cannot press Enter here. The correct check is the semantics that give Enter and Space in a
real browser. #67 covers the real NVDA pass.

Add this file as `monitor/frontend/src/app/poll/pause-refresh-a11y.spec.ts`:

```ts
import { Component, signal } from '@angular/core';
import { ComponentFixture, TestBed } from '@angular/core/testing';

import { PauseRefreshButton } from './pause-refresh-button';

/** A host that records each pressedChange event, the way the shell does. */
@Component({
  selector: 'app-a11y-test-host',
  imports: [PauseRefreshButton],
  template: `<app-pause-refresh-button
    [pressed]="paused()"
    (pressedChange)="onPressed($event)"
  />`,
})
class A11yTestHost {
  readonly paused = signal(false);
  readonly announcements: string[] = [];

  onPressed(value: boolean): void {
    this.paused.set(value);
    this.announcements.push(value ? 'Refresh paused' : 'Refresh on');
  }
}

describe('PauseRefreshButton accessibility', () => {
  let fixture: ComponentFixture<PauseRefreshButton>;

  beforeEach(async () => {
    await TestBed.configureTestingModule({
      imports: [PauseRefreshButton],
    }).compileComponents();

    fixture = TestBed.createComponent(PauseRefreshButton);
    document.body.appendChild(fixture.nativeElement);
    fixture.detectChanges();
  });

  afterEach(() => {
    fixture.nativeElement.remove();
  });

  function button(): HTMLButtonElement {
    return fixture.nativeElement.querySelector('button');
  }

  it('is a native button element with type="button" and no role override', () => {
    expect(button().tagName).toBe('BUTTON');
    expect(button().type).toBe('button');
    expect(button().getAttribute('role')).toBeNull();
  });

  it('is in the tab order and takes the DOM focus', () => {
    button().focus();
    expect(document.activeElement).toBe(button());
    expect(button().tabIndex).toBe(0);
    expect(button().hasAttribute('disabled')).toBe(false);
    expect(button().getAttribute('aria-disabled')).toBeNull();
  });

  it('keeps the accessible name "Pause refresh" in both states', () => {
    expect(button().textContent?.trim()).toBe('Pause refresh');
    expect(button().getAttribute('aria-label')).toBeNull();
    expect(button().getAttribute('aria-labelledby')).toBeNull();
    button().click();
    fixture.detectChanges();
    expect(button().getAttribute('aria-pressed')).toBe('true');
    expect(button().textContent?.trim()).toBe('Pause refresh');
  });

  it('writes aria-pressed as the text "false" and "true", never an absent attribute', () => {
    expect(button().hasAttribute('aria-pressed')).toBe(true);
    expect(button().getAttribute('aria-pressed')).toBe('false');
    button().click();
    fixture.detectChanges();
    expect(button().hasAttribute('aria-pressed')).toBe(true);
    expect(button().getAttribute('aria-pressed')).toBe('true');
  });

  it('adds no live region of its own (D33 keeps one region in the shell)', () => {
    const host: HTMLElement = fixture.nativeElement;
    expect(host.querySelector('[aria-live]')).toBeNull();
    expect(host.querySelector('[role="status"]')).toBeNull();
    expect(host.querySelector('[role="alert"]')).toBeNull();
    expect(host.querySelector('[role="log"]')).toBeNull();
  });

  // jsdom runs no default activation for a key on a native button. The native
  // element gives Enter and Space in a real browser. This test proves that the
  // component adds no key handler that replaces the native behaviour.
  it('adds no key handler that replaces the native Enter and Space behaviour', () => {
    button().focus();
    button().dispatchEvent(new KeyboardEvent('keydown', { key: 'a', bubbles: true }));
    button().dispatchEvent(new KeyboardEvent('keydown', { key: 'Enter', bubbles: true }));
    fixture.detectChanges();
    expect(button().getAttribute('aria-pressed')).toBe('false');
    button().click();
    fixture.detectChanges();
    expect(button().getAttribute('aria-pressed')).toBe('true');
  });
});

describe('PauseRefreshButton as a hook for the status region', () => {
  it('emits pressedChange, so the shell can announce the user action', async () => {
    await TestBed.configureTestingModule({ imports: [A11yTestHost] }).compileComponents();
    const host = TestBed.createComponent(A11yTestHost);
    host.detectChanges();
    const element: HTMLButtonElement = host.nativeElement.querySelector('button');

    element.click();
    host.detectChanges();
    element.click();
    host.detectChanges();

    expect(host.componentInstance.announcements).toEqual(['Refresh paused', 'Refresh on']);
    expect(element.getAttribute('aria-pressed')).toBe('false');
  });
});
```

### MINOR 4 — a passive sentence in a comment

`monitor/frontend/src/app/poll/pause-refresh-button.ts:12`

"True while the refresh is paused." uses the passive voice. Write "True while the user pauses
the refresh." The same file uses the active voice elsewhere, and `poll-store.ts:30` is correct.

### MINOR 5 — the pull request body holds a stale check box

The body marks the criterion "A comment on this issue records the test output" as open. The
comment on #25 holds the full output. Tick the box.

## Answers to the two open questions

**Does the status region announce "Refresh paused" and "Refresh on"?** Yes, it may. Both texts
follow a button press, thus they report a user action. D33 allows exactly that. Two points:

1. NVDA already reads the state change of the button. A press gives "Pause refresh toggle
   button pressed". A status text then gives a second announcement of the same fact. Prefer a
   text that states the consequence, for example "Refresh paused. The table does not update."
   and "Refresh on. The table updates each 60 seconds."
2. The pair "Refresh paused" and "Refresh on" mixes two word sets. Use one set, for example
   "paused" and "resumed".

**Does the component give a hook, with no region of its own?** Yes. `model(false)` creates the
input `pressed` and the output `pressedChange`. The shell binds it like this:

```html
<app-pause-refresh-button [pressed]="store.paused()" (pressedChange)="onPause($event)" />
```

`onPause` writes `store.paused` and writes the text of the shell status region. I proved this
path with the last test above. The component holds no region, so #19 keeps the one region of
D33.

## Notes, not findings

- **The focus indicator (WCAG 2.4.7, Level AA).** The component has no style, so the Chrome
  default focus ring applies. That ring is visible, and 2.4.7 passes today. A later style must
  keep a visible ring.
- **A correction on the criterion numbers.** WCAG 2.2 SC 2.4.11 is "Focus Not Obscured
  (Minimum)" (Level AA), and not the focus appearance. The focus appearance is SC 2.4.13
  (Level AAA). 2.4.11 applies to #19 and #20: a sticky header must not cover the focused
  button.
- **The target size (WCAG 2.5.8, Level AA).** The component has no style. A Chrome default
  button is about 21 CSS px tall, thus it is below the 24 px of 2.5.8 and of D33. #20 holds
  the criterion "each target is a minimum of 24 by 24 CSS px", and #20 mounts this button, so
  the owner exists. MAJOR 1 gives the rule to use.
- **WCAG 2.2.2 (Level A).** The criterion reads: "For any auto-updating information that
  (1) starts automatically and (2) is presented in parallel with other content, there is a
  mechanism for the user to pause, stop, or hide it or to control the frequency of the update
  unless the auto-updating is part of an activity where it is essential." This pull request
  builds the mechanism. The criterion itself is met only when #20 puts the button and the
  table on one page. The scope of this pull request is correct.
- **The choice of `aria-pressed`.** Some practitioners prefer a label change from "Pause" to
  "Resume" with no `aria-pressed` for a pause control. D29 fixes the choice on `aria-pressed`,
  and the code follows D29. Keep one pattern across the app.

## Commands that I ran

```
gh pr checkout 90                       # in C:/Users/xamcr/octo-wt/review-90-a11y
npm ci                                  # in monitor/frontend, 0 vulnerabilities
npm test                                # Test Files 6 passed (6), Tests 18 passed (18)
npm test  (plus my 6 new a11y checks)   # Test Files 7 passed (7), Tests 26 passed (26)
```

I deleted my clone after the review.

xamcross:
**Review (store): CHANGES REQUESTED**

I checked the store only. A second reviewer checks the accessibility of the
button and the stop rules. I used my own clone, `gh pr checkout 90`, and my
own tests. I did not trust the pull request description.

## Acceptance criteria

- **PASS** — The poll store keeps the last good data after a failed request
  and sets `error`. Evidence: `poll-store.ts:77-93`, plus my test R1. After
  the failed second request, `data()` stays `"v1"` and `error()` is an
  `Error`. The tick at 20 s sends a third request, and that success clears
  `error()`.
- **PASS** — The next tick does not cancel a slow request. Evidence: my test
  R2 with `HttpTestingController`. After two ticks, `first.cancelled` is
  `false`, and `expectNone('/api/apps')` proves that no second request runs
  in parallel.
- **PASS** — The refresh stops while `document.hidden` is true, and it runs
  one time when the tab shows again. Evidence: `poll-store.ts:62-66`, plus
  `poll-store.spec.ts:113-132`. See MINOR 2 for a near tick.
- **PASS** — The refresh stops while the focus is in `<tbody>`. Evidence:
  `poll-store.ts:38-40,66`. The start again is late; see MAJOR 2.
- **PASS** — "Pause refresh" has `aria-pressed`, and it stops the requests.
  Evidence: `pause-refresh-button.html:1` and
  `pause-refresh-integration.spec.ts:49-60`.
- **PASS** — `npm test` passes in `monitor/frontend`. My run:
  `Test Files 6 passed (6)`, `Tests 18 passed (18)`.
- **PASS** — A comment on issue #25 records the test output. The comment of
  2026-09-21 holds the full output. The pull request body leaves this box
  unchecked. Please check it.

## Agreement with D29 and D30

| Rule of the design | Result | Evidence |
| --- | --- | --- |
| `timer(0, ms)` plus `exhaustMap` | PASS | `poll-store.ts:75-77` |
| `catchError` inside the inner stream | PASS | `poll-store.ts:80` |
| Signals `data`, `lastSuccessAt`, `error` | PASS | `poll-store.ts:57-59` |
| No `httpResource` | PASS | The store uses `HttpClient`. |
| The state in signals, not in a `BehaviorSubject` | PASS | `poll-store.ts:57-60` |
| The interval from `refreshSeconds` of `GET /api/health` | PASS | `poll-store.ts:68-71`, my test R8 |
| The store reads the health route first | PASS | My test R8: the data request starts after the health answer only. |
| A refresh keeps the old rows | PASS | A failure writes `error` only. |

`lastSuccessAt` changes on a success only. My test R3 proves it: after the
failed request the `Date` object stays the same object.

## Life cycle

- **PASS** — The subscription ends on the destroy of the owner.
  `takeUntilDestroyed(destroyRef)` is the last operator, thus the unsubscribe
  reaches the `timer` and the `fromEvent`. My test R5 destroys the host
  component, advances the clock by 60 s, and sends a `visibilitychange`. The
  call count stays at 1.
- **PASS** — The `visibilitychange` listener goes away. My test R5 counts the
  calls of `document.addEventListener` and `document.removeEventListener`:
  one add, then one remove on the destroy.
- **PASS** — Two stores keep two separate states. My test R4: two stores, two
  health requests, separate `data()` and separate `paused()`.
- **NOTE** — There is no `focusin` listener and no `focusout` listener. The
  store reads `document.activeElement` at the tick. Thus nothing leaks, but
  see MAJOR 2.

## Interplay of the stop rules

- **PASS** — Paused plus hidden, then visible again while still paused, sends
  no request. My test R6: the count stays at 1 through the `visibilitychange`
  and through 30 s more. The count moves to 2 after `paused.set(false)` plus
  one tick.

## Zoneless change detection

- **PASS** — My test R11 mounts a component, flushes the health answer, and
  flushes the data answer. The text of the view changes to the new value. The
  test calls no `detectChanges`. The signal write from the RxJS callback
  drives the view.

## Scope

- **PASS** — Each new file is in `monitor/frontend/src/app/poll/`. The one
  change outside is `provideHttpClient()` in `app.config.ts:2,10`.
  `app.ts`, `app.html`, and `app.routes.ts` are not in the diff.
- **PASS** — No secret, no database file, and no build output is in the diff.

## Findings

### MAJOR 1 — `monitor/frontend/src/app/poll/poll-store.ts:68-71`

The store reads `/api/health` one time. A failed health request gives a fixed
60 000 ms, and the store never reads the health route again. My test R7
proves it: after the 503, the first poll comes at 60 s, and 10 more minutes
of ticks produce no second health request.

Two defects follow:
1. Line 26 of the design gives 5 seconds in dev mode. A developer who opens
   the view before the monitor server starts gets 60 seconds, twelve times
   the design value, until a page reload.
2. `catchError` swallows the health error. No signal holds it. Thus a view
   cannot show the "failed monitor API" banner of D30 at the start.

Fix: keep the fallback, and add a retry.

```ts
const intervalMs$ = http.get<HealthResponse>('/api/health').pipe(
  map((health) => health.refreshSeconds * 1000),
  tap(() => healthError.set(undefined)),
  retry({ delay: (err) => { healthError.set(err); return timer(HEALTH_RETRY_MS); } }),
  startWith(FALLBACK_INTERVAL_MS),
  distinctUntilChanged(),
);
```

`switchMap` then restarts the timer with the correct interval as soon as the
health route answers. Expose `healthError` as a signal.

### MAJOR 2 — `poll-store.ts:66,75-76` and `poll-store.spec.ts:134,151-153`

`filter(() => canPoll())` drops one tick. It does not gate a stream. Thus the
refresh does not start on a `focusout`. It waits for the next tick, which is
60 seconds in prod mode. My test R12 with a 60-second interval proves it:
after `button.blur()` the call count stays at 1 for 59 999 ms, and it moves to
2 at 60 000 ms.

The same delay applies to the pause button. The user presses "Pause refresh" a
second time, and the numbers stay old for up to 60 seconds. My test R12 proves
this part too.

The test `poll-store.spec.ts:134` has the name "stops the refresh while the
focus is inside a tbody, and starts again on focusout". The code has no
`focusout` behaviour. The assertion at line 152 advances the clock by a full
interval first, thus it proves the next tick and not the `focusout`. The name
is wrong.

Fix: make the gate a stream, and restart the timer when the gate opens.

```ts
const gateChanges$ = merge(
  fromEvent(document, 'visibilitychange'),
  fromEvent(document, 'focusin'),
  fromEvent(document, 'focusout'),
  toObservable(paused),
).pipe(auditTime(0));
```

`auditTime(0)` lets the browser move the focus first, thus `activeElement` is
correct. Then `switchMap` from `gateChanges$` to `timer(0, ms)` while
`canPoll()` is true, and to `EMPTY` while it is false. This gives an immediate
poll on a resume, and it also fixes MINOR 2. As a minimum, rename the test to
"starts again on the next tick after a focusout".

### MINOR 1 — `poll-store.ts:39`

```ts
return document.activeElement?.closest('tbody') !== null;
```

A `null` `document.activeElement` makes the optional chain give `undefined`,
and `undefined !== null` is `true`. The store then stops the refresh although
no focus is in a `<tbody>`. My test R9 proves the wrong branch: with a `null`
`activeElement` the store makes 0 requests; after I restore the property it
makes 1.

Fix: `return document.activeElement?.closest('tbody') != null;`

### MINOR 2 — `poll-store.ts:62-64,75`

The `visibilitychange` trigger adds one request, but it does not reset the
timer. My test R13 shows the tab with 10 ms left to a due tick: the store
sends two requests within 10 ms. The fix of MAJOR 2 removes this.

### MINOR 3 — `poll-store.ts:69`

The store does not check `refreshSeconds`. A value of 0 gives `timer(0, 0)`,
and the store then polls as fast as the server answers.

Fix: `map((health) => Math.max(1000, (health.refreshSeconds ?? 0) * 1000))`

### MINOR 4 — `poll-store.ts:23-32` (the API for #20)

The generic API is good for a list view. A caller writes one line in a
component field:

```ts
readonly store = createPollStore(() => this.http.get<AppRow[]>('/api/apps'));
```

`pause-refresh-integration.spec.ts:20-23` proves that this works in a
component. Two gaps remain for #20 and for D32:
1. There is no `refresh()` method. The management page adds, changes, and
   deletes an app, and it needs the new numbers at once.
2. There is no signal that tells "no poll yet" from "a poll runs now". D30
   asks for the `NEVER_POLLED` state and for a first-load state.

Fix: return `refresh(): void` that pushes into a `Subject`, and merge that
`Subject` with the timer.

### MINOR 5 — the language of two comments

- `poll-store.spec.ts:9-10`: "Builds a tbody with one focusable button, and
  adds it to the document." joins two instructions with "and". Rule 4 asks
  for one instruction in one sentence.
- `poll-store.ts:48`: "keeps a slow request alive" is a metaphor. Write
  "`exhaustMap` does not cancel a slow request."

## What I ran

In `monitor/frontend` of my own clone, on the branch `issue-25-poll-store`:

- `npm ci` — 0 vulnerabilities.
- `npm test` — `Test Files 6 passed (6)`, `Tests 18 passed (18)`.
- `npm run lint` — "All files pass linting."
- `npm run format:check` — "All matched files use Prettier code style!"
- `npm run build` — builds `dist/frontend` without an error.
- My own test file, 12 tests, all green. It uses Vitest fake timers and
  `HttpTestingController`. It uses no `fakeAsync`.

xamcross:
Thank you both. This comment lists each finding and its fix. The pull
request text has the full detail in "Review round 2".

## MAJOR

1. **Health route read once (Angular engineer).** Fixed. The store tries
   `GET /api/health` again at each tick until it gets a valid
   `refreshSeconds`. A failed health request now sets `error`. A
   `refreshSeconds` value that is not a number above 0 falls back to
   60 000 ms. `poll-store.ts`, `intervalMs$`.
2. **The gate only dropped a tick; no request at once on a reopen (both
   reviewers).** Fixed. The gate is now a stream over
   `visibilitychange`, `focusin`, `focusout`, and `paused`, through
   `auditTime(0)`. It restarts a fresh `timer(0, ms)` on each open, so an
   unpause, a focusout of a `<tbody>`, or a visible tab sends one request
   at once, with no double request near a due tick. `poll-store.ts`,
   `gateOpen$` and `scheduledTicks$`.
3. **The pressed state had no visible form (accessibility).** Fixed.
   `pause-refresh-button.css` adds a border form and a check-mark icon
   for the pressed state, plus a visible focus ring and a 24 by 24 CSS
   px target.

## MINOR

- **`isFocusInTbody()` treated a `null` `activeElement` as focus inside a
  `<tbody>` (both reviewers).** Fixed:
  `document.activeElement?.closest('tbody') != null`. New test: "treats
  a null document.activeElement as focus outside a tbody".
- **`refresh()` and a first-load signal (Angular engineer).** Added.
  `refresh(): void` sends one request now. `firstLoadPending` is true
  until the store gets its first answer, good or bad.
- **Test semantics of the button (accessibility).** Added: the native
  `<button type="button">` element, the fixed accessible name in both
  states, and `aria-pressed` as the text `"false"` or `"true"`.
- **The stale test name for the focusout case (both reviewers).** The
  gate fix makes the name true: the request now comes at once, so the
  test keeps its name and checks the immediate request instead of the
  next tick.
- **Two comments broke the language rules (both reviewers).** Fixed.
- **The stale check box in the pull request body (both reviewers).**
  Fixed.

## Verification

`npm ci`, `npm test` (6 files, 29 tests), `npm run lint`, `npm run
format:check`, and `npm run build` all pass in `monitor/frontend`. The
CI checks of the pull request are green.

xamcross:
**Review (second review, accessibility specialist): CHANGES REQUESTED**

I made a new clone and I checked out the head of `issue-25-poll-store` (`d9ecbef`). I checked
each finding against the code, and not against the response comment. I also measured the new
stylesheet in a real Chrome on Windows.

The store fixes are correct. The new stylesheet fixes MAJOR 1 of my first review, and it adds
two new defects. Both sit in the 4 new lines of CSS. Both have a short fix.

## The five commands

| Command | Result |
| ------- | ------ |
| `npm ci` | 0 vulnerabilities. |
| `npm test` | `Test Files 6 passed (6)`, `Tests 29 passed (29)`. |
| `npm run lint` | "All files pass linting." |
| `npm run format:check` | "All matched files use Prettier code style!" |
| `npm run build` | Builds `dist/frontend`. No error. |

| # | Finding | Status |
| - | ------- | ------ |
| MAJOR 1 | The pressed state had no visible form. | **CORRECTED**, and the correction adds NEW MAJOR 1 and NEW MAJOR 2 below. `pause-refresh-button.css` now gives a border form, a check mark, a focus ring, and a 24 by 24 px target. I measured the target: 108.34 by 24.67 px unpressed, 122.04 by 27.33 px pressed. WCAG 2.5.8 passes. |
| MAJOR 2 | The focus rule does not protect a screen reader user in browse mode. Owner #20. | **NOT CORRECTED.** #20 still holds no criterion for the order of the button and the table. See "The open action" below. This does not block this pull request. |
| MINOR 1 | `isFocusInTbody()` read a null `activeElement` as focus in a `<tbody>`. | **CORRECTED.** `poll-store.ts:57` now uses `!= null`. The test "treats a null document.activeElement as focus outside a tbody" (`poll-store.spec.ts:274`) proves it. |
| MINOR 2 | A test name claimed a `focusout` mechanism that the code did not hold. | **CORRECTED.** `poll-store.ts:114-134` now listens to `focusout` and restarts a fresh `timer(0, ms)`. The test at `poll-store.spec.ts:192` checks the request at once, with no full interval. The name is true now. |
| MINOR 3 | No test proved the button semantics or the keyboard. | **CORRECTED** for the three key checks (`pause-refresh-button.spec.ts:51,57,68`). Note: the accessible-name test is now false. NEW MAJOR 2 explains this. |
| MINOR 4 | A passive sentence in a comment. | **CORRECTED.** `pause-refresh-button.ts:13` reads "True while the user pauses the refresh." |
| MINOR 5 | A stale check box in the pull request body. | **CORRECTED.** Each box is ticked. |

Point 4 of the brief holds. I ran a search over `monitor/frontend/src` outside `src/app/poll/`.
It found no `aria-live`, no `role="status"`, no `role="alert"`, no `role="log"`, and no mount
of the button or the store. The diff holds 8 source files. It holds no secret, no database
file, and no build output. Each new commit body follows ASD-STE100.

## NEW MAJOR 1 — the focus ring disappears in the pressed state

`monitor/frontend/src/app/poll/pause-refresh-button.css:25`

```css
button:focus-visible {
  outline: 2px solid currentColor;
  outline-offset: 2px;
}
```

`outline-offset: 2px` draws the ring outside the border box. The ring sits on the page
background, and not on the button. `currentColor` is the `color` of the button. Line 16 sets
`color: HighlightText` for the pressed state. So the ring takes the colour of the text of a
selected item, and that colour must contrast with `Highlight`, not with the page.

I measured this in Chrome 141 on Windows 11, light palette, with the exact rules of the file:

| State | `outline-color` | Page background | Contrast |
| ----- | --------------- | --------------- | -------- |
| Not pressed | `rgb(0, 0, 0)` | `rgb(255, 255, 255)` | 21.00 : 1 — pass |
| **Pressed** | **`rgb(255, 255, 255)`** | `rgb(255, 255, 255)` | **1.00 : 1 — fail** |

The ring is white on a white page. It is invisible. This fails WCAG 1.4.11 Non-text Contrast
(Level AA, a minimum of 3:1) and WCAG 2.4.7 Focus Visible (Level A).

The state matters. A keyboard user presses the button, and the button stays under the focus.
So the focus vanishes at the exact moment that the user needs it. A keyboard user then loses
the position next to the table.

**The fix.** Do not take the ring colour from the text of the button. Use a colour that
contrasts with the page:

```css
button:focus-visible {
  outline: 2px solid CanvasText;
  outline-offset: 2px;
}
```

I measured `CanvasText` on the same machine with no `color-scheme` set: `rgb(0, 0, 0)`, thus
21:1 against the page. `Canvas` gives `rgb(255, 255, 255)`, so the pair holds in a dark
palette too. A two-colour ring is an alternative, and it works on any background.

## NEW MAJOR 2 — the check mark changes the accessible name

`monitor/frontend/src/app/poll/pause-refresh-button.css:19-21`

```css
button[aria-pressed='true']::before {
  content: '\2713\00a0';
}
```

CSS generated content enters the accessible name. The Accessible Name and Description
Computation states the rule for step 2F.ii:

> For `::before` pseudo elements, User agents MUST prepend CSS textual content, without a
> space, to the textual content of the current node.

I read the computed value in Chrome: `"✓ "` on the pressed button, and `none` on the
unpressed button. So the name is "Pause refresh" in one state and "✓ Pause refresh" in the
other state. NVDA reads the check mark, or the word for it, before the label.

This is the defect that my first review named, and it comes back through the CSS. The
Authoring Practices rule that I quoted still applies:

> Important: it is critical the label on a toggle does not change when its state changes.

The new test does not catch it. `pause-refresh-button.spec.ts:57-66` is named "keeps the
accessible name 'Pause refresh' in both states", and it reads `textContent`. `textContent`
never holds generated content. I proved this in Chrome: a button with only a `::before` of
"GENERATEDNAME" has a `textContent` of `""`. So the test passes and it cannot fail. It is
green for a reason that is not the reason in its name.

**The fix.** Use the alternative-text syntax of CSS generated content. The glyph stays
visible, and the pseudo-element gives nothing to the name:

```css
button[aria-pressed='true']::before {
  content: '\2713\00a0' / '';
}
```

I checked the support on the same Chrome. It parses the value as `"✓ " / ""`. Firefox 122 and
Safari 17.4 support it too.

Then rename the test to "keeps the text 'Pause refresh' in both states", because `textContent`
is what it reads. #67 covers the real name with NVDA.

## MINOR 1 — the button changes size when the user presses it

`monitor/frontend/src/app/poll/pause-refresh-button.css:19-21`

The check mark enters the layout, so the button grows. I measured 108.34 by 24.67 px
unpressed, and 122.04 by 27.33 px pressed. That is 13.70 px wider and 2.67 px taller. Each
press moves the content next to the button. A user with a magnifier loses the position.

**The fix.** Give the pseudo-element to both states, and switch the visibility:

```css
button::before {
  content: '\2713\00a0' / '';
  visibility: hidden;
}
button[aria-pressed='true']::before {
  visibility: visible;
}
```

I measured this form: 122.04 px wide in both states.

## MINOR 2 — the `Highlight` pair holds no contrast margin

`monitor/frontend/src/app/poll/pause-refresh-button.css:15-16`

I measured `HighlightText` on `Highlight`: `rgb(255, 255, 255)` on `rgb(0, 120, 212)`, which
is 4.53:1. WCAG 1.4.3 Contrast (Minimum) needs 4.5:1 for this text size. The margin is 0.03.

On Windows, Chrome takes `Highlight` from the accent colour of the user. The user can change
that colour, and the pair can then drop below 4.5:1. The project does not control it.

**The fix.** Use project colours for the normal mode. Keep the system colours for the forced
mode:

```css
@media (forced-colors: active) {
  button[aria-pressed='true'] {
    background: Highlight;
    color: HighlightText;
  }
}
```

## MINOR 3 — `refresh()` ignores the pause

`monitor/frontend/src/app/poll/poll-store.ts:41,136,163`

`refresh()` sends one request through `manualRefresh$`, and that stream misses the gate. The
test at `poll-store.spec.ts:287` states this: "refresh() sends one request now, and it does
not wait for a paused store".

A user action may update the data while the store is paused. WCAG 2.2.2 covers automatic
updates only, so a user action is correct. A timer or a route event is not.

**The fix.** Add one sentence to the doc comment at `poll-store.ts:40`: "Call `refresh()` only
from a user action. A pause must stop each automatic update." Then #20 has the rule in front
of it.

## The open action for the maintainer (MAJOR 2 of the first review)

#20 still holds no criterion from my first review. The focus rule of D29 stops the refresh
only while the DOM focus sits in a `<tbody>`. NVDA in browse mode moves a virtual cursor, and
that cursor does not move the DOM focus. So the rows change while a screen reader user reads
them, each 5 seconds in dev mode. Add these four criteria to #20 before #20 starts:

1. The "Pause refresh" button stands before the `<table>` in the DOM order, in the same region.
2. A short text before the table states the interval.
3. `@for` tracks the stable key of D29.
4. The paused state holds across a navigation between the three levels.

## What is correct

- The store fixes are sound. The gate now reacts to `focusout`, and it sends one request at
  once. The test at `poll-store.spec.ts:248` proves that a reopen near a due tick sends one
  request, and not two.
- The button stays a native `<button type="button">` with no role override.
- `aria-pressed` stays as the text "false" or "true". The attribute is always present.
- The visible label "Pause refresh" does not change.
- The forced-colours mode works. The file uses the system colour keywords, so the pressed pair
  survives there. Chrome forces `outline-color` in that mode, so NEW MAJOR 1 does not appear
  there. The failure is in the normal mode only. Note that `border-style: inset` renders flat
  in the forced mode, and the check mark then carries the state alone.
- The component still holds no live region. #19 keeps the one region of D33.

## How I measured

I copied the 4 rules of `pause-refresh-button.css` into a page in Chrome 141 on Windows 11. I
read `getComputedStyle` for each colour and each box, and I computed each contrast with the
WCAG relative-luminance formula. The OS sits in the dark preference, and the page declares no
`color-scheme`, so Chrome resolved the light palette: `ButtonFace rgb(240,240,240)`,
`ButtonText rgb(0,0,0)`, `Highlight rgb(0,120,212)`, `HighlightText rgb(255,255,255)`,
`Canvas rgb(255,255,255)`, `CanvasText rgb(0,0,0)`.

I deleted my clone after the review.

xamcross:
**Review (second review, Angular engineer): CHANGES REQUESTED**

I made a new clone, and I checked out the head of `issue-25-poll-store`
(`d9ecbef`). I checked each finding against the code, and not against the
response comment. I wrote 19 new checks of my own. 17 pass. 2 fail, and they
show one new MAJOR defect.

## Each finding of my first review

| First finding | Result | Evidence |
| --- | --- | --- |
| MAJOR 1 — the health route read one time, a fixed 60 s fallback, no error signal | **CORRECTED** | `poll-store.ts:90-107`; my checks S1, S3 |
| MAJOR 2 — the gate dropped a tick, thus no request on a reopen | **CORRECTED in the steady state**, see NEW 1 for the start | `poll-store.ts:114-134`; my checks S6, S7, S8 |
| MINOR 1 — a `null` `activeElement` stopped the refresh | **CORRECTED** | `poll-store.ts:57` uses `!= null` |
| MINOR 2 — a double request near a due tick | **CORRECTED** | My check S8 |
| MINOR 3 — no guard on `refreshSeconds` | **CORRECTED** | `poll-store.ts:51-53,95-97`; my check S3 |
| MINOR 4 — no `refresh()`, no first-load signal | **CORRECTED** | `poll-store.ts:41,84,163`; my checks S13, S14, S15 |
| MINOR 5 — two comments broke the language rules | **CORRECTED** | `poll-store.spec.ts:9`, `poll-store.ts:69` |
| The stale check box in the pull request body | **CORRECTED** | The body now marks each criterion. |

## Point 1 — the health route

- **PASS** — The retry sends no parallel request. My check S1: `exhaustMap`
  holds the tick. Two retry ticks pass while the first request is open, and
  `expectNone('/api/health')` holds each time.
- **PASS** — The retry period is 5 s after the answer. S1: nothing at
  4 999 ms, one request at 5 000 ms.
- **PASS** — The retry stops after a valid answer. S1: ten more minutes give
  no health request, while the data poll runs.
- **PASS** — A failed health request sets `error`. A good one clears it.
- **PASS** — An invalid `refreshSeconds` gives 60 000 ms with no retry. My
  check S3 uses the string `"fast"`: the first poll comes at once, the next
  at 60 000 ms, and no health request follows.

## Point 2 — the stop gate

- **PASS** — One request at once after an unpause (S6), after the focus
  leaves the `<tbody>` (S7), and after the tab shows again (S8). Each check
  uses a 60 s or 10 s interval, and the request lands 1 ms after the reopen.
- **PASS** — One request only. S6 and S7 hold the count for 1 000 ms more.
- **PASS** — No double request near a due tick. S8 stops 11 ms before the due
  tick of the first timer, reopens the gate, and then passes the old due
  moment. The count stays at 2, and the next request comes 10 s after the
  reopen.
- **PASS** — A focus move inside one `<tbody>` sends no request (S9).
  `distinctUntilChanged` holds it.
- **PASS** — No subscription stays after a destroy. My check S10 counts the
  listener calls: one add and one remove for each of `visibilitychange`,
  `focusin`, and `focusout`. Ten more minutes and one `visibilitychange`
  after the destroy give no request. My check S11 destroys the host while a
  health request is open: the retry stops, and `httpMock.verify()` is clean.
- **FAIL** — See NEW 1.

## Point 3 — the test helpers

- **`settleGate()` (`poll-store.spec.ts:38-41`) hides no defect.** I measured
  the two steps. My checks S6 and S7 print `afterZero=1 afterOne=2`: the
  request needs the `auditTime(0)` macro task and then the first tick of the
  fresh timer. The interval in both checks is 60 s, and the helper moves the
  clock 1 ms. Thus a code that waits for the next tick cannot pass these
  tests. The 1 ms is a property of the Vitest fake clock, not a real wait.
- **`flushEffects()` (`poll-store.spec.ts:28-30`) hides no defect.**
  `ApplicationRef.tick()` runs the `toObservable` effect. A zoneless app runs
  the same tick by itself after a signal write. My check S4 uses the same
  helper, and it still finds the defect of NEW 1.
- **NOTE** — The helpers are honest, but the test set has a hole. Each test
  answers the health route before it moves the gate. Thus no test covers the
  window between the store creation and the health answer. NEW 1 sits in that
  window.

## Point 4 — `refresh()` and `firstLoadPending`

- **PASS** — `refresh()` sends one request at once, also before the health
  route answers (S13). It does not cancel an open request (S17).
- **PASS** — `refresh()` works while `paused` is true. This is the documented
  intent. See MINOR 5.
- **PASS** — `firstLoadPending` is true at the start, and it stays true while
  the health route fails (S15). It becomes false after the first answer, good
  or bad.
- **NOTE** — `refresh()` does not move the interval. A refresh at 5 s of a
  10 s interval gives the next automatic request 5 s later (S14). This is
  correct for a manual action, but a view must not call `refresh()` in a loop.

## Other checks

- **PASS** — `lastSuccessAt` changes on a success only (S16).
- **PASS** — A slow request is not cancelled, and no parallel request starts
  (S17).
- **PASS** — Two stores keep two separate states (S12).
- **PASS** — Zoneless change detection: a signal write from the RxJS callback
  updates the view with no `detectChanges` call (S18).
- **PASS** — Scope. The one change outside `src/app/poll/` is
  `app.config.ts:2,10`. `app.ts`, `app.html`, and `app.routes.ts` are not in
  the diff. No secret, no database file, and no build output is in the diff.
- **PASS** — No test is empty, and no test is always green. Each new test of
  `pause-refresh-button.spec.ts:51-76` asserts a real attribute.

## New findings

### NEW 1 (MAJOR) — `monitor/frontend/src/app/poll/poll-store.ts:121-122`

```ts
    map(() => canPoll()),
    startWith(canPoll()),
```

`startWith(canPoll())` reads the gate **when `createPollStore` builds the
pipe**. The store subscribes to `gateOpen$` much later, when the health route
gives a valid interval. Thus the first gate value is the state of the store
creation, and not the live state.

A gate change inside that window leaks one request through a closed gate.

- My check S4: I create the store, I set `paused` to true while the health
  request is open, and then I answer the health route. The store sends one
  request. Output: `S4 afterOpen=1 afterMinute=1`. The expected count is 0.
- My check S5: the same with `document.hidden` set to true. The store sends
  one request.
- My check S19: a pause right after the creation shows the same leak
  (`S19 calls=1`).

The leaked request writes `data` and `lastSuccessAt`, thus the table updates
while the user holds "Pause refresh" pressed. This breaks WCAG 2.2.2. The
window is not small: while the monitor server is down, the health retry runs
each 5 seconds, and D30 expects the user to see that state. A user who pauses
or leaves the tab at that moment hits the defect.

The correction after the leak works: the gate closes one macro task later, and
the count stays at 1.

**Fix (I tested it).** Move `startWith` before `map`, so the map runs at the
subscription:

```ts
  ).pipe(
    auditTime(0),
    startWith(null),
    map(() => canPoll()),
    distinctUntilChanged(),
  );
```

`defer()` around the whole pipe does not work, because `toObservable()` needs
an injection context. The order above needs no `defer`.

I applied this one change in my clone. My checks S4, S5, and S19 then pass,
and each of the 29 tests of this pull request stays green (`4 files, 43
tests`).

Add a regression test: set `paused` to true, or `document.hidden` to true,
while the health request is open, then answer the health route, and assert 0
requests.

### NEW 2 (MINOR) — `poll-store.spec.ts:274-285`

The test defines `document.activeElement` as `null` with
`Object.defineProperty`, and it never removes that property. Each later test
in this file runs with a `null` `activeElement`. The file passes today by
luck. A later focus test in this file will fail for a hidden reason.

Fix: restore the property in the `afterEach`:
`delete (document as unknown as Record<string, unknown>)['activeElement'];`

### NEW 3 (MINOR) — `poll-store.ts:90-107` plus `poll-store.ts:132-134`

The store sends no data request at all while the health route fails. My check
S2 proves it: five failed health answers give 0 data requests. The earlier
version used a fallback interval, thus it polled. Both routes come from the
same server, so I accept the choice. Please record it in the doc comment, and
tell issue #20: the first-load state can last minutes, and only
`firstLoadPending` plus `error` describe it.

### NEW 4 (MINOR) — `poll-store.spec.ts:32-37`

The doc comment of `settleGate` is one sentence of 26 words. The language
policy gives a maximum of 25 words for a description. Split it into two
sentences.

### NEW 5 (MINOR) — `poll-store.ts:112` and `poll-store.ts:41`

- Line 112: "so the focus change lands first" uses an idiom. Write "so the
  browser moves the focus first".
- Line 41: `refresh()` passes the gate, thus it polls while `paused` is true.
  The comment states the intent, and my check S13 confirms it. Please add one
  sentence: a view must not call `refresh()` on a timer, because that breaks
  the pause of WCAG 2.2.2.

## What I ran

In `monitor/frontend` of my own clone, on `d9ecbef`:

- `npm ci` — 497 packages, 0 vulnerabilities.
- `npm test` — `Test Files 6 passed (6)`, `Tests 29 passed (29)`.
- `npm run lint` — "All files pass linting."
- `npm run format:check` — "All matched files use Prettier code style!"
- `npm run build` — builds `dist/frontend` without an error.
- My own 19 checks — 17 pass, 2 fail. The 2 failures are NEW 1.

One MAJOR finding remains, thus I cannot approve. Each other point of my first
review is correct now.

xamcross:
## Correction round for the second review

I fixed each finding of the two second reviews. `npm test` gives 6 test
files and 31 tests, all green.

### MAJOR findings

1. **Angular — `poll-store.ts:121-122` (now `129-132`).** The gate read
   `canPoll()` at the pipe build, not at the subscription. A pause or a
   hidden tab in that window let one request pass a closed gate. I wrote
   two failing tests first, in `poll-store.spec.ts`: "sends no request
   when paused becomes true while the health request is open" and "sends
   no request when document.hidden becomes true while the health request
   is open". Both failed against the old code (`calls` was 1, not 0).
   Fix: `startWith(null)` now stands before `map(() => canPoll())`. Both
   tests pass now.
2. **Accessibility — `pause-refresh-button.css:25` (now `38-40`).** The
   focus ring took `currentColor`, and vanished in the pressed state.
   Fix: `outline: 2px solid CanvasText;`.
3. **Accessibility — `pause-refresh-button.css:19-21` (now `13-16`).** The
   check mark of `::before` entered the accessible name. Fix:
   `content: '\2713\00a0' / '';`. I renamed the test at
   `pause-refresh-button.spec.ts:57` to "keeps the text 'Pause refresh' in
   both states", because it reads `textContent`. `npm run build` accepts
   the value and builds `dist/frontend` without an error.

### MINOR findings

- **Accessibility MINOR 1 — the button changed size when pressed.** Fix
  in `pause-refresh-button.css:13-16,25-27`: `button::before` holds the
  check mark and `visibility: hidden` in both states;
  `button[aria-pressed='true']::before` sets `visibility: visible`.
- **Accessibility MINOR 2 — the `Highlight` pair held no margin.** Fix in
  `pause-refresh-button.css:18-23,29-35`: the normal mode now uses
  `#ffffff` text on a `#003366` background for the pressed state, a
  contrast of 12.6:1. `Highlight` and `HighlightText` stay inside
  `@media (forced-colors: active)` only.
- **Accessibility MINOR 3 and Angular MINOR — the same point.** Fix in
  `poll-store.ts:40-45`: the doc comment of `refresh()` now reads "Call
  `refresh()` only from a user action. A pause must stop each automatic
  update."
- **Angular MINOR — a test left `document.activeElement` as `null` for
  the rest of the file.** Fix in `poll-store.spec.ts:56-61`: the shared
  `afterEach` now deletes the property.
- **Angular MINOR — a 26-word doc comment.** Fix in
  `poll-store.spec.ts:32-37`: the comment now holds three sentences, each
  at or under 20 words.
- **Angular MINOR — the idiom at `poll-store.ts:112` (now `119`).** Fix:
  "so the focus change lands first" now reads "so the browser moves the
  focus first".
- **Angular MINOR — the failed health route.** Fix in `poll-store.ts:76-77`:
  one sentence records that the store sends no data poll while the health
  route fails, for issue #20. The behaviour itself did not change.

### What I did not change

The button and the store stay unmounted. No live region was added. Issue
#20 owns that work, as the second review of the accessibility specialist
also states.

### Commands run in `monitor/frontend`

- `npm ci`: 497 packages, 0 vulnerabilities.
- `npm test`: 6 test files, 31 tests, all green.
- `npm run lint`: "All files pass linting."
- `npm run format:check`: "All matched files use Prettier code style!"
- `npm run build`: builds `dist/frontend` without an error.

xamcross:
**Review (third review, accessibility specialist): APPROVE**

I made a new clone at the head `9265988`. I measured each point in Chrome 141 on Windows 11,
with the exact rules of `pause-refresh-button.css`. Each finding of my first review and of my
second review is corrected. I found no new defect.

## The measurements

**1. The focus ring — CORRECTED.** I pressed a real Tab key, so `:focus-visible` matched
(`focusVisible: true`). The used value is `outline: 2px solid rgb(0, 0, 0)`, `outline-offset:
2px`, in **both** states:

| State | `outline-color` | Contrast against the page |
| ----- | --------------- | ------------------------- |
| Not pressed | `rgb(0, 0, 0)` | **21.00 : 1** |
| Pressed | `rgb(0, 0, 0)` | **21.00 : 1** |

`CanvasText` resolves to `rgb(0, 0, 0)` and `Canvas` to `rgb(255, 255, 255)`. The ring now
follows the page, and not the text of the button, so the state cannot change it. WCAG 1.4.11
(a minimum of 3:1) and WCAG 2.4.7 pass.

A note on the method: a computed `outline-color` on a button with no focus falls back to
`currentColor`, and it reports white for the pressed state. That reading is not the rule. The
numbers above come from a real keyboard focus.

**2. The accessible name — CORRECTED.** `content` computes to `"✓ " / ""` in both states. The
alternative text is the empty string, thus the pseudo-element gives nothing to the name
(accname 1.2, step 2F.ii). `textContent` is "Pause refresh" in both states. The glyph still
renders; I confirmed it in a screenshot. The renamed test at `pause-refresh-button.spec.ts:57`
now matches what it reads. #67 gives the final proof with NVDA.

**3. The size — CORRECTED.** 122.04 by 27.33 CSS px in **both** states. The difference is 0.00
by 0.00 px, thus no layout shift. I checked the scale with a 24 px reference box, and it
measured exactly 24 by 24, so the numbers are true CSS px. WCAG 2.5.8 passes.

**4. The project colours — CORRECTED.**

| Measure | Value | Rule | Result |
| ------- | ----- | ---- | ------ |
| `#ffffff` on `#003366` | **12.61 : 1** | 1.4.3 needs 4.5:1 | pass |
| Pressed button against the page | **12.61 : 1** | 1.4.11 needs 3:1 | pass |
| Pressed background against the not-pressed background | 11.06 : 1 | — | the two states differ strongly |
| Not-pressed border `rgb(0,0,0)` against the page | 21.00 : 1 | 1.4.11 | the control boundary stays clear |

The state stays clear without colour. The check mark shows only in the pressed state, and the
border goes from `solid` to `inset`. The screenshot confirms the glyph. WCAG 1.4.1 passes.

The forced-colours mode keeps the system pair inside `@media (forced-colors: active)`, which
is the correct pattern. `visibility` is not a forced property, so the check mark still carries
the state there. Chrome forces `outline-color` in that mode, so the ring stays visible. I did
not switch Windows to the high-contrast mode, so this point rests on the rule structure and
not on a measurement. #67 covers the real check.

**5. The `refresh()` sentence — CORRECTED.** `poll-store.ts:40-45` reads "Call `refresh()`
only from a user action. A pause must stop each automatic update."

## The checks

`npm ci` 0 vulnerabilities. `npm test` 6 files, 31 tests, all green. `npm run lint` clean.
`npm run format:check` clean. `npm run build` builds `dist/frontend`.

The component still holds no live region, and nothing mounts the button or the store. The diff
holds 8 source files, no secret, and no build output.

## One note for a later issue, not a finding

`#003366` is a fixed value, and it does not follow a page theme. The app declares no
`color-scheme` today, so the page stays light and the numbers above hold. If a later issue
adds a dark theme, the pressed button falls to about 1.5:1 against a `#121212` page, and it
fails 1.4.11. Hold the two colours in custom properties when that theme arrives.

I deleted my clone after the review.

xamcross:
**Review (third review, Angular engineer): APPROVE**

I made a new clone, and I checked out the head `9265988`. I checked the code,
and not the response comment.

## Point 1 — the gate order

- **CORRECTED** — `poll-store.ts:129-131` holds `auditTime(0)`,
  `startWith(null)`, then `map(() => canPoll())`. The first gate value now
  reads `canPoll()` at the subscription.
- My checks S4, S5, and S19 pass: a pause or a hidden tab while the health
  request is open gives 0 requests, and the count stays at 0 for one minute.
  Against the old code each check gave 1.
- **The two new tests can fail.** I put the old order back in my clone
  (`map(() => canPoll())`, then `startWith(canPoll())`) and ran the suite:
  `sends no request when paused becomes true while the health request is
  open` and `sends no request when document.hidden becomes true while the
  health request is open` both failed, with 5 failures in total. I then
  restored the file. The two tests are real.

## Point 2 — each MINOR finding

| Finding | Result | Evidence |
| --- | --- | --- |
| A test left `document.activeElement` as `null` | **CORRECTED** | `poll-store.spec.ts:61-62`: the shared `afterEach` deletes the property. |
| No sentence on the failed health route | **CORRECTED** | `poll-store.ts:76-77`: "The store sends no data poll while the health route fails. Issue #20 must show that state to the user." |
| A 26-word doc comment | **CORRECTED** | `poll-store.spec.ts:32-37`: three sentences, each at or under 20 words. |
| The idiom "lands first" | **CORRECTED** | `poll-store.ts:119-120`: "so the browser moves the focus first". |
| No warning on `refresh()` | **CORRECTED** | `poll-store.ts:40-45`: "Call `refresh()` only from a user action. A pause must stop each automatic update." |

## Point 3 — the accessibility fixes

- The commit `9265988` touches four files. The accessibility part is
  `pause-refresh-button.css` (24 lines) plus one test name at
  `pause-refresh-button.spec.ts:57`. The assertions of that test did not
  change.
- The CSS broke no test and no build. `npm test` gives 31 tests, all green.
  `npm run build` builds `dist/frontend` without an error.

## Scope

The one change outside `src/app/poll/` stays `provideHttpClient()` in
`app.config.ts`. `app.ts`, `app.html`, and `app.routes.ts` are not in the
diff. No secret and no build output is in the diff.

## What I ran

In `monitor/frontend`, on `9265988`:

- `npm ci` — 497 packages, 0 vulnerabilities.
- `npm test` — `Test Files 6 passed (6)`, `Tests 31 passed (31)`.
- `npm run lint` — "All files pass linting."
- `npm run format:check` — "All matched files use Prettier code style!"
- `npm run build` — builds `dist/frontend` without an error.
- My own 4 checks — all pass. The mutation run above proves they can fail.

No BLOCKER and no MAJOR finding remains. I approve the store.

Issue #35, opened 2026-09-21
Delete the events of one user id in the kit stores


xamcross:
**Goal.** `deleteByUserId` removes each event of one user id from the app log. It also removes each event with `userId: null` of a session of that user (rule E1).

**Why it blocks production.** The app log keeps the events of a deleted account for 30 days, and the monitor can import them again.

**Implementation steps.**
1. In `kit/jvm-core`, implement `deleteByUserId` in the in-memory store with rule E1.
2. In `kit/jvm-mongo`, read the distinct `sessionId` values of that user id first.
3. Delete each event with that `userId`, and each event with `userId: null` whose `sessionId` is in that list.
4. Keep a row of a second user of the same session. That row holds another user id.
5. Add no new index. A collection scan of a maximum of 200 000 documents is acceptable for a rare action.
6. Add a Testcontainers test for the MongoDB store with `@Testcontainers(disabledWithoutDocker = true)`.

**Acceptance criteria.**
- [x] In each of the two stores, no document of that user id remains after the call, and the other documents stay.
- [x] A document with `userId: null` of a session of that user is gone.
- [x] A document of a second user in the same session stays.
- [x] A document with `userId: null` of another session stays.
- [x] `./gradlew :kit:jvm-core:test` and `./gradlew :kit:jvm-mongo:test` pass.
- [x] A comment on this issue records the test output.

**Related.**
- #40 documents the call.
- #64 documents the erasure order of D15.
- #61 uses the same rule in the monitor.
- The design change of 2026-09-21 added rule E1 to this issue.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, D15, D18, D22.

xamcross:
Body edit of 2026-09-21: rule E1 of the design change. An anonymous row of a session joins the erasure of the user of that session, because the two rows name one person. A row of a second user in the same session stays. The issue keeps 6 steps. The first line drops the reference to #26, because #26 closed on 2026-09-21.

The maintainer made this edit. The source is the design brief in the first comment of #101.

xamcross:
Body edit on 2026-09-22: #11 closed. No open blocker remains, thus the first line and the label `blocked` are removed. The work can start.

xamcross:
Pull request #157 implements this issue. Test output below, JDK 21, Docker
present on the build machine, 2026-09-22.

| Task | Tests | Skipped | Failures | Errors |
|---|---|---|---|---|
[...]

Issue #34, opened 2026-09-21
Stop the ingest above the event cap


xamcross:
**Goal.** Above `OCTOMETER_MAX_EVENTS` the MongoDB store drops each new batch, thus the metrics never fill the cluster.

**Why it blocks production.** A full Atlas M0 cluster refuses each write of the app, not only the metrics.

**Implementation steps.**
1. In `kit/jvm-mongo`, read `OCTOMETER_MAX_EVENTS` with the default 200 000.
2. Read `estimatedDocumentCount()` a maximum of one time each 60 seconds. Use an injected clock.
3. Above the cap, drop the batch in `append`.
4. Write one warning when the store drops a batch.
5. In `kit/jvm-ktor`, add a test with a store that drops the batch. The route of #12 returns 204 (section 4.2).
6. Add the Testcontainers tests with `@Testcontainers(disabledWithoutDocker = true)`.

**Acceptance criteria.**
- [x] Above `OCTOMETER_MAX_EVENTS` the store drops the batch, the route returns 204, and the store writes one warning.
- [x] Fixed clock: the store calls `estimatedDocumentCount()` a maximum of one time in 60 seconds.
- [x] The test deletes documents, as the TTL does. After the next count the store accepts a batch again.
- [x] `./gradlew :kit:jvm-mongo:test` and `./gradlew :kit:jvm-ktor:test` pass.
- [x] A comment on this issue records the test output.

**Related.** #117 adds the daily anonymous cap and the user-agent filter on the same route.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, D21, section 4.2.

xamcross:
Body edit of 2026-09-21: rules S3 and S4 of the design change. The daily anonymous cap and the robot filter of #117 extend the event cap of this issue. A Related line names #117. The steps stay as they are.

The maintainer made this edit. The source is the design brief in the first comment of #101.

xamcross:
Body edit on 2026-09-22: #11 closed. The first line now names only the open blockers: #12.

xamcross:
Body edit on 2026-09-22: #12 closed. No open blocker remains, thus the first line and the label `blocked` are removed. The work can start.

xamcross:
Pull request #165 implements this issue. Test output, all on JDK 21:

- `./gradlew :kit:jvm-core:test`: 238 tests, 0 failures, 0 errors.
[...]

Issue #11, opened 2026-09-21
Build the MongoDB event store in kit/jvm-mongo


xamcross:
**Goal.** `kit/jvm-mongo` holds an `EventLogStore` for the MongoDB database of the app.

**Implementation steps.**
1. Create the module `kit/jvm-mongo` with Java 21 and `compileOnly` `org.mongodb:mongodb-driver-sync:5.0.1`.
2. Add the store class for `octometer_events`. Its constructor takes a `com.mongodb.client.MongoDatabase` from the app, and the kit never creates a client.
3. Implement `append`. The kit creates the `_id` inside the insert call.
4. At the start, create the TTL index on `ts` from `OCTOMETER_RETENTION_DAYS` (default 30 days). On error code 85, run `collMod`.
5. Write a warn-level log line for an index error. Never fail the app start.
6. Confirm the newest 5.x driver version with Maven Central before you use it.

**Acceptance criteria.**
- [x] A stored document has exactly the fields of section 4.1 and no `_class` field.
- [x] A second start with a different `OCTOMETER_RETENTION_DAYS` changes `expireAfterSeconds` and does not fail.
- [x] An index error gives a warn-level log line, and the store still works.
- [x] `./gradlew :kit:jvm-mongo:test` passes against driver 5.0.1 and the newest 5.x. The tests use Testcontainers MongoDB with `@Testcontainers(disabledWithoutDocker = true)`.
- [x] A comment on this issue records the test output for the two driver versions.

**Related.**
- #35 adds `deleteByUserId`.
- #34 adds the event cap.
- #105 writes `path` and `referrerHost` into the event document.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, section 4.1, D17, D22.

xamcross:
Body edit on 2026-09-21: #26 closed. No open blocker remains, thus the first line and the label `blocked` are removed. The work can start.

xamcross:
Body edit of 2026-09-21: rule K1 of the design change. The MongoDB store writes the two new fields `path` and `referrerHost`. Issue #105 makes that change, thus a Related line names it. The steps stay as they are.

The maintainer made this edit. The source is the design brief in the first comment of #101.

xamcross:
Pull request #126 adds `kit/jvm-mongo`. The tests ran twice, against the two
driver versions, both locally and on the CI job "JVM modules" (Ubuntu,
Docker). Each test printed its own result; no test skipped.

[...]

Issue #10, opened 2026-09-21
Build the ingest logic of kit/jvm-core


xamcross:
**Goal.** `kit/jvm-core` holds the parser, the validation, and the `ts` rule of section 4.2.

**Implementation steps.**
1. Create the module `kit/jvm-core`: Java 21, no run-time dependency, no Kotlin library.
2. Add a strict parser that takes the raw body as a `String`.
3. Add the validation of sections 4.1 and 4.2.
4. Calculate `ts = receivedAt - ageMs`, with the clamp of an `ageMs` above 600 000.

**Acceptance criteria.**
- [x] `./gradlew :kit:jvm-core:test` passes. One test covers each invalid input:
  - 51 clicks
  - a 101-character element
  - a space in an element
  - a non-UUID session id
  - a negative `ageMs`
  - a body above 16 KB
  - an unknown top-level type
- [x] With a fixed clock `ts` is correct, and an `ageMs` of 700 000 becomes 600 000.
- [x] A comment on this issue records the test output.

**Related.**
- #8 defines the format.
- #26 adds the `UserIdResolver`, the `EventLogStore`, the in-memory store, and the rule of D19.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, 4.1, 4.2, D17, D18.

xamcross:
Body edit on 2026-09-21: #3 closed. No open blocker remains, thus the first line and the label `blocked` are removed. The work can start.

xamcross:
## Test output

Pull request: https://github.com/xamcross/octometer/pull/89

Command: `./gradlew :kit:jvm-core:clean :kit:jvm-core:test --console=plain --rerun`

```
> Task :kit:jvm-core:clean
> Task :kit:jvm-core:compileJava
> Task :kit:jvm-core:processResources NO-SOURCE
> Task :kit:jvm-core:classes
> Task :kit:jvm-core:compileTestJava
> Task :kit:jvm-core:processTestResources NO-SOURCE
> Task :kit:jvm-core:testClasses
> Task :kit:jvm-core:test

BUILD SUCCESSFUL in 5s
4 actionable tasks: 4 executed
```

Test result files, 26 tests, 0 failures, 0 errors:

```
TEST-octometer.kit.core.ingest.EventFieldValidatorTest.xml:       tests="10" failures="0" errors="0"
TEST-octometer.kit.core.ingest.IngestParserContractExamplesTest.xml: tests="7"  failures="0" errors="0"
[...]

PR #94, opened 2026-09-21
feat(core): add the store interface and the in-memory store to kit/jvm-core

### LINKED ISSUE #26: Add the store interface and the in-memory store to kit/jvm-core
**Goal.** `kit/jvm-core` gives each valid event, with the user id from the app, to an `EventLogStore`.

**Implementation steps.**
1. Add the interface `UserIdResolver` to `kit/jvm-core`. The app gives the implementation.
2. Add the interface `EventLogStore` with `append` and `deleteByUserId`.
3. Add an in-memory store that implements `EventLogStore`.
4. Give each valid event to `append`, with the user id from the `UserIdResolver`.
5. Drop an event without a user id unless `OCTOMETER_RECORD_ANONYMOUS=true`.

**Acceptance criteria.**
- [x] A `userId` field in the body does not change the stored user id.
- [x] Without a user id and with the flag off, the store gets 0 events.
- [x] Without a user id and with the flag on, the store gets the event with `userId: null`.
- [x] `./gradlew :kit:jvm-core:dependencies --configuration runtimeClasspath` shows no dependency.
- [x] `./gradlew :kit:jvm-core:test` passes.
- [x] A comment on this issue records the two outputs.

**Related.**
- #11 and #12 use the two interfaces.
- #35 implements `deleteByUserId` in the stores.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, section 4.2, D17, D18, D19.

xamcross:
## Summary

This pull request adds `UserIdResolver` and `EventLogStore` to `kit/jvm-core`. It adds an
in-memory store for `EventLogStore`. `IngestPipeline.ingest` gives the whole batch of valid
events to the store in one call, with the user id from `UserIdResolver`. A missing user id
drops the batch, unless `OCTOMETER_RECORD_ANONYMOUS` is on (design decisions D18, D19).

**Correction round 1** (commit `ff99e55`) answers two reviews. The full list of findings and
fixes is in my response comment on this pull request.

## Public API of #10

`IngestEvent` stays as it was in #10 (no new field). `EventLogStore.append` takes a
`List<IngestEvent>` and the user id as two separate parameters, so the store holds the user
id with no change to `IngestEvent`. A default method wraps one event in a batch of one. This
is the only design choice beyond the issue text.

## The in-memory store is for a test and a demo

`InMemoryEventLogStore` holds each event in the heap, with no size limit, and it loses each
event at a restart. An app in production uses the MongoDB store of issue #11; that store
holds the event cap `OCTOMETER_MAX_EVENTS` of design decision D21 and issue #34. Issue #34
does not touch this in-memory store; each of its steps names `kit/jvm-mongo` or
`kit/jvm-ktor`. My first pull request text said the wrong thing here; this text and the class
comment now agree. `deleteByUserId` is part of the store, per the issue text; issue #35 adds
it to the other stores. It rejects a `null` value and an empty text.

## The trust boundary of the user id

The client never sets the user id (rules C6, C16, C32); the tests attack this boundary with
several body shapes. A bad value from the app's own `UserIdResolver` is a defect of the app,
not of the client: `ingest` throws `IllegalStateException`, never `IngestException`, so an
adapter never returns 400 for a valid body. `StoredEvent.toString()` holds no user id and no
session id (design decision D15).

## Acceptance criteria

- [x] A `userId` field in the body does not change the stored user id.
  Evidence: `IngestPipelineIngestTest.aUserIdFieldInTheBodyDoesNotChangeTheStoredUserId`, plus
  `aUserIdFieldInsideAClickEntryDoesNotChangeTheStoredUserId` and
  `aUserIdFieldWithADifferentLetterCaseDoesNotChangeTheStoredUserId`.
- [x] Without a user id and with the flag off, the store gets 0 events.
  Evidence: `IngestPipelineIngestTest.withoutAUserIdAndWithTheFlagOffTheStoreGetsZeroEvents`.
- [x] Without a user id and with the flag on, the store gets the event with `userId: null`.
  Evidence: `IngestPipelineIngestTest.withoutAUserIdAndWithTheFlagOnTheStoreGetsTheEventWithUserIdNull`.
- [x] `./gradlew :kit:jvm-core:dependencies --configuration runtimeClasspath` shows no
  dependency.
  Evidence (run on 2026-09-21, JDK 21 at `C:/Users/xamcr/.jdks/ms-21.0.9`):
  ```
  runtimeClasspath - Runtime classpath of source set 'main'.
  No dependencies
  ```
- [x] `./gradlew :kit:jvm-core:test` passes.
  Evidence (run on 2026-09-21): `BUILD SUCCESSFUL`, 69 tests, 0 failures, 0 errors.
- [x] A comment on this issue records the two outputs.
  Evidence: https://github.com/xamcross/octometer/issues/26#issuecomment-5765049285

## Thread safety

`InMemoryEventLogStore` holds its events in a `ConcurrentLinkedQueue`. One test runs 8
append threads with 500 appends each. A second test runs 4 append threads together with one
delete thread, then checks the final count and that no event of the deleted user remains
(the test calls `deleteByUserId` a second time after the threads finish, because
`ConcurrentLinkedQueue.removeIf` is weakly consistent).

## Test-driven development

Each new test of this round failed first, against the code of the first commit, before I
changed the production classes. Each test then passed after the fix.

Closes #26

Generated with Claude Code


xamcross:
**Review (security and privacy engineer): CHANGES REQUESTED**

First review. I made my own clone, I checked out the head of `issue-26-event-log-store`
(commit `fdf9110`), and I ran each command myself with JDK 21. I also wrote 15 probe tests and
two small programs of my own. They stay in my clone; they are not a part of this pull request.

## The acceptance criteria

|---|---|---|---|
| 1 | A `userId` field in the body does not change the stored user id. | PASS | 6 probe tests. See "Point 1" below. |
| 2 | Without a user id and with the flag off, the store gets 0 events. | PASS | Probe test: `store.events()` is empty. |
| 3 | Without a user id and with the flag on, the store gets the event with `userId: null`. | PASS | Probe test: 1 event, `userId()` is `null`. |
| 4 | `./gradlew :kit:jvm-core:dependencies --configuration runtimeClasspath` shows no dependency. | PASS | Output: `runtimeClasspath - Runtime classpath of source set 'main'.` and `No dependencies`. |
| 5 | `./gradlew :kit:jvm-core:test` passes. | PASS | `BUILD SUCCESSFUL`. The XML reports give 56 tests, 0 failures, 0 errors. |
| 6 | A comment on the issue records the two outputs. | PASS | The comment of 2026-09-21 on issue #26 holds both outputs. |

## The points of my role

**Point 1. The body cannot bring a user id in.** PASS, and the design is strong. `IngestParser`
extracts only `sessionId`, `element`, and `ageMs`. `ParsedIngestRequest` holds no user id
field. A user id from the body is therefore not only ignored; it has no place to go. I proved
each attack shape:

- `"userId"` at the top level: the store gets the value of the resolver.
- `"userId"` inside a click entry: the same result.
- `"UserId"`, `"USERID"`, and `"userid"` together: the same result. The key match is
  case-sensitive, thus each one is an unknown field.
- A duplicate `"userId"` key: the parser throws `DUPLICATE_FIELD` (rule C36). The store stays
  empty.
- A nested object `{"auth": {"user": {"userId": "..."}}}`: `skipValue` discards it.
- `"userId": null` and `"user_id": 42`: the same result.

**Point 2. The resolved user id.** Mostly PASS. `IngestPipeline.ingest` line 71 calls
`EventFieldValidator.validateUserId`. An empty text and a text of 255 characters throw
`IngestException` with the reason `USER_ID_LENGTH`, and the store stays empty. A rejected user
id is therefore not stored as `null`. That part is correct. A text of 254 characters passes. A
resolver that throws propagates its own exception, and the store stays empty. Two gaps remain:
see MAJOR 3 and MINOR 2.

**Point 3. The flag `OCTOMETER_RECORD_ANONYMOUS`.** PASS. I ran a small program with the
compiled class and 15 different values. The flag is on for `true`, `TRUE`, `True`, and `TrUe`.
It is off for an unset variable, an empty text, ` true` (a leading space), `true ` (a trailing
space), `1`, `yes`, `on`, `YES`, `enabled`, `false`, and `0`. The result matches D19 exactly,
and every unknown value fails closed. With the flag off and no user id, the store stays empty.
The module holds no logger, therefore a log line leaks nothing. See MINOR 3.

**Point 4. `deleteByUserId`.** One defect. See MAJOR 2. The other cases are correct.
`Objects.equals` is an exact match, thus `deleteByUserId("user-1")` keeps `user-10`, `user-1x`,
and `USER-1`. An empty text removes nothing. A concurrent test with one thread that appends
5000 events of the same user and one thread that erases that user completed with no exception,
and it kept each event of the other user.

**Point 5. Logs and exceptions.** PASS for the code, with one latent defect. The module holds
no logger, no `System.out`, and no `printStackTrace`. Each `IngestException` message is a
constant text. No message holds a user id, a session id, or the raw body. I checked the message
of the 255-character case: it holds neither the value nor the session id. But see MAJOR 4 for
the `toString` of the new record.

**Point 6. The in-memory store has no cap.** The claim of the author is wrong. See MAJOR 1.

**Point 7. The scope.** PASS. The diff touches only `kit/jvm-core` (10 files). It changes no
build file, thus no new dependency enters. The diff holds no secret, no database file, and no
build output.

## The findings

### MAJOR 1 — `kit/jvm-core/src/main/java/octometer/kit/core/store/InMemoryEventLogStore.java:11-12`

The comment says "Issue #34 adds the event cap `OCTOMETER_MAX_EVENTS`". That claim is wrong. I
read issue #34. Each of its six steps names `kit/jvm-mongo` or `kit/jvm-ktor`. Its acceptance
criteria name `:kit:jvm-mongo:test` and `:kit:jvm-ktor:test`. Issue #34 is also blocked by #11
and #12. No issue therefore gives a cap to this class. The `ConcurrentLinkedQueue` sits in the
ingest path of the app server and it grows with no limit. An app that wires this store into
production runs out of heap memory. That is a denial of service of the whole app server, not
only of the metrics.

**The fix.** Replace the class comment with this text:

```java
/**
 * An in-memory {@link EventLogStore}. A call from more than one thread is
 * safe.
 *
 * <p><b>Use this store only in a test and in a demo.</b> It holds each
 * event in the heap with no size limit, and it loses each event at a
 * restart. An app in production uses the MongoDB store of issue #11. That
 * store holds the event cap {@code OCTOMETER_MAX_EVENTS} of design
 * decision D21 and issue #34.
 */
```

### MAJOR 2 — `InMemoryEventLogStore.java:26` and `EventLogStore.java:20-22`

`deleteByUserId(null)` erases every anonymous event of every visitor.
`Objects.equals(storedEvent.userId(), null)` is true for each event with no user id. My probe
test proves it: a store with one event of `user-1` and two anonymous events keeps only the
`user-1` event after `deleteByUserId(null)`.

Design decision D15 says that the erasure starts in the app, and the app erases through this
interface. Design decision D13 gives the route shape: `DELETE
/api/apps/{appId}/events?userId=<id>`. An absent query parameter is `null` in most frameworks.
A request with no `userId` parameter therefore deletes the whole anonymous log. D13 also shows
that an anonymous selection needs its own parameter, `anonymous=true`; `null` is never a valid
selector on its own. The interface comment states no rule for `null`, thus issue #35 carries
the same defect into the MongoDB store.

**The fix.** Reject `null` and an empty text in the interface and in the implementation.

In `EventLogStore.java`, replace the comment of `deleteByUserId` with this text:

```java
    /**
     * Deletes each stored event with the given user id. The value of
     * {@code userId} must not be {@code null}, and it must not be an
     * empty text. An implementation throws
     * {@link NullPointerException} for {@code null}. An anonymous event
     * has no erasure, because it holds no user id (contract rule C6).
     */
```

In `InMemoryEventLogStore.java`, add the check as the first line of the method:

```java
    @Override
    public void deleteByUserId(String userId) {
        Objects.requireNonNull(userId, "userId must not be null");
        events.removeIf(storedEvent -> userId.equals(storedEvent.userId()));
    }
```

Add one test for the `null` case.

### MAJOR 3 — `kit/jvm-core/src/main/java/octometer/kit/core/ingest/IngestPipeline.java:71`

A bad value from the app's own resolver throws `IngestException`. Contract rules C19 and C33
tie that exception to status 400, "an invalid body". But the body is valid here; the app is at
fault. An adapter therefore returns 400 to the tracker for a correct request. Two results
follow. The operator reads a client error and never repairs the resolver. The metrics of that
one signed-in user stop for ever, and nobody sees the cause. The adapter cannot separate the
two faults from the exception type, because the module documents only `IngestException`.

**The fix.** Give the adapter the information that it needs. Add this text to the comment of
the reason `USER_ID_LENGTH` in `IngestException.java:35`:

```java
        /**
         * The `userId` value is not 1 to 254 characters (rule C6). The
         * client never sends this value (rules C16 and C32), thus this
         * reason always names a fault of the app's
         * {@code UserIdResolver}. An adapter maps this one reason to
         * status 500, and not to status 400.
         */
```

Add the same rule to the comment of `IngestPipeline.ingest` (line 63-65). A separate exception
class, for example `UserIdResolverException`, is the cleaner answer. Either fix is acceptable.

### MAJOR 4 — `kit/jvm-core/src/main/java/octometer/kit/core/store/StoredEvent.java:10`

`StoredEvent` is a record, thus Java generates its `toString`. That text holds the user id and
the session id. I ran it:

```
StoredEvent.toString -> StoredEvent[sessionId=..., element=a.b, ts=..., userId=internal-user-42]
```

Design decision D15 states: "A log line never holds a user id or a connection string." D15 also
adds a sentinel test that searches the captured log. An adapter author who writes one debug
line with a `StoredEvent` breaks D15, and nothing warns that author. This record is new in this
pull request, and it adds the exact field that D15 names.

**The fix.** Override `toString` in `StoredEvent`:

```java
    /**
     * Returns a text with no personal data (design decision D15). The
     * user id and the session id stay out of this text, because a log
     * line must never hold them.
     */
    @Override
    public String toString() {
        return "StoredEvent[element=" + element + ", ts=" + ts
                + ", userId=" + (userId == null ? "null" : "<redacted>") + "]";
    }
```

Add one test that checks that the text holds neither the user id nor the session id.

### MINOR 1 — `kit/jvm-core/src/test/java/octometer/kit/core/ingest/IngestPipelineIngestTest.java:76`

The name of the test is `aUserIdAboveTheLengthLimitMakesTheRequestInvalid`. The resolver of
line 82 returns an empty text, which is below the limit. The name states a behaviour that the
test does not check. Rename it to `anEmptyResolvedUserIdMakesTheRequestInvalid`. Add a second
test with a user id of 255 characters. `EventFieldValidatorTest` already covers the two
boundary values for the validator alone.

### MINOR 2 — `kit/jvm-core/src/main/java/octometer/kit/core/ingest/EventFieldValidator.java:58-67`

`validateUserId` checks only the length. My probe proves that an email address, an IP address,
and a text with a carriage return, a line feed, a null character, and an ANSI escape all reach
the store. Contract rule C6 and owner decision O5 forbid an email address and an IP address. A
kit cannot detect each one, but it can reject a character that no internal id needs. Add a
rule: each character must be above U+001F, and it must not be U+007F. That rule protects a
later log line and a later export.

### MINOR 3 — `kit/jvm-core/src/main/java/octometer/kit/core/ingest/IngestSettings.java:16`

The flag accepts only a case-insensitive `true`. That choice matches D19, and it fails closed.
But `1`, `yes`, `on`, and ` true` with a leading space give no signal. An operator who writes
one of those believes that the record of an anonymous click is on. Add a `trim()`, and write
one warning at the start for a value that is neither empty nor a form of `true` or `false`. The
warning holds no personal data, thus D15 stays correct.

### MINOR 4 — `IngestPipeline.java:63-65`

The comment of `ingest` says nothing about a resolver that throws. I proved that the exception
of the resolver propagates, and that the store stays empty. State that rule in the comment. An
adapter then knows that it must catch the exception of its own resolver.

### MINOR 5 — `IngestPipelineIngestTest.java`

Acceptance criterion 1 has one test, for a `userId` field at the top level. Five other shapes
reach the same code path. I proved each one, and each one is correct today. Add the strongest
two as tests: a `userId` inside a click entry, and a `userId` with a different letter case. A
future change of the parser then cannot break the rule in silence.

## Summary

4 MAJOR findings and 5 MINOR findings. No BLOCKER. The trust boundary of the user id is
correct by construction, and I could not break it. The flag matches D19 and it fails closed.
The four MAJOR findings are about the erasure path, the response code of a server fault, the
memory limit, and the `toString` of the new record.

xamcross:
**Review (Java library and API engineer): CHANGES REQUESTED**

I made my own clone. I ran each command myself. Branch `issue-26-event-log-store`, head `fdf9110`.
JDK 21 at `C:/Users/xamcr/.jdks/ms-21.0.9`.

## Acceptance criteria

1. **PASS** — A `userId` field in the body does not change the stored user id.
   `IngestPipelineIngestTest.aUserIdFieldInTheBodyDoesNotChangeTheStoredUserId` passes. I then
   changed `IngestPipeline.java:76` to `store.append(event, "body-user")`. The test failed. I
   restored the file.
2. **PASS** — Without a user id and with the flag off, the store gets 0 events.
   I replaced the condition at `IngestPipeline.java:72` with `if (false)`. The test
   `withoutAUserIdAndWithTheFlagOffTheStoreGetsZeroEvents` failed. I restored the file.
3. **PASS** — Without a user id and with the flag on, the store gets the event with `userId: null`.
   I replaced the condition at `IngestPipeline.java:72` with `if (userId == null)`. The test
   `withoutAUserIdAndWithTheFlagOnTheStoreGetsTheEventWithUserIdNull` failed. I restored the file.
4. **PASS** — `./gradlew :kit:jvm-core:dependencies --configuration runtimeClasspath` prints
   `runtimeClasspath - Runtime classpath of source set 'main'.` and `No dependencies`.
5. **PASS** — `./gradlew :kit:jvm-core:test` gives `BUILD SUCCESSFUL`. The XML reports hold 56
   tests, 0 skipped, 0 failures, 0 errors.
6. **PASS** — The comment of 2026-09-21 on issue #26 holds both outputs.

Each new test asserts a real value. No test is empty. No test is always green.

## Scope

- Only `kit/jvm-core` changes: 10 files, 352 added lines, 5 removed lines.
- No Kotlin file. `kit/jvm-core/build.gradle.kts` does not change, thus there is no new dependency.
- The public API of issue #10 stays as it was. `IngestPipeline.process` keeps its signature, and
  `IngestEvent` keeps its three components.
- The language of each new comment, each commit body, and the pull request text follows
  ASD-STE100. I found no violation.

## Findings

### MAJOR 1 — `kit/jvm-core/src/main/java/octometer/kit/core/store/EventLogStore.java:17`

`append` takes one event, thus a store cannot drop a batch. Design decision D21 and contract rule
C19 state that the store drops **the batch** above the event cap. Issue #34, step 3, repeats it:
"Above the cap, drop the batch in `append`". The present signature gives one event at a time, thus
issue #34 must change this interface. `IngestPipeline.ingest` also calls `append` one time for
each click. The MongoDB store of issue #12 then runs 50 inserts for one batch of 50 clicks. A
failure in the middle of the loop leaves a part of the batch in the store.

Fix: make the batch the unit of the method.

```java
void append(java.util.List<IngestEvent> events, String userId);

default void append(IngestEvent event, String userId) {
    append(java.util.List.of(event), userId);
}
```

In `IngestPipeline.ingest`, replace the loop with `store.append(events, userId);`. Add a test that
counts one call for a batch of two clicks.

### MAJOR 2 — `EventLogStore.java:12-17`

The failure rule of `append` is not defined. The Javadoc does not say what `append` does when the
store fails. The MongoDB store of issue #12 throws a driver exception. `IngestPipeline.ingest`
lets that exception pass, and an adapter of issue #11 has no rule for the HTTP status. Rule C19
names 204, 400, 415, and 429 only.

Fix: add to the Javadoc of `append`: "A store throws an unchecked exception when the write fails."
Add to the Javadoc of `IngestPipeline.ingest`: "A store exception passes to the caller. An adapter
maps it to status 500."

### MAJOR 3 — `kit/jvm-core/src/main/java/octometer/kit/core/ingest/IngestPipeline.java:71`

A broken `UserIdResolver` gives status 400 to the client. `EventFieldValidator.validateUserId`
throws `IngestException` with the reason `USER_ID_LENGTH`. The Javadoc of `IngestException` says
"A rejected ingest request", thus an adapter maps each `IngestException` to 400 (rule C19). The
user id comes from the app and not from the client (rule C16). A client with a valid body then
gets a 400 for a fault of the app.

Fix: throw a different unchecked exception for the resolved user id.

```java
String userId = userIdResolver.resolve();
try {
    EventFieldValidator.validateUserId(userId);
} catch (IngestException cause) {
    throw new IllegalStateException(
            "The UserIdResolver returned a userId that breaks rule C6.", cause);
}
```

State in the Javadoc of `ingest` that an adapter maps this exception to status 500. Also rename
the test at `IngestPipelineIngestTest.java:76`. Its name holds "above the length limit", but the
resolver at line 82 returns an empty string.

### MAJOR 4 — `EventLogStore.java:19-22` and `InMemoryEventLogStore.java:25-27`

`deleteByUserId(null)` removes each anonymous event. The predicate
`Objects.equals(storedEvent.userId(), userId)` matches each event with a `null` user id. Issue #35
uses this method to remove the data of one user. A `null` value from a caller then removes the
anonymous events of every user, and the method gives no signal.

Fix: state the rule in the interface Javadoc: "The value of `userId` must not be `null`. The
method throws `NullPointerException` for a `null` value." Add
`Objects.requireNonNull(userId, "userId must not be null");` as the first line of
`InMemoryEventLogStore.deleteByUserId`. Add a test for the `null` value.

### MAJOR 5 — `kit/jvm-core/src/main/java/octometer/kit/core/ingest/IngestSettings.java:15-17`

`fromEnvironment()` has no test. `IngestSettingsTest` tests the record constructor only. The one
line that reads `OCTOMETER_RECORD_ANONYMOUS` has no test. A test does not find a typo in the name
of the variable, or a change of the accepted value. The default off also has no test.

Fix: add a package-private method that takes the raw value, then test that method.

```java
public static IngestSettings fromEnvironment() {
    return fromValue(System.getenv("OCTOMETER_RECORD_ANONYMOUS"));
}

static IngestSettings fromValue(String rawValue) {
    return new IngestSettings("true".equalsIgnoreCase(rawValue));
}
```

Test these values: `null` gives off, `"true"` gives on, `"TRUE"` gives on, `"false"` gives off,
`"1"` gives off, `"yes"` gives off, `" true "` gives off. The last row needs one sentence in the
Javadoc, because `fromValue` does not trim the value.

### MAJOR 6 — `InMemoryEventLogStore.java:10-12`

The statement about issue #34 is wrong, and the store grows without a limit. I read issue #34.
Each step of #34 names `kit/jvm-mongo` or `kit/jvm-ktor`. Step 1 says "In `kit/jvm-mongo`, read
`OCTOMETER_MAX_EVENTS`". Issue #34 does not touch the in-memory store. The class comment and the
pull request text state that #34 owns the cap of this store. That statement is wrong, and the next
agent will trust it. This class is public API, thus an app can use it in production, and the
process can fill its memory.

Fix, one of the two:

- Change the class comment: "This store is for a test and for a demo. It holds each event until
  the process stops. An app in production uses the MongoDB store of issue #12." Remove the
  sentence about issue #34. Correct the same statement in the pull request text.
- Or give the constructor a maximum count, with the same rule in the Javadoc.

Then open a backlog issue for the limit of this store, or add the in-memory store to issue #34.

## MINOR findings

1. `InMemoryEventLogStore.java:33` — `List.copyOf(new ArrayList<>(events))` copies two times.
   `List.copyOf(events)` gives the same result with one copy.
2. `IngestEvent.java:6-7` — the tag `{@link IngestPipeline` + newline + `#ingest}` holds a space.
   I ran `javadoc` on the module. The output holds
   `<a href="IngestPipeline.html"><code>#ingest</code></a>`. The link points to the class, and the
   text is `#ingest`. Write `{@link IngestPipeline#ingest ingest}` on one line.
3. `InMemoryEventLogStoreTest.java` — there is no test of `append` and `deleteByUserId` at the
   same time. I ran my own check: one append thread, one delete thread, one read thread, 20
   rounds. I saw no exception, and a final delete left no matching event. The behaviour is
   correct. The test is absent. Note that `removeIf` of `ConcurrentLinkedQueue` is weakly
   consistent, thus an append during a delete can survive the delete.
4. `IngestPipeline.java:67` — `ingest` is a static method with five parameters. An instance with a
   constructor `(Clock, UserIdResolver, EventLogStore, IngestSettings)` and a method
   `ingest(String rawBody)` suits the dependency injection of issue #11 better. `process` can stay
   static.
5. `IngestPipeline.java:67` — the method does not check its parameters. A `null` store gives a
   `NullPointerException` with no message.
6. `UserIdResolver.java:5-7` — the Javadoc does not say when the app creates the implementation. A
   Ktor adapter cannot read the current call from a thread-local value. Add one sentence: "An
   adapter can give a new implementation for each request, for example a lambda that reads the
   current call."

## Notes, not findings

**The API questions.**

- `IngestPipeline.ingest` returns `void`. That is enough for rule C19. The server returns 204 for
  a success and for a batch that the cap removes. An `IngestException` gives 400. The adapter
  gives 415 and 429 from its own rules. A return value with the count of the stored events would
  help a log line and a metric, but rule C19 does not need it.
- The failure rule of `append` is absent. See MAJOR 2.
- The names and the packages are clear and stable: `octometer.kit.core.store`,
  `octometer.kit.core.user`, `StoredEvent`, `IngestSettings`.
- The read method `events()` is on the concrete class and not on the interface. That choice is
  correct, because the monitor reads from MongoDB and not from the interface.

**The two new optional fields (`path`, and `referrerHost` on a session start entry).**

The present form takes them. `append` passes the `IngestEvent` object, thus a new component
reaches each store with no change of `EventLogStore`. Inside the kit, three places need an edit:
the record `IngestEvent`, the record `StoredEvent`, and the factory `StoredEvent.of`. Two points
for the later design:

- A new component of the public record `IngestEvent` breaks each caller of the canonical
  constructor. Each test builds an `IngestEvent` in that way.
- `StoredEvent` copies the three fields of `IngestEvent`. The form
  `record StoredEvent(IngestEvent event, String userId)` takes a new event field with no edit.

This is a note only. The present form does not make that change hard.

xamcross:
Correction round 1. Commit `ff99e55` fixes each MAJOR finding and each MINOR finding, or
states the reason for a rejection below. `./gradlew :kit:jvm-core:test` gives `BUILD
SUCCESSFUL`, 69 tests, 0 failures, 0 errors. `./gradlew :kit:jvm-core:dependencies
--configuration runtimeClasspath` still shows `No dependencies`. Both commands ran on
2026-09-21 with JDK 21 at `C:/Users/xamcr/.jdks/ms-21.0.9`.

## MAJOR findings

**Java MAJOR 1, `EventLogStore.java:17` — the batch.** Fixed.
`EventLogStore.append(List<IngestEvent> events, String userId)` is now the interface method
(`kit/jvm-core/src/main/java/octometer/kit/core/store/EventLogStore.java:24`), with a default
method for one event at line 30. `IngestPipeline.ingest` calls `store.append(events, userId)`
one time
(`kit/jvm-core/src/main/java/octometer/kit/core/ingest/IngestPipeline.java:99`). New test:
`IngestPipelineIngestTest.ingestGivesTheWholeBatchToAppendInOneCall`, which counts the calls
of a fake store with a 2-click body.

**Java MAJOR 2, `EventLogStore.java:12-17` — the failure rule.** Fixed. The Javadoc of
`append` and of `deleteByUserId` now states: "A store throws an unchecked exception... An
adapter maps that exception to status 500"
(`EventLogStore.java:21-22` and `:42-43`). `IngestPipeline.ingest` states the same rule at
`IngestPipeline.java:74-76`.

**Java MAJOR 3 / Security MAJOR 3, `IngestPipeline.java:71` — the resolver's own defect.**
Fixed. `ingest` now wraps a broken `EventFieldValidator.validateUserId` result in
`IllegalStateException`, never `IngestException`
(`IngestPipeline.java:90-95`). The message names the rule and holds no user id. The Javadoc
states that an adapter maps this exception, and an exception of the resolver itself, to
status 500 (`IngestPipeline.java:67-72`). I renamed the test at
`IngestPipelineIngestTest.java` from `aUserIdAboveTheLengthLimitMakesTheRequestInvalid` to
`anEmptyResolvedUserIdMakesTheRequestInvalid`, and added
`aResolvedUserIdOf255CharactersMakesTheRequestInvalid`. Both now assert
`IllegalStateException`.

**Java MAJOR 4 / Security MAJOR 2, `deleteByUserId(null)`.** Fixed.
`InMemoryEventLogStore.deleteByUserId` (`InMemoryEventLogStore.java:30-36`) rejects `null`
with `Objects.requireNonNull`, and rejects an empty text with `IllegalArgumentException`, per
the maintainer's decision. The interface Javadoc states the rule
(`EventLogStore.java:35-40`). New tests: `deleteByUserIdRejectsANullValue` and
`deleteByUserIdRejectsAnEmptyText` in `InMemoryEventLogStoreTest.java`.

**Java MAJOR 5, `IngestSettings.java:15-17` — test coverage of the flag.** Fixed.
`IngestSettings.fromValue(String)` is a new package-private method
(`IngestSettings.java:26-28`); `fromEnvironment` calls it. Test
`IngestSettingsTest.fromValueAcceptsOnlyTheExactTextTrue` checks `null`, `"true"`, `"TRUE"`,
`"false"`, `"1"`, `"yes"`, and `" true "`, per the maintainer's decision. Only the exact text
`true` turns the flag on; this is a stricter rule than the earlier case-insensitive check.

**Java MAJOR 6 / Security MAJOR 1, `InMemoryEventLogStore.java:10-12` — the wrong claim.**
Fixed. The class comment now states: "Use this store only in a test and in a demo... An app
in production uses the MongoDB store of issue #11" (`InMemoryEventLogStore.java:8-15`). I
used issue #11 for the MongoDB store, not #12: issue #11 is "Build the MongoDB event store in
kit/jvm-mongo"; issue #12 is the Ktor route. The pull request text no longer claims that
issue #34 caps this store.

**Security MAJOR 4, `StoredEvent.toString()`.** Fixed. `StoredEvent` overrides `toString()`
(`kit/jvm-core/src/main/java/octometer/kit/core/store/StoredEvent.java:25-29`) and holds
neither the user id nor the session id. New test:
`StoredEventTest.toStringHoldsNoUserIdAndNoSessionId`.

## MINOR findings

**Java MINOR 1, double copy.** Fixed: `events()` is now `List.copyOf(storedEvents)`
(`InMemoryEventLogStore.java:42`).

**Java MINOR 2, the `{@link}` line break.** Fixed:
`{@link IngestPipeline#ingest ingest}` on one line (`IngestEvent.java:7`).

**Java MINOR 3, no append+delete concurrency test.** Fixed: new test
`InMemoryEventLogStoreTest.staysCorrectWithConcurrentAppendAndDelete`. It runs 4 append
threads and 1 delete thread together, then calls `deleteByUserId` once more after the
threads finish, because `removeIf` of `ConcurrentLinkedQueue` is weakly consistent, as the
review states.

**Java MINOR 4, an instance instead of a static method.** Rejected. `process` and `ingest`
stay static. Evidence: `process` is already static and public, from issue #10, and the
class stays a stateless utility (`private IngestPipeline()`, `IngestPipeline.java:19-22`).
Issues #11 and #12 do not exist yet, so a dependency-injection shape now is a guess. This
choice suits a later issue that wires the adapter, and it can change `ingest` with no change
of `process` or of the two interfaces.

**Java MINOR 5, no null check of the parameters.** Fixed: `ingest` now calls
`Objects.requireNonNull` for each of its five parameters, with a message
(`IngestPipeline.java:82-86`).

**Java MINOR 6, `UserIdResolver` Javadoc.** Fixed: added "An adapter can give a new
implementation for each request, for example a lambda that reads the current call."
(`UserIdResolver.java:5-7`).

**Security MINOR 1, the test name and the boundary test.** Fixed together with MAJOR 3
above: renamed test, and added the 255-character boundary test.

**Security MINOR 2, no character rule for `userId`.** Fixed.
`EventFieldValidator.validateUserId` now rejects a character below U+0020 and the delete
character U+007F (`EventFieldValidator.java:70-76`). New enum value
`IngestException.Reason.USER_ID_CHARACTER` (`IngestException.java:37-38`). New tests in
`EventFieldValidatorTest.java`: `rejectsAUserIdWithAControlCharacter`,
`rejectsAUserIdWithTheDeleteCharacter`, `acceptsAUserIdWithThePrintableBoundaryCharacters`.

**Security MINOR 3, trim and a warning for an unrecognised flag value.** Partly fixed, partly
rejected. Fixed: the maintainer's decision makes the match exact (`"true".equals(rawValue)`,
`IngestSettings.java:27`), so a leading space, a trailing space, and a different case all
give the flag off. This removes the ambiguity that the review names, with no need to read
the value twice. Rejected: a warning line. `kit/jvm-core` keeps zero runtime dependencies
(design decision D17) and holds no logging class of its own; a `System.err` line would be a
new, uncontrolled side effect with no test and no precedent in this module. The exact-match
rule already gives a predictable, fail-closed result for each value that the review names.

**Security MINOR 4, the resolver's own exception.** Fixed: `ingest`'s Javadoc now states
"This method also lets an exception of `userIdResolver` itself pass to the caller."
(`IngestPipeline.java:70-72`).

**Security MINOR 5, more body shapes for acceptance criterion 1.** Fixed: added
`aUserIdFieldInsideAClickEntryDoesNotChangeTheStoredUserId` and
`aUserIdFieldWithADifferentLetterCaseDoesNotChangeTheStoredUserId` to
`IngestPipelineIngestTest.java`.

## Notes I did not act on

Both reviews' "Notes, not findings" sections (the return value of `ingest`, the two optional
future fields) ask no fix. I made no change for them.

xamcross:
**Review (second review, security and privacy engineer): APPROVE**

I made a new clone and I checked out the head `ff99e55`. I judged each finding against the
code, not against the response comment. I wrote 21 probe tests of my own, and I ran two small
programs with a real process environment. My probes stay in my clone.

`./gradlew :kit:jvm-core:test` with JDK 21: `BUILD SUCCESSFUL`. The XML reports give 90 tests,
0 failures, 0 errors: 69 tests of the pull request and 21 probes of mine.

## The four MAJOR findings of my first review

**MAJOR 1 (no cap, and the wrong claim on issue #34): CORRECTED.**
`InMemoryEventLogStore.java:8-15` now says: "Use this store only in a test and in a demo. It
holds each event in the heap, with no size limit, and it loses each event at a restart. An app
in production uses the MongoDB store of issue #11." The class comment names no cap from issue
#34. The pull request text states that issue #34 does not touch this store. The reference to
issue #11 is right; issue #11 builds the MongoDB store.

**MAJOR 2 (`deleteByUserId(null)` erased each anonymous event): CORRECTED.**
`InMemoryEventLogStore.java:31-34` throws `NullPointerException` for `null` and
`IllegalArgumentException` for an empty text. `EventLogStore.java:35-40` states the rule for
each implementation, thus issue #35 inherits it. My probes prove the result:

- `deleteByUserId(null)` throws, and all 4 events stay, the 2 anonymous events included.
- `deleteByUserId("")` throws, and all 4 events stay.
- No accepted value selects an anonymous event. I tried ` `, `null`, `NULL`, `anonymous`, `*`,
  `%`, `.*`, and `0`. Each call left both anonymous events in place. A stored user id is never
  `null` and never an empty text at the same time as an accepted argument, thus no call can
  erase the anonymous events of each visitor.
- `deleteByUserId("user-1")` still keeps `user-10`, `user-1x`, and `USER-1`.
- One delete thread and one append thread of the same user ran together with no exception. The
  2000 events of the other user and the 2000 anonymous events all stayed.

**MAJOR 3 (a fault of the app gave status 400): CORRECTED.**
`IngestPipeline.java:90-95` catches the `IngestException` of `validateUserId`, then throws
`IllegalStateException`. The comment on lines 67-72 tells the adapter to map it to status 500,
never to 400. The check on line 91 runs before the anonymous branch on line 96, thus a bad
value never becomes an anonymous event. My probes prove it for an empty text, 255 characters,
a carriage return with a line feed, U+0000, U+007F, and an escape character, and with the flag
on and with the flag off:

- The call throws `IllegalStateException`, with the `IngestException` as the cause.
- A store that records each call got zero calls. The event reaches no store.
- I printed the whole stack trace of the chain. It holds neither the user id nor the session
  id. Each message in the module is a constant text; no message uses a concatenation or a
  format.

**MAJOR 4 (`StoredEvent.toString` leaked the user id): CORRECTED.**
`StoredEvent.java:25-29` overrides `toString`. The text holds `<redacted>` in place of a user
id, and it holds no session id. A list of stored events prints the same redacted text, because
`List.toString` calls this method for each element. An anonymous event still prints
`userId=null`, which is correct and holds no personal data. Test
`StoredEventTest.toStringHoldsNoUserIdAndNoSessionId` checks both values.

I checked each other type for the same leak, per point 3 of the task:

- `IngestSettings` is a record with one `boolean`. Its text holds no personal data.
- `ParsedClick` and `ParsedIngestRequest` are package-private. They never leave the module.
- `EventLogStore` and `UserIdResolver` are interfaces.
- `IngestEvent` keeps the generated `toString`. See MINOR 6 below.

## The five MINOR findings of my first review

- MINOR 1 (the wrong test name, and no boundary test): CORRECTED.
  `IngestPipelineIngestTest.java:78` is now `anEmptyResolvedUserIdMakesTheRequestInvalid`, and
  line 97 adds `aResolvedUserIdOf255CharactersMakesTheRequestInvalid`.
- MINOR 2 (no character rule): CORRECTED. `EventFieldValidator.java:70-76` rejects each
  character below U+0020 and the character U+007F, with the new reason `USER_ID_CHARACTER`. My
  probes confirm the four hostile values above.
- MINOR 3 (a trim and a warning for the flag): CORRECTED in substance. The exact-match rule
  removes the ambiguity that I named. I accept the refusal of a warning line: `kit/jvm-core`
  keeps zero runtime dependencies (design decision D17), and a `System.err` line has no test
  and no precedent here.
- MINOR 4 (the exception of the resolver): CORRECTED. `IngestPipeline.java:70-72` states the
  rule. My probe confirms that the exception passes to the caller, and that the store gets no
  call.
- MINOR 5 (more body shapes): CORRECTED. `IngestPipelineIngestTest.java:112` and `:126` add
  the click entry and the letter case.

## Point 5 of the task: the new batch API and the trust boundary

I ran seven body shapes against `append(List<IngestEvent>, String)`. The user id of the body
still has no path to the store. The parser extracts only `sessionId`, `element`, and `ageMs`,
and `ParsedIngestRequest` holds no user id field.

| The shape in the body | The result |
|---|---|
| `"userId"` at the top level | The store gets the value of the resolver. |
| `"userId"` inside a click entry | The store gets the value of the resolver. |
| `"UserId"`, `"USERID"`, `"userid"`, `"UserID"` together | The store gets the value of the resolver. |
| A duplicate `"userId"` key | `DUPLICATE_FIELD` (rule C36). The store gets zero calls. |
| A nested object, two levels deep, with an array | The store gets the value of the resolver. |
| `"userId": null` with `"user_id": 42` | The store gets the value of the resolver. |
| 50 click entries, each with its own `"userId"` | One call: `append(50)` with the resolver's user id. |

With the flag off and no user id, a store that records each call got zero calls. Nothing
reaches the store, and the module writes no log line.

## Point 6 of the task: the flag

I ran a program with a real process environment and 13 values. Only the exact text `true`
turns the flag on. `TRUE`, `True`, ` true`, `true `, `1`, `yes`, `on`, `enabled`, `false`,
`0`, an empty text, and an unset variable all give the flag off. `IngestSettings.java:26-28`
holds the one rule, and `fromEnvironment` calls it. The rule matches D19, and each wrong value
fails closed.

## The scope

The diff touches 14 files, each one in `kit/jvm-core`. It changes no build file, thus no new
dependency enters. It holds no secret, no database file, and no build output.

## One new MINOR finding

**MINOR 6 — `kit/jvm-core/src/main/java/octometer/kit/core/ingest/IngestEvent.java:12`.**
`IngestEvent` keeps the generated `toString`, thus it prints the session id:
`[IngestEvent[sessionId=3fa85f64-..., element=a.b, ts=...]]`. `append` now takes a
`List<IngestEvent>`, so one debug line with the batch prints each session id. D15 names only a
user id and a connection string, thus this breaks no rule today. But the new `StoredEvent`
hides the session id, and this type shows it. A later issue can give `IngestEvent` the same
`toString`, for one rule across the module. This finding does not prevent the approval.

## The verdict

All 4 MAJOR findings and all 5 MINOR findings of my first review are corrected. 1 new MINOR
finding stays open. The trust boundary of the user id holds against each shape that I tried,
the erasure path can no longer destroy the anonymous log, a fault of the app now maps to 500,
and no text form of a new type leaks a user id. I approve this pull request.

xamcross:
**Review (second review, Java API engineer): CHANGES REQUESTED**

I made a new clone and I checked the code, not the response comment. Branch
`issue-26-event-log-store`, head `ff99e55`. JDK 21 at `C:/Users/xamcr/.jdks/ms-21.0.9`.

`./gradlew :kit:jvm-core:test` gives `BUILD SUCCESSFUL`. The XML reports hold 69 tests, 0 skipped,
0 failures, 0 errors. `./gradlew :kit:jvm-core:dependencies --configuration runtimeClasspath`
prints `No dependencies`. Commit `ff99e55` touches 14 files, each one in `kit/jvm-core`. There is
no Kotlin file, no change of a build file, no secret, and no build output.

Each of my 6 MAJOR findings is corrected. Two new MAJOR findings come from the correction itself.

| Finding | Result |
|---|---|
| MAJOR 1, the batch in `append` | **CORRECTED** |
| MAJOR 2, the failure rule of `append` | **CORRECTED** |
| MAJOR 3, `IllegalStateException` for a bad resolver value | **CORRECTED** |
| MAJOR 4, `deleteByUserId` with `null` | **CORRECTED** |
| MAJOR 5, the test of the flag | **CORRECTED** |
| MAJOR 6, the wrong claim about issue #34 | **CORRECTED** |
| MINOR 1, the double copy | **CORRECTED** |
| MINOR 2, the `{@link}` line break | **CORRECTED** |
| MINOR 3, the append and delete test | **CORRECTED** |
| MINOR 4, a static method | **REJECTED, and I accept the reason** |
| MINOR 5, the null check of each parameter | **CORRECTED** |
| MINOR 6, the `UserIdResolver` Javadoc | **CORRECTED** |

**My evidence for the three main criteria.** I broke the code three times and I restored the file
each time.

- `IngestPipeline.java:99` to `store.append(events, "body-user")`: 4 tests failed, among them
  `aUserIdFieldInTheBodyDoesNotChangeTheStoredUserId`.
- `IngestPipeline.java:96` to `if (userId == null && settings == null)`:
  `withoutAUserIdAndWithTheFlagOffTheStoreGetsZeroEvents` failed.
- `IngestPipeline.java:96` to `if (userId == null)`:
  `withoutAUserIdAndWithTheFlagOnTheStoreGetsTheEventWithUserIdNull` failed.
- `IngestSettings.java:27` to `"true".equalsIgnoreCase(rawValue)`:
  `fromValueAcceptsOnlyTheExactTextTrue` failed. The new flag test is real.

**MAJOR 3, my own run.** A resolver of 300 characters gives
`java.lang.IllegalStateException`. The message is "The UserIdResolver of the app returned a userId
that breaks contract rule C6." The cause message is "The userId value must have 1 to 254
characters." Neither text holds the user id. The store stays empty, because line 99 comes after
line 95.

**MAJOR 4, my own run.** `deleteByUserId(null)` throws `NullPointerException` and
`deleteByUserId("")` throws `IllegalArgumentException`. The two tests assert it, and a rejected
call leaves each anonymous event in place.

**MAJOR 6, a correction of my own text.** You are right about the issue numbers. Issue #11 is
"Build the MongoDB event store in kit/jvm-mongo" and issue #12 is "Build the Ktor ingest route in
kit/jvm-ktor". My first review named #12 for the MongoDB store. Your class comment names #11, and
that is correct.

## New findings

### MAJOR A — `IngestPipeline.java:99` and `EventLogStore.java:24`: an empty batch reaches `append`

An empty `clicks` array is a legal body. Rule C13 sets no minimum, and rule C17 sets the maximum
of 50. I ran `IngestPipeline.process` with
`{"sessionId": "3fa85f64-5717-4562-b3fc-2c963f66afa6", "clicks": []}`. It returns a list of 0
events. I then ran `ingest` with a fake store. The output is `append called with size 0`.

The one-call change of MAJOR 1 introduced this path. The loop of the first commit called the store
0 times for an empty body. The MongoDB store of issue #11 will call `insertMany`. That driver
method throws `IllegalArgumentException` for an empty list, thus a legal request from a client
gives status 500. Rule C19 asks for status 204.

Fix: stop before the call, and state the rule.

```java
        List<IngestEvent> events = process(rawBody, clock);
        if (events.isEmpty()) {
            return;
        }
```

Add this sentence to the Javadoc of `EventLogStore.append`: "The list holds one event or more. The
pipeline never calls this method with an empty list." Add a test that gives a body with an empty
`clicks` array to a fake store, then asserts 0 calls of `append`.

### MAJOR B — `InMemoryEventLogStore.java:21-27`: the store breaks the promise of its own interface

`EventLogStore.java:17` states: "The batch is the unit of the write". The in-memory store adds one
event at a time in a loop, and it checks each element inside that loop. Two results follow:

- A batch that the store rejects leaves a part of itself in the store. I ran `append` with a list
  of three events, and the third one is `null`. The call throws `NullPointerException`, and
  `events()` then returns 2 events.
- A reader of `events()` can see one half of a batch, because each `add` is its own linearization
  point.

Fix: build the whole batch first, then add it with one call.

```java
    @Override
    public void append(List<IngestEvent> events, String userId) {
        Objects.requireNonNull(events, "events must not be null");
        List<StoredEvent> batch = new ArrayList<>(events.size());
        for (IngestEvent event : events) {
            Objects.requireNonNull(event, "event must not be null");
            batch.add(StoredEvent.of(event, userId));
        }
        storedEvents.addAll(batch);
    }
```

`ConcurrentLinkedQueue.addAll` builds a private chain of nodes, then links that chain with one
compare-and-set. The JDK source names that operation "the linearization point for all elements to
be added to this queue". A reader then sees the whole batch or none of it. Add a test: a batch with
a `null` element leaves the store empty.

A second form of this fix is also correct: change `EventLogStore.java:17` to state that the batch
is the unit of the **call** and not of the write, and that a store can hold a part of a batch after
a failure. I prefer the code fix, because issue #11 needs a clear rule for a retry.

## MINOR findings

1. **`IngestSettings.java:26-28`, the silent flag value.** I accept the exact-match rule against
   decision D19. D19 writes `OCTOMETER_RECORD_ANONYMOUS=true`, and the strict rule is fail-closed
   for privacy. The silence is the problem: an operator who writes `TRUE` in a Windows environment
   variable gets the flag off, and no text tells that operator why.
   I reject your reason for the rejection. `kit/jvm-core` does have a log facility with no
   dependency: `java.lang.System.Logger` is in `java.base` since Java 9. It routes each line to
   the log system of the app through a `ServiceLoader`, and
   `./gradlew :kit:jvm-core:dependencies --configuration runtimeClasspath` still prints
   `No dependencies`. Decision D17 forbids a dependency; it does not forbid a log line. One line is
   enough:
   ```java
   if (rawValue != null && !"true".equals(rawValue) && "true".equals(rawValue.trim().toLowerCase(Locale.ROOT))) {
       System.getLogger("octometer.kit.core").log(System.Logger.Level.WARNING,
               "OCTOMETER_RECORD_ANONYMOUS holds a value that is not the exact text true. The flag stays off.");
   }
   ```
   This stays MINOR. It does not prevent an approval.
2. **`IngestPipeline.java:19-22`, the static method.** I accept your reason. `process` is already
   static and public from issue #10, the class holds no state, and a later issue can add an
   instance form with no change of `process` and no change of the two interfaces.
3. **`EventLogStore.java:13-24`.** The Javadoc does not say what a store can do with the list of
   the caller. I ran the check: `InMemoryEventLogStore` leaves the caller's list as it was. Add one
   sentence: "A store reads the list. It never changes the list, and it never keeps a reference to
   the list after the call."
4. **`EventLogStore.java:39-40`, the language.** "An anonymous event has no erasure" uses a noun
   where a verb is clearer. ASD-STE100 prefers the simple verb. Write: "A store never erases an
   anonymous event with this method, because that event holds no user id."
5. **`IngestException.java:36-38`.** The Javadoc of `IngestException` says "A rejected ingest
   request", thus an adapter maps each reason to status 400. `USER_ID_LENGTH` and the new
   `USER_ID_CHARACTER` never leave `ingest` in that class now; `ingest` wraps them in
   `IllegalStateException`. Add one sentence to the class Javadoc: "The two `USER_ID_` reasons come
   from the app and not from the client. `IngestPipeline.ingest` wraps them in
   `IllegalStateException`."

## The rest of the check

- The language of each new comment, of the commit body, and of the pull request text follows
  ASD-STE100, with the one exception of MINOR 4 above.
- The public API of issue #10 stays as it was. `IngestPipeline.process` keeps its signature, and
  `IngestEvent` keeps its three components.
- The new character rule of `EventFieldValidator.validateUserId` is stricter than rule C6. The
  Javadoc states that plainly, thus the rule is clear. I accept it.
- `fromEnvironment()` still has no test of its own. That is correct: it is one line that calls
  `fromValue`, and `fromValue` holds each rule and each test.

Two MAJOR findings remain, thus I cannot approve this round. Both fixes are small.

xamcross:
Correction round 2. Commit `a2adb33`. `./gradlew :kit:jvm-core:test` gives `BUILD SUCCESSFUL`,
74 tests, 0 failures. `./gradlew :kit:jvm-core:dependencies --configuration runtimeClasspath`
still shows `No dependencies`. Both ran on 2026-09-21 with JDK 21.

**MAJOR A, the empty batch.** Fixed. `IngestPipeline.ingest` returns after `process` when the
list is empty, before it resolves the user id
(`kit/jvm-core/src/main/java/octometer/kit/core/ingest/IngestPipeline.java:89-91`). The
Javadoc of `ingest` and of `EventLogStore.append` states the rule
(`IngestPipeline.java:74-76`, `EventLogStore.java:20-22`). New test:
`IngestPipelineIngestTest.anEmptyClicksArrayGivesZeroCallsOfAppend`.

**MAJOR B, the partial batch.** Fixed. `InMemoryEventLogStore.append` builds a `List<StoredEvent>`
first, then calls `storedEvents.addAll(batch)` one time
(`kit/jvm-core/src/main/java/octometer/kit/core/store/InMemoryEventLogStore.java:20-32`). New
test: `InMemoryEventLogStoreTest.aRejectedBatchLeavesNoPartOfItselfInTheStore`, a batch of 3
with a `null` third element.

**MINOR, the silent flag value.** Fixed, per the decision. `IngestSettings.fromValue` writes
one warning through `System.Logger` when the value is neither `null`, nor the exact text
`true`, nor the exact text `false`
(`kit/jvm-core/src/main/java/octometer/kit/core/ingest/IngestSettings.java:34-39`). Two new
tests read the warning through a test-only `LoggerFinder` (`CapturingLoggerFinder.java`, a
`META-INF/services` file, no new dependency): `fromValueWarnsForAValueThatIsNeitherTrueNorFalseWithNoRepeatOfTheValue`
and `fromValueGivesNoWarningForTheExactTextsTrueAndFalseAndForANullValue`. The dependency
command still shows `No dependencies`.

**MINOR, `IngestEvent.toString()`.** Fixed. It now reads
`IngestEvent[element=..., ts=...]`, with no session id
(`kit/jvm-core/src/main/java/octometer/kit/core/ingest/IngestEvent.java:14-19`). New test:
`IngestEventTest.toStringHoldsNoSessionId`. No test of issue #10 reads the old text; I
searched the test sources for a call of `IngestEvent`'s `toString` and found none.

**Java MINOR 3, the list rule of `append`.** Fixed: "A store reads the list. It never changes
the list, and it never keeps a reference to the list after the call."
(`EventLogStore.java:26-27`).

**Java MINOR 4, the wording.** Fixed: "A store never erases an anonymous event with this
method, because that event holds no user id." (`EventLogStore.java:39-40`).

**Java MINOR 5, the two `USER_ID_` reasons.** Fixed: added to the class Javadoc of
`IngestException`: "The two `USER_ID_` reasons come from the app and not from the client.
`IngestPipeline.ingest` wraps them in `IllegalStateException`."
(`kit/jvm-core/src/main/java/octometer/kit/core/ingest/IngestException.java:4-9`).

No finding rejected this round.

xamcross:
**Review (third review, Java API engineer): APPROVE**

I made a new clone and I checked the code. Branch `issue-26-event-log-store`, head `a2adb33`.
JDK 21 at `C:/Users/xamcr/.jdks/ms-21.0.9`. `./gradlew :kit:jvm-core:test` gives `BUILD
SUCCESSFUL`; the XML reports hold 74 tests, 0 skipped, 0 failures, 0 errors.
`./gradlew :kit:jvm-core:dependencies --configuration runtimeClasspath` prints `No dependencies`.

**MAJOR A — CORRECTED.** `IngestPipeline.java:93-95` returns before the store call. I ran my own
probe with the body `{"sessionId": "3fa85f64-5717-4562-b3fc-2c963f66afa6", "clicks": []}` and a
fake store. The result is `NO CALL`. The Javadoc of `EventLogStore.append` states the minimum:
"The list holds one event or more." The test
`IngestPipelineIngestTest.anEmptyClicksArrayGivesZeroCallsOfAppend` proves it. I broke the guard
to `if (events.isEmpty() && rawBody.isEmpty())`; that test failed. I restored the file.

**MAJOR B — CORRECTED.** `InMemoryEventLogStore.java:22-34` builds the batch first, then calls
`storedEvents.addAll(batch)`. I ran my probe again: a batch of three events with a `null` third
element throws `NullPointerException`, and `events()` then returns **0 events**. I put the
`add` call back inside the loop; `aRejectedBatchLeavesNoPartOfItselfInTheStore` failed. I restored
the file.

**The warning — CORRECT.** `IngestSettings.java:39-43` uses `java.lang.System.Logger`. The text
names `OCTOMETER_RECORD_ANONYMOUS` and the rule, and it never repeats the value. The dependency
report still prints `No dependencies`, because `System.Logger` is a part of `java.base`. The two
tests check the warning and the silence for `true`, `false`, and `null`.

**The test-only `LoggerFinder` — CORRECT.** `CapturingLoggerFinder.java` sits in
`kit/jvm-core/src/test/java/`, and the service file sits in
`kit/jvm-core/src/test/resources/META-INF/services/`. I built the jar and listed it: it holds 14
main classes and `META-INF/MANIFEST.MF` only. Neither the finder class nor a `services` entry
reaches the jar.

**`IngestEvent.toString()` — CORRECT.** The text holds `element` and `ts` only.
`IngestEventTest.toStringHoldsNoSessionId` proves it. I searched each test of the module: no test
of issue #10 reads the text of `IngestEvent`. The record `equals` does not change, thus each
existing assertion stays valid.

**MINOR, no block.** `CapturingLoggerFinder.java:17` holds the messages in a static `ArrayDeque`,
which is not safe for more than one thread. The module runs its tests in one thread today, and I
found no parallel setting, thus the two flag tests are correct. A later issue that turns on the
parallel mode must change this field to a `ConcurrentLinkedDeque`.

Each of my earlier findings is now corrected. I approve this pull request.

Issue #36, opened 2026-09-21
Add the page lifecycle flush and the retry rule to the tracker


xamcross:
**Goal.** The tracker sends its queue before the browser hides or unloads the page. It sends a failed batch again one time.

**Implementation steps.**
1. In `kit/tracker`, add a `pagehide` listener that sends the queue with `fetch` and `keepalive: true`.
2. Add a `visibilitychange` listener that sends the queue in the same way when the document becomes hidden.
3. Send a batch again one time, after a delay: 500 ms plus a random value below 500 ms after a network error or a 5xx response, and 2000 ms plus a random value below 2000 ms after a 429 response. Drop a batch after each other 4xx response.
4. Send a batch only one time from the `pagehide` listener and from the `visibilitychange` listener. Only the timer flush sends a batch a second time.
5. Drop the oldest entry when a click enters a queue of 200 entries.
6. Measure the encoded body. Split a batch whose body passes 15 000 bytes, also below 50 entries.

**Acceptance criteria.**
- [x] Each rule of steps 1 to 6 has a test with fake timers.
- [x] A failed request throws no error into the page.
- [x] The sum of the request bodies of one lifecycle flush stays below 64 KB, also for a full queue of 200 entries at the maximum sizes of the contract. A test without a spy on the splitter asserts the sum, and a code comment names the invariant `MAX_QUEUE_SIZE / MAX_BATCH_ENTRIES * MAX_BODY_BYTES < 65536`.
- [x] No second send goes out after `stop()`.
- [x] A test with fake timers proves that the second send starts after the delay of step 3, and not before it.
- [x] No request body passes 15 000 bytes.
- [x] A batch whose encoded body passes 15 000 bytes goes in two requests, and each request stays below 15 000 bytes.
- [x] 50 entries with a 150-byte `path` each give a body below the 16 KB limit of rule C18.
- [x] `npm test` in `kit/tracker` passes.
- [x] A comment on this issue records the output of `npm test`.

**Related.** #106 adds `path` to each click entry, thus each entry grows.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, D24, and the rule T6.

xamcross:
Body edit on 2026-09-21: #13 closed. No open blocker remains, thus the first line and the label `blocked` are removed. The work can start.

xamcross:
[...]

Issue #44, opened 2026-09-21
Attach the tracker package to each GitHub Release


xamcross:
**Blocked by #39.** The package comes from the tracker of #13, and the step goes into the workflow of #39. Nothing can start now.

**Goal.** An app installs the tracker from a release URL, without an npm account and without a token.

**Implementation steps.**
1. Add a check to `.github/workflows/release.yml` that compares the tag with the version in `kit/tracker/package.json`.
2. Add a step that runs `npm pack` in `kit/tracker`. The result has the name `octometer-tracker-X.Y.Z.tgz`.
3. Attach the file to the GitHub Release.
4. Add the install line with the release URL to `kit/tracker/README.md`.

**Acceptance criteria.**
- [ ] The workflow fails when the version in `kit/tracker/package.json` differs from the tag.
- [ ] The release holds `octometer-tracker-X.Y.Z.tgz`.
- [ ] `npm install <tgz URL>` works in a clean folder, and `npm ci` works after it.
- [ ] In that folder, a Node script imports the package without an error.
- [ ] A comment on this issue records the release URL and the install line.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, D25, D26.

xamcross:
Body edit on 2026-09-21: #13 closed. The first line now names only the open blockers: #39.

PR #95, opened 2026-09-21
feat(tracker): add the click tracker core

### LINKED ISSUE #13: Build the click tracker core
**Goal.** The tracker records a click on a `data-octo` element and sends it to the ingest route.

**Implementation steps.**
1. Create the npm project `kit/tracker`: TypeScript, ESM, `"sideEffects": false`, zero dependencies.
2. Write `start()`, `stop()`, and the options of D24.
3. Keep the session id in `sessionStorage`, with the fallbacks of D24.
4. Add one capture-phase `click` listener that searches `event.composedPath()` for `data-octo` and skips a disabled element.
5. Add the queue of 200 entries with the one-shot `setTimeout` that sends it with `fetch`.
6. Add the `tracker` job (`npm ci`, `npm test`, `npm run build`) to `.github/workflows/ci.yml`.

**Acceptance criteria.** The tests use jsdom and fake timers.
- [x] A click on a child of `data-octo="save"` records `save`.
- [x] A click without the attribute records no entry.
- [x] The tracker records no click before `start()`.
- [x] `stop()` empties the queue and removes the `sessionStorage` key.
- [x] 120 clicks give requests with a maximum of 50 clicks each.
- [x] The body holds only `sessionId` and `clicks`.
- [x] `dependencies` in `package.json` is empty.
- [x] `npm test` passes in `kit/tracker`.
- [x] The `tracker` job is green.
- [x] A comment on this issue records the bundle size.

**Related.**
- #8 defines the format.
- #36 adds the lifecycle flush and the retry rule.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, D24, section 4.2.

xamcross:
## Summary

This pull request adds the npm project `kit/tracker`. The tracker records a
click on a `data-octo` element and sends it to the ingest route. It follows
design decision D24 and design section 4.2. It adds the `tracker` job to
`.github/workflows/ci.yml`.

## Acceptance criteria

- [x] A click on a child of `data-octo="save"` records `save`. Evidence: the
  test "records a click on a child of a data-octo element" in
  `kit/tracker/src/index.test.ts`.
- [x] A click without the attribute records no entry. Evidence: the test
  "records no entry for a click without the data-octo attribute".
- [x] The tracker records no click before `start()`. Evidence: the test
  "records no click before start()".
- [x] `stop()` empties the queue and removes the `sessionStorage` key.
  Evidence: the tests "stop() empties the queue, so a restart sends only
  the new click", "stop() clears the pending flush timer", and "stop()
  removes the sessionStorage key that the flush wrote". Correction round 1
  replaced one test that proved nothing with these three.
- [x] 120 clicks give requests with a maximum of 50 clicks each. Evidence:
  the test "gives requests with a maximum of 50 clicks each for 120
  clicks" (3 requests of 50, 50, and 20 clicks).
- [x] The body holds only `sessionId` and `clicks`. Evidence: the test
  "sends a body that holds only sessionId and clicks, each click only
  element and ageMs".
- [x] `dependencies` in `package.json` is empty. Evidence:
  `kit/tracker/package.json` shows `"dependencies": {}`.
- [x] `npm test` passes in `kit/tracker`. Evidence: `npm test` gives "Test
  Files 1 passed (1)" and "Tests 40 passed (40)".
- [x] The `tracker` job is green. Evidence: `gh pr checks 95`, after the
  correction-round push.
- [x] A comment on issue #13 records the bundle size. Evidence: three issue
  comments so far, one for each correction round. The current build gives
  6202 bytes raw, 1866 bytes gzip.

## Click 201

The queue holds a maximum of 200 entries. Click 201 drops click 1: the
tracker drops the oldest entry and keeps the newest 200 entries. The test
"drops the oldest entry when the queue holds 200 entries" proves this rule.

## sessionStorage fallback (step 3)

Each `sessionStorage` access has its own try/catch. A blocked store, or a
private window, gives no error to the caller:
- A failed read falls back to a new UUID.
- A failed write keeps the generated id in memory for the life of the
  tracker instance, with no persisted key.
- A failed removal on `stop()` does not throw.

The tests "falls back to an in-memory sessionId when sessionStorage.getItem
throws", "falls back to an in-memory sessionId when sessionStorage.setItem
throws", and "does not throw when sessionStorage.removeItem throws during
stop()" prove these three fallbacks.

## Out of scope

Issue #36 owns the page lifecycle flush (`pagehide`, `visibilitychange`),
the `keepalive` request option, and the retry rule after a network error, a
5xx, or a 429. This pull request adds only the one-shot `setTimeout` flush
of step 5.

## Confirm results

- The `monitor/frontend/.nvmrc` file holds `24.21.0`. The new
  `kit/tracker/.nvmrc` file holds the same value.
- The `.github/workflows/ci.yml` `frontend` job pins each action to the
  commit SHA shown in the file. The new `tracker` job uses the same pinned
  SHA for `actions/checkout` and `actions/setup-node`.

## Correction round 1

Two reviews requested changes. The response comment on this pull request
lists each finding with its fix, its file, and its line. In short:
- Replaced the false test for `stop()` with three tests that each fail
  under their own mutation.
- `ageMs` now uses a monotonic clock; it stays correct through a system
  clock jump.
- A throw from the app's `headers()` function, or from inside a flush, no
  longer reaches the page.
- Added `npm run typecheck` and its CI step; the test file now compiles
  under `strict` and `noUncheckedIndexedAccess`.
- The `fetch` call now sets `referrerPolicy: 'no-referrer'`.

## Correction round 2

The corrections of round 1 brought two new defects. The response comment
on this pull request lists each finding with its fix. In short:
- `stop()` now removes the `sessionStorage` key of an earlier page view
  too, with a new `everStarted` flag. It still touches no storage before a
  first `start()`.
- The `Content-Type` guard now drops each case of that header name, so a
  lower-case app header cannot merge into two header values.
- A failed flush now empties the queue first, so it drops only its own
  batch. A later click still starts a new timer.
- The build runs in two steps, so the shipped JS file drops each comment
  while the type declaration file keeps its doc comments.
- `package.json` `files` ships `src/index.ts` only, not the test file.

Closes #13

Generated with Claude Code


xamcross:
**Review (privacy and security engineer, first review): CHANGES REQUESTED**

I made my own clone, I checked out `issue-13-tracker-core` at `e52c3b9`, and I ran each command
with the portable Node 24.21.0. I also wrote my own test file with 16 tests in my clone
(`src/privacy.review.test.ts`, not committed). All 16 tests pass. The findings below cite that
evidence.

## Acceptance criteria of issue #13

| --- | --- | --- |
| A click on a child of `data-octo="save"` records `save` | PASS | `npm test`: `Tests 20 passed (20)` |
| A click without the attribute records no entry | PASS | `npm test`, same run |
| The tracker records no click before `start()` | PASS | My test "point 1b": no timer, and no request after 600000 ms |
| `stop()` empties the queue and removes the `sessionStorage` key | PASS for the code, FAIL for the proof | My test "point 1d" proves the removal. The test of the pull request cannot fail. See MAJOR 2 |
| 120 clicks give requests with a maximum of 50 clicks each | PASS | `npm test`: 3 requests of 50, 50, and 20 |
| The body holds only `sessionId` and `clicks` | PASS | My test "point 2" |
| `dependencies` in `package.json` is empty | PASS | `kit/tracker/package.json:26`; `npm ls --all` shows 115 development packages and 0 production packages |
| `npm test` passes in `kit/tracker` | PASS | `Test Files 1 passed (1)`, `Tests 20 passed (20)` |
| The `tracker` job is green | PASS | `gh pr checks 95`: `Tracker pass 17s` |
| A comment on issue #13 records the bundle size | PASS | My build gives `6256` bytes and `2088` bytes after gzip. The two numbers agree with the comment |

## The eight points of my role

1. **The gate.** PASS. `createTracker()` adds no listener, reads no storage, writes no storage,
   starts no timer, and sends no request (my test "point 1a", with a spy on
   `document.addEventListener` and on `Storage.prototype`). A click that is in the queue at the
   time of `stop()` never leaves the browser: `stop()` empties the queue and clears the timer
   (`src/index.ts:78-82`), and no request follows after 600000 ms (my test "point 1c"). After a
   flush and a `stop()`, the `sessionStorage` key is absent (my test "point 1d"). See MINOR 5 for
   a `stop()` call before `start()`.
2. **The body.** PASS. The body holds only `sessionId` and `clicks`. An entry holds only
   `element` and `ageMs` (`src/index.ts:140-152`, `dist/index.js:102-114`). My test "point 2"
   puts a text, a `value` attribute, and a card-like number on the element. The body holds none
   of them, and it holds no URL, no user agent, and no user id. The built file `dist/index.js`
   holds no `localStorage`, no `indexedDB`, no `document.cookie`, no `Math.random`, no
   `navigator`, no `location`, no `textContent`, no `clientX`, and no `console` call.
3. **The `data-octo` value.** PASS. The tracker checks the value against rule C4 before the
   entry enters the queue (`src/index.ts:98`). My test "point 3" clicks eight bad values: 101
   characters, an empty value, a value with a space, a value with markup, a value with a
   newline, a path value, a value with a null character, and a value with a quotation mark. No
   request follows. A value of exactly 100 characters passes.
4. **The session id.** PASS. `crypto.randomUUID()` runs first, and the fallback uses
   `crypto.getRandomValues()` with the correct version and variant bits (`src/index.ts:210-227`).
   `Math.random()` is absent. My tests "point 4a" and "point 4b" check both paths against the
   pattern of a version 4 UUID. The key name is `octo_session_id` in `sessionStorage` only. My
   test "point 4c" confirms an empty `localStorage`, an unchanged `document.cookie`, and one key
   in `sessionStorage`.
5. **The request.** PASS with two findings. The default of `credentials` is `same-origin`
   (`src/index.ts:43`). An absolute endpoint on a different origin gets the same default, thus
   the browser sends no cookie (my test "point 5a"). The `endpoint` option is mandatory in the
   type. The `headers` function of the app adds a CSRF header (my test "point 5b"). A failed
   request writes nothing to the console: my test "point 5d" spies on `log`, `info`, `warn`,
   `error`, and `debug`, and none of them runs. See MAJOR 1 for the referrer, and MINOR 2 for the
   header order.
6. **The supply chain.** PASS. `dependencies` is `{}`. The three development dependencies are
   `jsdom`, `typescript`, and `vitest`. Each one has a clear purpose, and each version agrees
   with `monitor/frontend/package.json`. The lock file is committed. All 115 package entries have
   an `integrity` value and a `resolved` value on `registry.npmjs.org`. The project has no
   `postinstall` script. `npm audit` reports `found 0 vulnerabilities`. Only `fsevents` has an
   install script; it is optional and limited to `darwin`. `dist/` stays in `.gitignore`, thus the
   diff holds no build output, and the built file holds no test code.
7. **The CI job `tracker`.** PASS. The job pins `actions/checkout` to `3d3c42e5aac5...` and
   `actions/setup-node` to `820762786026...`. The two SHA values are the same as in the job
   `frontend`. The job runs `npm ci`, it adds no permission, it uses no secret, and it sets
   `persist-credentials: false` like the other jobs. See MINOR 6 for the absent lint step.
8. **Secrets.** PASS. The diff, the pull request text, and the issue comments hold no secret and
   no text with the form of a credential. The job `Secret scan` passes.

## Findings

### MAJOR 1 — the beacon sends the full page URL in the `Referer` header

`kit/tracker/src/index.ts:148-153`. The `fetch` call sets no `referrerPolicy`. The default policy
is `strict-origin-when-cross-origin`. The endpoint is same-origin in the normal case, thus the
browser adds a `Referer` header with the full page URL, the query string included. The access log
of the app then holds the page URL next to the session id. The design keeps the URL out of the
event, and decision O5 keeps the IP address out too. A query string can hold personal data, for
example an email address. The tracker must not add this field.

Fix, in `sendBatch`:

```ts
    void fetch(endpoint, {
      method: 'POST',
      credentials,
      headers,
      referrerPolicy: 'no-referrer',
      body: JSON.stringify({ sessionId: id, clicks }),
    }).catch(() => {
```

Add one test that reads `init.referrerPolicy` and expects `'no-referrer'`.

### MAJOR 2 — the test of the `sessionStorage` key is always green

`kit/tracker/src/index.test.ts:131-146`, the assertion on line 145. The tracker creates the
session id on the first flush (`src/index.ts:126`). The test clicks one time and calls `stop()`
before the flush, thus the key never exists. My test "point 1e" proves this: the key is `null`
before `stop()` too. The assertion `expect(...).toBeNull()` passes with a `stop()` that removes
nothing. The criterion "removes the `sessionStorage` key" has no proof. The code is correct; only
the test is empty.

Fix, in that test body after `tracker.start()`:

```ts
    clickElement(host);
    vi.advanceTimersByTime(5000);
    expect(window.sessionStorage.getItem('octo_session_id')).not.toBeNull();
    fetchMock.mockClear();
    clickElement(host);

    tracker.stop();

    vi.advanceTimersByTime(60000);
    expect(fetchMock).not.toHaveBeenCalled();
    expect(window.sessionStorage.getItem('octo_session_id')).toBeNull();
```

The second click proves the queue part. The two storage assertions prove the key part.

### MINOR 1 — no test for the default `credentials`

`kit/tracker/src/index.test.ts:230-251`. The only test of the request options passes
`credentials: 'include'`. The privacy-critical default `same-origin` has no test. My test
"point 5a" proves the default. Please add the same check to the suite.

### MINOR 2 — the `headers` function of the app can replace `Content-Type`

`kit/tracker/src/index.ts:144-147`. `Object.assign` writes the app headers over the default
header. An app that returns `Content-Type` breaks contract rule C12. It also removes the CORS
preflight for a cross-origin endpoint. My test "point 5c" proves the override. Fix:

```ts
    const headers: Record<string, string> = {};
    if (getExtraHeaders) {
      Object.assign(headers, getExtraHeaders());
    }
    headers['Content-Type'] = 'application/json';
```

### MINOR 3 — `endpoint` has no check at run time

`kit/tracker/src/index.ts:52`. A JavaScript caller can pass no `endpoint`. `fetch(undefined)`
then posts the session id and the clicks to the path `undefined` of the current origin. Add a
check in `createTracker` that throws a `TypeError` for an empty `endpoint`.

### MINOR 4 — two calls have no guard, thus an error leaves the timer callback

`kit/tracker/src/index.ts:215` calls `crypto.getRandomValues` after the guard on
`crypto.randomUUID` fails. `src/index.ts:148` calls `fetch` with no guard. In an environment
without `crypto` or without `fetch`, the error leaves the `setTimeout` callback and reaches the
app. My test "point 5e" proves the `fetch` case. Please wrap the body of `flush()` in a
`try/catch`, or test both globals before the send.

### MINOR 5 — `stop()` before `start()` touches `sessionStorage`

`kit/tracker/src/index.ts:88`. A `stop()` call before the consent signal calls
`sessionStorage.removeItem` (my test "point 5f"). The call removes only the own key of the
tracker, thus the risk is small. A guard keeps the promise "no storage access before `start()`"
exact.

### MINOR 6 — the `tracker` job runs no lint step and no format step

`.github/workflows/ci.yml:106-131`. The job `frontend` runs `npm run lint` and
`npm run format:check`. The project `kit/tracker` has no ESLint config and no Prettier config,
thus the new job cannot run them. Please add the two tools and the two steps, for one standard
across the repository.

### MINOR 7 — the branch on a null session id is dead code

`kit/tracker/src/index.ts:129-132`. `getOrCreateSessionId()` returns `string`, never `null`. The
comment states a rule that the code cannot apply. Remove the branch, or make the function return
`string | null` for a failed UUID.

### MINOR 8 — no integration note for the app team

The package holds no `README.md`. An integrator needs one page: the consent gate, the two fields
that leave the browser, the `sessionStorage` key name, and the CSRF header hook. Design section 8
also names `docs/privacy.md`, and that file does not exist yet. That file stays owner work.

## Language

The module comments, the test names, and the pull request text follow ASD-STE100. I found no
defect there.

xamcross:
**Review (TypeScript library engineer): CHANGES REQUESTED**

I made my own clone and I checked out `issue-13-tracker-core` at `e52c3b9`. I ran `npm ci`,
`npm test`, and `npm run build` in `kit/tracker` with Node 24.21.0. I also broke the code in my
clone for each main criterion, I ran the suite again, and I restored each file. My clone is clean
and deleted. I changed nothing on the branch.

- `npm ci`: `added 90 packages`, `found 0 vulnerabilities`.
- `npm test`: `Test Files 1 passed (1)`, `Tests 20 passed (20)`.
- `npm run build`: no error. `dist/index.js` = 6256 bytes, gzip = 2088 bytes. The numbers match
  the bundle-size comment on issue #13 exactly.
- `gh pr checks 95`: the five jobs pass, `Tracker` included.

---

## 1. The acceptance criteria

|---|---|---|---|
| 1 | A click on a child of `data-octo="save"` records `save` | PASS | `src/index.test.ts:41`. The test clicks a child `<button>`, and the body holds `save`. |
| 2 | A click without the attribute records no entry | PASS | `src/index.test.ts:58`. |
| 3 | No click before `start()` | PASS | `src/index.test.ts:70`. |
| 4 | `stop()` empties the queue and removes the `sessionStorage` key | **FAIL** | See BLOCKER 1. Three separate mutations of `src/index.ts` each keep 20 of 20 tests green. |
| 5 | 120 clicks give requests of a maximum of 50 clicks | PASS | `src/index.test.ts:162`. A mutation to `MAX_BATCH_SIZE = 60` makes it fail. |
| 6 | The body holds only `sessionId` and `clicks` | PASS | `src/index.test.ts:199`. |
| 7 | `dependencies` is empty | PASS | `package.json:26`. |
| 8 | `npm test` passes | PASS | 20 of 20 in my clone. |
| 9 | The `tracker` job is green | PASS | `gh pr checks 95`. |
| 10 | The issue holds the bundle size | PASS | I measured the same two numbers. |

My mutation log for the criteria 1 to 5:

| Mutation of `src/index.ts` | Result |
|---|---|
| `MAX_QUEUE_SIZE = 300` | 1 test fails: "drops the oldest entry…" |
| `MAX_BATCH_SIZE = 60` | 1 test fails: "gives requests with a maximum of 50 clicks…" |
| Drop the `isDisabled(target)` check | 1 test fails: "skips a disabled element" |
| `if (false)` in place of the `started` guard | 1 test fails: "…a second start() adds no second listener" |
| Delete `queue = [];` in `stop()` (line 78) | **20 of 20 still pass** |
| Delete the `clearTimeout` block in `stop()` (lines 79-82) | **20 of 20 still pass** |
| Delete `removeStoredSessionId();` in `stop()` (line 88) | **20 of 20 still pass** |

## 2. The rules of D24

| Rule of D24 | Result | My evidence |
|---|---|---|
| No code at import time | PASS | The module holds only declarations. A plain Node import gives no error. |
| No action in a server-side render | PASS | `src/index.ts:67`. I deleted `globalThis.document` and I called `start()`. It threw no error. A later `start()` with a real DOM works, because the code sets `started` after the guard. |
| `endpoint`, `credentials`, `headers`, `flushIntervalMs` | PASS | `src/index.ts:13-22` and `:52-55`. The defaults are `same-origin` and 5000. |
| The capture-phase listener | PASS | `src/index.ts:73`. I put a `stopPropagation()` handler on a wrapper element, and the tracker still records the click. |
| `composedPath()` with an open shadow root | PASS | `src/index.test.ts:83`. |
| A disabled element | PARTIAL | See MINOR 2. |
| The queue of 200, the oldest entry dropped | PASS | `src/index.ts:104-108`. |
| Batches of a maximum of 50 | PASS | `src/index.ts:133-135`. A full queue of 200 gives 4 requests of 6463 bytes. That is below the 16 KB of rule C18 and below the 64 KB of D24. |

**The disabled element in detail.** D24 demands one rule: "It skips a disabled element". The code
tests the IDL property `disabled` on the element that holds `data-octo`. This covers a disabled
`<button>` or `<input>`. It does not cover a control inside a disabled `<fieldset>`: the IDL
property reflects only the content attribute, so `button.disabled` is `false` there, while
`button.matches(':disabled')` is `true`. I measured both values in jsdom, and the tracker records
that click. D24 says nothing about `aria-disabled`, and the code does not test it. A click on an
element with `aria-disabled="true"` gives an entry. I report the `fieldset` case as MINOR 2, and I
report the `aria-disabled` case as a note only, because D24 does not demand it.

## 3. `start()` and `stop()` in each order

I wrote my own probe tests for this. All four orders are correct:

- `start()` two times: one listener, one entry for one click.
- `stop()` two times: no error.
- `stop()` before `start()`: no error.
- `start()` after `stop()`: one listener, one timer, and one entry for one click.

The session id: `start()` after `stop()` makes a new id. That agrees with D24, because `stop()`
removes the `sessionStorage` key. Two flushes of one running tracker keep one id
(`src/index.test.ts:267`).

## 4. The timer and a failed `fetch`

- The one-shot `setTimeout` starts when a click enters an empty queue (`src/index.ts:110-112`). No
  interval exists. PASS.
- A rejected `fetch` gives no error to the page, and the next send still runs. I proved this with a
  probe: `mockRejectedValueOnce`, then a second click. The tracker made two requests, and the timer
  threw nothing. PASS, but no test in the suite covers it (MINOR 8).
- The scope of issue #36 stays out: no `pagehide`, no `visibilitychange`, no `keepalive`, no retry.
  PASS.
- The code makes one part of #36 harder. See MINOR 3.

## 5. `ageMs`

- An integer: PASS. `Math.round` in `src/index.ts:142` gives a JSON integer (rule C37).
- Never negative: PASS. `Math.max(0, …)` in the same line (rule C15).
- A clock that a system time change does not break: **FAIL**. See MAJOR 1.

## 6. The package

| Point | Result |
|---|---|
| `exports`, `types`, `files` | PASS. `files: ["dist"]` keeps the source out of the tarball. |
| `"type": "module"` | PASS. |
| `"sideEffects": false` | PASS. The module has no side effect at import time, so a bundler can drop it. |
| The build output | PASS. `tsc -p tsconfig.build.json` gives `dist/index.js` and `dist/index.d.ts`. The test file stays out. |
| The target | PASS. `ES2022` plus the `DOM` lib. This fits a modern browser and `crypto.randomUUID`. |
| No `dependencies` | PASS. |
| A small `devDependencies` list | PASS. Three entries, each with a clear reason: `typescript` builds, `vitest` runs the tests, `jsdom` gives the DOM environment that `vitest.config.ts` names. |
| A committed lock file | PASS. `package-lock.json`, 1830 lines. |
| `.nvmrc` | PASS. `24.21.0`, the same value as `monitor/frontend/.nvmrc`. |
| Issue #44 packs a `.tgz` | PASS. `npm pack --dry-run` gives `octometer-tracker-0.1.0.tgz`, 2.9 kB, 3 files. `"private": true` blocks `npm publish` but it permits `npm pack`. That agrees with D25 ("No npm account"). |

The layout permits issue #44. Two small points for that issue: the tarball holds no README
(MINOR 9), and the package has no `prepack` script, so #44 must run `npm run build` before
`npm pack`.

## 7. The CI job `tracker`

PASS on each point.

- `actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1` and
  `actions/setup-node@820762786026740c76f36085b0efc47a31fe5020`: the same two SHAs as the job
  `frontend`.
- No new `permissions` block. The workflow keeps `contents: read`.
- The diff of `.github/workflows/ci.yml` adds 26 lines and changes no other job.
- `persist-credentials: false`, the npm cache, and `node-version-file` follow the `frontend` job.

The diff holds no secret, no database file, and no build output. `.gitignore` already covers
`node_modules/` and `dist/`.

---

# The findings

## BLOCKER

**BLOCKER 1 — `kit/tracker/src/index.test.ts:131-146`. The test for acceptance criterion 4 proves
none of its three claims.**

The test creates the tracker, it records one click, and it calls `stop()` before the timer fires.
The tracker writes the session id only in `flush()` (`src/index.ts:126`), so this test never writes
the key. The assertion `expect(window.sessionStorage.getItem('octo_session_id')).toBeNull()` is true
before the test starts. The comment on line 139 says "so read it before stop()", but the test does
not read it. The other two claims in the test name are also free: `stop()` empties the queue **and**
it clears the timer, so the removal of either one alone keeps the final `expect(fetchMock).not
.toHaveBeenCalled()` true.

My evidence: I deleted each of the three lines of `stop()` in turn, and each time the whole suite
stayed green at 20 of 20. The behaviour of the code is correct; only the test is false. Acceptance
criterion 4 has no evidence.

**The exact fix.** Replace the one test with these three tests. I ran them in my clone: all three
pass on the branch as it is, and each one fails under its own mutation.

```ts
it('stop() empties the queue, so a restart sends only the new click', () => {
  const host = document.createElement('div');
  host.setAttribute('data-octo', 'save');
  document.body.appendChild(host);

  tracker = createTracker({ endpoint: ENDPOINT });
  tracker.start();
  clickElement(host);
  tracker.stop();
  tracker.start();
  clickElement(host);
  vi.advanceTimersByTime(5000);

  const calls = parseCalls(fetchMock);
  expect(calls).toHaveLength(1);
  expect(calls[0].body.clicks).toHaveLength(1);
});

it('stop() clears the pending flush timer', () => {
  const host = document.createElement('div');
  host.setAttribute('data-octo', 'save');
  document.body.appendChild(host);

  tracker = createTracker({ endpoint: ENDPOINT });
  tracker.start();
  clickElement(host);
  const withTimer = vi.getTimerCount();
  tracker.stop();

  expect(vi.getTimerCount()).toBe(withTimer - 1);
});

it('stop() removes the sessionStorage key that the flush wrote', () => {
  const host = document.createElement('div');
  host.setAttribute('data-octo', 'save');
  document.body.appendChild(host);

  tracker = createTracker({ endpoint: ENDPOINT });
  tracker.start();
  clickElement(host);
  vi.advanceTimersByTime(5000);
  expect(window.sessionStorage.getItem('octo_session_id')).not.toBeNull();

  tracker.stop();
  expect(window.sessionStorage.getItem('octo_session_id')).toBeNull();
});
```

One warning for the second test: jsdom schedules its own timer for the `storage` event on each
`sessionStorage` write. I measured this. Therefore the test compares the count before and after
`stop()`. A plain `expect(vi.getTimerCount()).toBe(0)` is not correct here.

## MAJOR

**MAJOR 1 — `kit/tracker/src/index.ts:109` and `:139`. `ageMs` uses the wall clock `Date.now()`. A
change of the system time corrupts it.**

The server calculates `ts = receivedAt - ageMs` (rules C14 and C15). A clock correction between the
click and the flush therefore moves the stored click time. I proved both directions with
`vi.setSystemTime`:

- A backward jump of 30 minutes gives `ageMs: 0`. The click looks new, and `ts` is wrong by
  30 minutes.
- A forward jump of one day gives `ageMs: 86400000`. The server clamps it to 600000, so `ts` is
  wrong by about 10 minutes.

An NTP correction, a time zone service, and a manual clock change all cause this. `performance.now()`
is monotonic, and it does not move with the system clock.

**The exact fix.** Add the helper, then use it in the two places.

```ts
function monotonicNow(): number {
  if (typeof performance !== 'undefined' && typeof performance.now === 'function') {
    return performance.now();
  }
  return Date.now();
}
```

- `src/index.ts:109`: `queue.push({ element, queuedAt: monotonicNow() });`
- `src/index.ts:139`: `const now = monotonicNow();`

Keep `Math.max(0, Math.round(now - entry.queuedAt))`, because `performance.now()` gives a fraction.
I applied this exact change in my clone: `tsc` gives no error, and the 20 tests stay green. Vitest
fake timers also fake `performance.now()`, so the assertion `ageMs: 5000` holds. Add one test with
`vi.setSystemTime` for a backward jump.

**MAJOR 2 — `kit/tracker/src/index.ts:145-147`. A `headers()` callback that throws throws into the
page, and it loses every batch of that flush.**

`Object.assign(headers, getExtraHeaders())` has no guard. The callback belongs to the app, and a
common body is a read of a CSRF token or of a session store. Such a read can throw. The error then
escapes from the `setTimeout` callback into the page, and the app sees an uncaught error from the
tracker. The loop of `flush()` also stops, so every later batch of the same flush is lost.

My probe: a `headers()` that throws, plus 120 clicks. The result is 0 requests, and the error
reaches the page.

**The exact fix.** Guard the callback, and guard the whole timer callback.

```ts
    const headers: Record<string, string> = { 'Content-Type': 'application/json' };
    if (getExtraHeaders) {
      try {
        Object.assign(headers, getExtraHeaders());
      } catch {
        // The tracker sends the batch with the default headers only.
      }
    }
```

Add the same guard around the body of the timer callback in `scheduleFlush`
(`src/index.ts:116-119`):

```ts
    timerId = setTimeout(() => {
      timerId = null;
      try {
        flush();
      } catch {
        // The tracker never gives an error to the page.
      }
    }, flushIntervalMs);
```

Add one test: a `headers()` that throws must not stop the request, and it must not throw.

**MAJOR 3 — `kit/tracker/package.json:18-22` and `kit/tracker/tsconfig.json`. No command type-checks
the test file, and the test file does not compile.**

`npm run build` uses `tsconfig.build.json`, which excludes `src/**/*.test.ts`. `vitest run` does not
type-check. Therefore `tsconfig.json` never runs in CI, and the strict options that it declares
(`strict`, `noUncheckedIndexedAccess`, `exactOptionalPropertyTypes`) do not protect the tests.

My evidence: `npx tsc -p tsconfig.json --noEmit` gives 24 errors, all in `src/index.test.ts`. For
example `src/index.test.ts:55`, `:97`, `:175`, `:176`, `:177`, `:210`, `:211`, `:212`, `:213`:
`error TS2532: Object is possibly 'undefined'`, because `noUncheckedIndexedAccess` makes `calls[0]`
possibly `undefined`. The first error is `src/index.test.ts(11,35): error TS2345`.

Issue #36 adds more tests to this file. It will hit the same wall.

**The exact fix.**

1. Add the script to `kit/tracker/package.json`:
   `"typecheck": "tsc -p tsconfig.json --noEmit"`.
2. Add the step to the `tracker` job in `.github/workflows/ci.yml`, before the test step:
   ```yaml
      - name: Type check the tracker code
        run: npm run typecheck
   ```
3. Make `src/index.test.ts` compile. The smallest change is a helper that gives a checked entry:
   ```ts
   function callAt(index: number) {
     const call = parseCalls(fetchMock)[index];
     if (call === undefined) {
       throw new Error(`no request at index ${index}`);
     }
     return call;
   }
   ```
   Then use `callAt(0).body` in place of `calls[0].body`. Fix the tuple type on line 11 with
   `fetchMock.mock.calls.map((call) => { const [url, init] = call as [string, RequestInit]; … })`.

## MINOR

A MINOR finding does not block the approval. I list 13.

1. `src/index.ts:129-132`. Dead code. `getOrCreateSessionId()` returns `string`, so `id === null` is
   never true. Delete the branch, or give the function the return type `string | null`.
2. `src/index.ts:206-208`. The disabled check misses a control inside a disabled `<fieldset>`. I
   measured `button.disabled === false` and `button.matches(':disabled') === true` for that case.
   Fix: `return element.matches(':disabled');` inside a try/catch, because an old engine can throw
   on that selector.
3. `src/index.ts:148`. `void fetch(...)` drops the promise. Issue #36 needs the promise for the
   lifecycle flush and for the retry rule. Change `sendBatch` to `return fetch(...)`, and collect
   the promises in `flush()`. The call site keeps `void`, so the behaviour of this pull request does
   not change.
4. `src/index.ts:215`. `crypto.getRandomValues(bytes)` throws a `ReferenceError` when `crypto` is
   absent. Wrap `generateUuid` in a try/catch, or test `typeof crypto === 'undefined'` first.
5. `src/index.ts:115-120`. `scheduleFlush()` overwrites `timerId` with no guard. The present code
   has no path that makes two timers, but one line makes it safe for ever:
   `if (timerId !== null) { return; }`.
6. `src/index.ts:88`. `stop()` before `start()` removes the `sessionStorage` key of a different
   tracker instance. I measured it. Call `removeStoredSessionId()` only when the instance made or
   read the id.
7. `src/index.test.ts:352`. The test name "does not call start() twice" says the opposite of the
   test, which calls `start()` two times. Rename it to "a second start() adds no second listener".
8. Test gaps. The suite has no test for: the capture phase itself (a wrapper with
   `stopPropagation()`), a rejected `fetch`, the SSR guard of `start()`, the default
   `credentials: 'same-origin'`, `stop()` two times, `stop()` before `start()`, and `start()` after
   `stop()`. I proved each behaviour with my own probe, and each one is correct today. A test keeps
   it correct.
9. `kit/tracker` has no `README.md`. The tarball of issue #44 then holds three files and no text for
   the reader. Add a short README with `start()`, `stop()`, and the four options.
10. `kit/tracker/package.json:23-25`. `"node": ">=20"` does not agree with `.nvmrc` 24.21.0. Use
    `">=24"`, or state the reason for the lower bound.
11. `kit/tracker/package.json:8-13`. `exports` has no `"./package.json": "./package.json"` entry.
    Some tools read that file through the package name.
12. Language (ASD-STE100). Four small points: `src/index.ts:106` uses "once it is full"; "when it is
    full" is the word of D24. `src/index.ts:154` says "A later issue"; name issue #36.
    `src/index.test.ts:139` describes a step that the test does not do; remove it with BLOCKER 1.
    The pull request text says "Click 201 pushes out click 1"; "pushes out" is a phrasal verb, so
    write "drops click 1". The rest of the comments, the commit bodies, and the pull request text
    follow the standard well.
13. `kit/tracker/tsconfig.build.json`. The build gives no `declarationMap` and no source map. An app
    developer then steps into the compiled file. Add `"declarationMap": true` and
    `"sourceMap": true` to the build config. A source map also needs the source in the tarball, so
    add `"src"` to the `files` list of `package.json`.

---

## Note on the design change in review (not a finding)

The change adds an optional field `path` to each click entry, and one entry `octo:session-start`
with `path` and `referrerHost` at each new session id. The present structure takes this change
without a rewrite. Three points help the author of that change:

1. `QueueEntry` (`src/index.ts:45-48`) is a private type. A field `path` costs two lines there, one
   line in `enqueue`, and one line in the `map` of `sendBatch`. The capture of `location.pathname`
   belongs in `handleClick`, not in `sendBatch`, because a single-page app changes the path before
   the flush.
2. `ELEMENT_PATTERN` (`src/index.ts:37`) already accepts `octo:session-start`, because rule C4
   permits the colon. No change is needed there.
3. One point needs care. `getOrCreateSessionId()` runs inside `flush()` (`src/index.ts:126`), after
   the code moves the queue into `entries`. A session-start entry that the code makes there cannot
   enter the batch, and its `path` would hold the value at the flush time, up to 5 seconds late. The
   clean change moves the call into `start()` or into the first `enqueue()`, and it pushes the
   session-start entry into the queue at that moment. That is about 15 lines. Note the side effect:
   the `sessionStorage` write then happens without a click, so the privacy reviewer must agree.

A fourth point for the size limit: 50 clicks with a 100-character element already give 6463 bytes. A
`path` field of 100 characters raises that to about 12 KB. That is still below the 16 KB of rule
C18, but the margin becomes small. A byte-based split may be needed in the same change.

---

## The summary

The library is well built. The structure is correct, the design agreement is high, and the package
layout fits both issue #36 and issue #44. One test is false and it hides acceptance criterion 4.
Two defects reach the app: a wall clock for `ageMs`, and a `headers()` callback that throws. One
gap lets a broken test file into `main`. Fix the one BLOCKER and the three MAJOR findings, and I
approve.

xamcross:
Correction round 1. I fixed each BLOCKER and each MAJOR finding. I fixed
each MINOR finding, with one rejection below. Commit `1abee2a` holds every
fix. `npm ci`, `npm run typecheck`, `npm test`, and `npm run build` pass in
`kit/tracker`, and `gh pr checks 95` is green.

## BLOCKER (both reviews)

**The `stop()` test proved nothing.** I removed the false test at the old
line 131 of `kit/tracker/src/index.test.ts`. I added three tests in its
place: "stop() empties the queue, so a restart sends only the new click",
"stop() clears the pending flush timer", and "stop() removes the
sessionStorage key that the flush wrote".

Proof, one mutation at a time, on `kit/tracker/src/index.ts`:
- I removed `queue = [];` from `stop()`. The first new test failed:
  "expected [] to have a length of 1 but got +0". I restored the line.
- I removed the `clearTimeout` block from `stop()`. The second new test
  failed: "expected 1 to be 0". I restored the block.
- I removed the `removeStoredSessionId();` call from `stop()`. The third
  new test failed: "expected '<uuid>' to be null". I restored the call.

All 37 tests pass with the code as it now stands.

## MAJOR

1. **`ageMs` used the wall clock.** Fixed. `kit/tracker/src/index.ts` now
   has a `monotonicNow()` helper on `performance.now()`, with a `Date.now()`
   fallback. `enqueue()` and `sendBatch()` use it. Two new tests set the
   system clock backward and forward with `vi.setSystemTime` and check that
   `ageMs` stays correct.
2. **A `headers()` throw stopped the flush.** Fixed. `sendBatch()` wraps the
   `headers()` call in a try/catch. `scheduleFlush()` wraps the whole
   `flush()` call in a try/catch, so no error reaches the page. New test:
   "sends the batch with the default headers when the headers() callback
   throws".
3. **No command checked the test file.** Fixed. `package.json` has a new
   `typecheck` script (`tsc -p tsconfig.json --noEmit`). The `tracker` job
   in `.github/workflows/ci.yml` runs it before `npm test`. I fixed the
   `noUncheckedIndexedAccess` errors in the test file with a small `at()`
   helper that throws a clear error for a missing array entry, in place of
   a plain index. `npx tsc -p tsconfig.json --noEmit` now gives no error.
4. **The `Referer` header could leak the page URL.** Fixed (privacy
   review). `sendBatch()` sets `referrerPolicy: 'no-referrer'` on the
   `fetch` call. New test: "sets referrerPolicy to no-referrer".

## MINOR, fixed

- Dead code: the `id === null` branch in `flush()` could never run. Removed.
- `isDisabled()` missed a control inside a disabled `<fieldset>`. Fixed with
  `element.matches(':disabled')`, in a try/catch, with the old IDL check as
  a fallback. New test: "skips a control inside a disabled fieldset".
- `sendBatch()` dropped its promise. It now returns `Promise<void>`; `flush()`
  calls it with `void`, so issue #36 can later collect the promises.
- `crypto.getRandomValues` could throw a bare `ReferenceError`. `generateUuid()`
  now throws a clear `Error` when `crypto` is absent, and the `flush()`
  try/catch of MAJOR 2 catches it.
- `scheduleFlush()` could not start a second timer in the code as it stood,
  but it now returns early when a timer is already pending, for safety.
- `stop()` before `start()` removed a `sessionStorage` key that this
  instance never wrote. Fixed: `stop()` calls `removeStoredSessionId()`
  only when `sessionId !== null`. New test: "does not touch sessionStorage
  when stop() runs before a flush ever wrote it".
- Renamed the test "does not call start() twice..." to "a second start()
  adds no second listener".
- Added the missing tests: the capture phase against a bubble-phase
  `stopPropagation()`, a rejected `fetch` that does not stop the next
  flush, the SSR guard on `start()`, and the default `credentials` value
  of `same-origin`. `stop()` twice and `stop()` before `start()` each have
  their own "does not throw" test.
- Added `kit/tracker/README.md`: `start()`, `stop()`, the four options, the
  two body fields, and the `sessionStorage` key name.
- `package.json` `engines.node` was `>=20`; `.nvmrc` holds `24.21.0`. Fixed
  to `>=24`.
- `package.json` `exports` had no `./package.json` entry. Added it.
- The app's `headers()` could replace `Content-Type`. Fixed: `sendBatch()`
  now sets `Content-Type` after `Object.assign`, so it always wins. New
  test: "does not let the caller headers() replace the Content-Type
  header".
- `endpoint` had no run-time check. `createTracker()` now throws a
  `TypeError` for an empty or a missing `endpoint`. New test: "throws a
  TypeError when the caller gives no endpoint".
- Language: "once it is full" is now "when it is full"
  (`kit/tracker/src/index.ts`, the `enqueue()` comment). "A later issue" now
  names issue #36 (the `sendBatch()` catch comment). The pull request text
  now reads "drops click 1" in place of "pushes out click 1".
- `tsconfig.build.json` now sets `declarationMap` and `sourceMap`. `src` is
  in the `files` list of `package.json`, so the source map resolves.

## MINOR, rejected

**The lint step and the format step of the `tracker` CI job (privacy
review).** I did not add them. Evidence: the issue text of #13 states
"`dependencies` in `package.json` stays empty" and "A `devDependencies`
entry needs a reason; keep the list small (TypeScript, a test runner with
jsdom)". An ESLint config, a Prettier config, and their packages are a new
scope that issue #13 does not ask for. I leave this point for a later
issue, together with the `README.md` badge and lint parity across the
kit modules.

## Bundle size

The fixes changed the built file. New size, from a clean `npm ci` and
`npm run build`: `dist/index.js` is 7860 bytes raw, 2586 bytes gzip. A
separate comment on issue #13 records this.

xamcross:
**Review (second review, TypeScript library engineer): CHANGES REQUESTED**

I made a new clone and I checked out `issue-13-tracker-core` at `1abee2a`. I ran each command with
Node 24.21.0. I judged the code, not the response comment.

- `npm ci`: `found 0 vulnerabilities`.
- `npm run typecheck`: **zero errors**, the test file included.
- `npm test`: `Tests 37 passed (37)`.
- `npm run build`: no error. `dist/index.js` = 7860 bytes, gzip = 2586 bytes.
- `gh pr checks 95`: the five jobs pass.

Every finding of round 1 is corrected. Two new defects came with the corrections, and both block the
approval. Each one has a small fix that I verified in my clone.

---

## The findings of round 1

| Finding | Status | My evidence |
|---|---|---|
| BLOCKER 1 — the false test for `stop()` | **CORRECTED** | See the mutation table below. Each of the three mutations now makes its own test fail. |
| MAJOR 1 — `ageMs` on the wall clock | **CORRECTED** | `src/index.ts:115`, `:149`, `:216-222`. See the detail below. |
| MAJOR 2 — a `headers()` that throws | **CORRECTED** | `src/index.ts:155-161`. A mutation that removes the try/catch makes the test on line 403 fail. See MINOR 3 below for the second guard. |
| MAJOR 3 — no type check of the test file | **CORRECTED** | `npm run typecheck` gives zero errors. The CI step is on line 127 of `ci.yml`. |
| MINOR 1 — dead `id === null` branch | CORRECTED | The branch is gone from `flush()`. |
| MINOR 2 — a disabled `<fieldset>` | CORRECTED | `src/index.ts:233-243` plus the test on line 155. |
| MINOR 3 — the dropped promise | CORRECTED | `sendBatch` returns `Promise<void>` (`src/index.ts:148`). |
| MINOR 4 — `crypto` absent | CORRECTED | `src/index.ts:249-251`. See MINOR 2 below for a second effect. |
| MINOR 5 — a second timer | CORRECTED | `src/index.ts:122-125`. |
| MINOR 6 — `stop()` before `start()` | CORRECTED, but too far | See **MAJOR 4**. |
| MINOR 7 — the wrong test name | CORRECTED | `src/index.test.ts:608`. |
| MINOR 8 — the test gaps | CORRECTED | New tests cover the SSR guard, the capture phase, a rejected `fetch`, the default `credentials`, `stop()` two times, `stop()` before `start()`, `start()` after `stop()`, and both clock jumps. |
| MINOR 9 — no README | CORRECTED | `kit/tracker/README.md`. But see MAJOR 4: one sentence of it is now wrong. |
| MINOR 10 — `engines` | CORRECTED | `">=24"`. |
| MINOR 11 — `exports` | CORRECTED | `"./package.json"` is present. |
| MINOR 12 — the language | CORRECTED | Line 112 says "when it is full". Line 173 names issue #36. The pull request text says "drops click 1". |
| MINOR 13 — no source map | CORRECTED | `tsconfig.build.json`. But see MINOR 4 below for the tarball. |

## 1. The three mutations of `stop()`

Each mutation makes its own test fail. The BLOCKER is closed.

| Mutation of `src/index.ts` | Result |
|---|---|
| Delete `queue = [];` (line 81) | 1 fails: "stop() empties the queue, so a restart sends only the new click" |
| Delete the `clearTimeout` block (lines 82-85) | 1 fails: "stop() clears the pending flush timer" |
| Delete `removeStoredSessionId();` (line 93) | 2 fail: "stop() removes the sessionStorage key that the flush wrote" and "creates a new sessionId after stop() and a later start()" |

## 2. `monotonicNow()`

| Point | Result | My evidence |
|---|---|---|
| The clock is monotonic | PASS | A mutation back to `Date.now()` makes the two clock tests fail. |
| A backward jump | PASS | A jump of 30 minutes gives `ageMs: 1000`, the true age. |
| A forward jump | PASS | A jump of one day gives `ageMs: 1000`. |
| An integer | PASS | With a real clock and a real fractional `performance.now()`, the body holds `{"element":"save","ageMs":1}`. `Math.round` (`src/index.ts:152`) does the work. |
| Never negative | PASS | `Math.max(0, …)` on the same line. |
| The fallback without `performance` | PASS | I removed the global `performance`. The tracker threw nothing, and it sent `ageMs: 1000`. The claim in the comment on line 216 holds. |

## 3. The try/catch around `headers()` and around the flush

| Point | Result | My evidence |
|---|---|---|
| A throw of `headers()` never reaches the page | PASS | The test on line 403, plus the mutation above. The request still goes, with `Content-Type` only. **The batch stays** in that case; the tracker does not drop it. |
| A throw inside `flush()` never reaches the page | PASS (untested) | I removed `crypto`, so `generateUuid()` throws inside `flush()`. Nothing reached the page. But a mutation that deletes the whole try/catch of lines 128-132 keeps 37 of 37 green. See MINOR 1. |
| The next flush still runs | PASS after a `headers()` throw and after a rejected `fetch` (the test on line 450). **FAIL after a throw inside `flush()`.** See MINOR 2. |
| No session id in the console | PASS | With a rejected `fetch` I counted the calls of `console.error`, `console.warn`, and `console.log`: `[0,0,0]`. The library writes nothing at all. |
| The pull request text names the fate of the batch | **NO** | The text says only "no longer reaches the page". Add one sentence: a throw inside a flush drops that batch, and issue #36 owns the retry rule. The code comment on line 131 already says it. |

## 4. `typecheck` and the CI step

All PASS.

- `npx tsc -p tsconfig.json --noEmit`: zero errors, `src/index.test.ts` included. The test file now
  uses a checked `at()` helper.
- `.github/workflows/ci.yml` grows by two lines only. The step "Type check the tracker code" runs
  after `npm ci`.
- The `tracker` job keeps `actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1` and
  `actions/setup-node@820762786026740c76f36085b0efc47a31fe5020`, the same two SHAs as `frontend`.
- No other job changed. The `permissions` block stays `contents: read`.

## 5. The rejected MINOR about ESLint and Prettier

**I accept the rejection, but for a different reason.** The stated reason is not exact: issue #13
says nothing about `devDependencies`. Its step 1 and its acceptance criterion name the runtime
`dependencies`, and that list is empty.

The correct reason is the scope. Step 6 of issue #13 names the three steps of the job: `npm ci`,
`npm test`, `npm run build`. A lint step and a format step belong to a later issue. Please open that
issue, because the `frontend` job runs `npm run lint` and `npm run format:check`, and the two Node
projects should match.

## 6. The growth of the bundle

**No, not every new byte is necessary.** The file grew from 6256 to 7860 bytes, and gzip from 2088
to 2586 bytes. The new logic explains most of it. One part does not: `tsc` keeps every comment in
the output, so the shipped file carries all the JSDoc and all the line comments.

I measured it: `tsc -p tsconfig.build.json --removeComments` gives **5953 bytes raw and 1802 bytes
gzip**. That is smaller than the file of round 1, with all the new logic inside. The reader of the
source loses nothing, because `files` now ships `src` and the build makes a source map.

Fix: add `"removeComments": true` to `kit/tracker/tsconfig.build.json`. Then measure the new size
for the comment on issue #13.

---

# The findings

## MAJOR (both block the approval)

**MAJOR 4 — `kit/tracker/src/index.ts:90-94`. `stop()` no longer always removes the
`sessionStorage` key. A session id of an earlier page load survives a consent withdrawal.**

The fix of my MINOR 6 went one step too far. `stop()` now calls `removeStoredSessionId()` only when
this instance made or read an id. The instance reads the id in `flush()` only. `sessionStorage`
lives for the whole browser tab, so the key survives a navigation.

The scenario: a user opens page 1 in a tab, gives consent, clicks, and a flush writes
`octo_session_id`. The user opens page 2 in the same tab. The app calls `start()`. The user
withdraws the consent inside the 5-second window, so the app calls `stop()` before the first flush.
The key stays.

My evidence: I put a key in `sessionStorage`, I started the tracker, I recorded one click, and I
called `stop()` before the timer. The key `id-of-the-earlier-page-load` is still there.

This breaks three statements:
- D24: "`stop()` empties the queue and removes the `sessionStorage` key."
- Acceptance criterion 4 of issue #13, the same sentence.
- `kit/tracker/README.md`: "`stop()` … removes the session id from `sessionStorage`."

The test on line 249 locks the wrong behaviour in place.

**The exact fix.** Remember the started state, and remove the key for each started instance. This
keeps MINOR 6 fixed, because an instance that never started still touches nothing.

```ts
  function stop(): void {
    const wasStarted = started;
    started = false;
    queue = [];
    if (timerId !== null) {
      clearTimeout(timerId);
      timerId = null;
    }
    if (clickListener !== null && typeof document !== 'undefined') {
      document.removeEventListener('click', clickListener, true);
    }
    clickListener = null;
    sessionId = null;
    if (wasStarted) {
      // A started tracker always removes the key, also with no flush before.
      removeStoredSessionId();
    }
  }
```

I applied this in my clone: `tsc` is clean, and 36 of 37 tests pass. Only the test on line 249
fails, as it must. Change that test to use a tracker that never started:

```ts
  it('does not touch sessionStorage when stop() runs before start()', () => {
    const removeItemSpy = vi.spyOn(window.sessionStorage.__proto__, 'removeItem');
    tracker = createTracker({ endpoint: ENDPOINT });
    tracker.stop();

    expect(removeItemSpy).not.toHaveBeenCalled();
  });
```

Then add the test for the case of this finding:

```ts
  it('stop() removes a session id of an earlier page load, also before a flush', () => {
    window.sessionStorage.setItem('octo_session_id', 'an-older-id');
    tracker = createTracker({ endpoint: ENDPOINT });
    tracker.start();
    tracker.stop();

    expect(window.sessionStorage.getItem('octo_session_id')).toBeNull();
  });
```

**MAJOR 5 — `kit/tracker/src/index.ts:162-163`. The guard on the `Content-Type` header works only
for one spelling. The comment claims more than the code does.**

The comment says "The tracker sets Content-Type last, so an app header cannot replace it". A
JavaScript object key is case-sensitive, and an HTTP header name is not. An app that returns
`{ 'content-type': 'text/plain' }` puts a second key in the object. Both keys reach `fetch`.

My evidence: the header object becomes
`{"content-type":"text/plain","Content-Type":"application/json"}`. A real `Headers` object makes one
field of it: `content-type: text/plain, application/json`. The server then reads a content type that
is not `application/json`, and contract rule C19 makes it answer 415. Every request of that app
fails, and the tracker writes nothing to the console. Contract rule C12 demands
`Content-Type: application/json`.

The reach is small, because few apps set a content type in that callback. The cost is total: the app
loses every click. A comment that promises a guarantee must hold for each spelling.

**The exact fix.** Remove each spelling of the name before the tracker sets its own.

```ts
    // The tracker owns the content type, so it drops any case of that name.
    for (const name of Object.keys(headers)) {
      if (name.toLowerCase() === 'content-type') {
        delete headers[name];
      }
    }
    headers['Content-Type'] = 'application/json';
```

I applied this in my clone: `tsc` is clean and 37 of 37 tests pass. Add one test with
`headers: () => ({ 'content-type': 'text/plain' })`, and assert that the header object holds exactly
one key with that name.

## MINOR

Four MINOR findings. They do not block the approval.

1. `src/index.ts:128-132`. The try/catch around `flush()` has no test. I deleted the whole block and
   37 of 37 tests stayed green. Add a test that removes the global `crypto` and asserts that
   `vi.advanceTimersByTime()` throws nothing.
2. `src/index.ts:140-142`. After a throw inside `getOrCreateSessionId()`, the tracker stays silent
   for ever. The line `queue = []` comes after the call, so the queue keeps its entries. A later
   click then finds `queue.length` above 1, and `enqueue()` schedules no new timer. I measured 0
   pending timers after the failed flush, and still 0 after a later click, with a working `crypto`.
   The fix is one move: put `const entries = queue; queue = [];` **above** the
   `getOrCreateSessionId()` call. A throw then drops one batch, and the next click starts a new
   timer.
3. `src/index.ts:52-54`. `createTracker()` now throws a `TypeError` for an empty endpoint. D24 does
   not ask for this, and no review finding asked for it. The guard is reasonable, but it changes the
   contract of the function. Name it in the README under Options, so an app author knows it.
4. `package.json:14-17`. `files: ["dist", "src"]` ships the test file. `npm pack --dry-run` shows
   `src/index.test.ts` at 21.3 kB, the largest file of the 8. Issue #44 puts this tarball on a
   GitHub Release. Use `"files": ["dist", "src/index.ts"]`.

---

## The summary

The corrections are good work. The three mutations of `stop()` each fail now, the clock is
monotonic and it gives a correct integer `ageMs` in every case that I tested, a throw of the app
never reaches the page, and CI type-checks the test file. The test count of 37 is honest: I broke
the code in six places and each break made a test fail, with one exception that I list as MINOR 1.

Two defects came in with the corrections. Both have a fix of a few lines that I ran in my clone.
Fix MAJOR 4 and MAJOR 5, add `"removeComments": true`, and I approve.

xamcross:
**Review (privacy and security engineer, second review): CHANGES REQUESTED**

I made a new clone, I checked out `issue-13-tracker-core` at `1abee2a`, and I ran each command
with the portable Node 24.21.0. I checked the code and the built file, not the response comment.
I wrote 16 new tests of my own in my clone (`src/privacy.review2.test.ts`, not committed). All 16
pass. The two MAJOR findings of my first review are correct now. One fix of a MINOR finding
creates a new MAJOR defect. See MAJOR 3.

## My commands

| Command | Result |
| --- | --- |
| `npm ci` | `added 90 packages`, `found 0 vulnerabilities` |
| `npm run typecheck` | No error |
| `npm test` | `Test Files 1 passed (1)`, `Tests 37 passed (37)` |
| `npm run build` | `dist/index.js` is 7860 bytes, 2586 bytes after gzip. The two numbers agree with the new comment on issue #13 |
| `npm audit` | `found 0 vulnerabilities` |

## The findings of the first review

**MAJOR 1, the `Referer` header: CORRECTED.** `src/index.ts:168` sets
`referrerPolicy: 'no-referrer'`. The built file holds the same line
(`dist/index.js:132`). The test at `src/index.test.ts:437` checks it. My test "r1" checks the
`fetch` init of a cross-origin endpoint: it holds only `body`, `credentials`, `headers`,
`method`, and `referrerPolicy`, and `credentials` is `same-origin`.

**MAJOR 2, the false `stop()` test: CORRECTED.** The old test is gone. Three tests take its
place (`src/index.test.ts:188`, `:206`, `:222`). I made the mutation that you asked for: I
commented out `removeStoredSessionId();` in `stop()` (`src/index.ts:93`) in my clone. Two tests
then failed:
- `stop() removes the sessionStorage key that the flush wrote`
- `creates a new sessionId after stop() and a later start()`

I restored the file, and all 37 tests pass again. My test "r3" also proves the queue rule: a
click that is in the queue at the time of `stop()` gives no request after 600000 ms, and the
`sessionStorage` key is absent.

**MINOR 1, the default `credentials`: CORRECTED** (`src/index.test.ts:424`, my test "r1").
**MINOR 2, the `Content-Type` override: CORRECTED** (`src/index.ts:162-163`, my test "r11": the
app header `X-XSRF-TOKEN` arrives, and `Content-Type` stays `application/json`).
**MINOR 3, the run-time check of `endpoint`: CORRECTED** (`src/index.ts:52-54`, my test "r12"
for an empty value, a missing key, and `undefined`).
**MINOR 4, the guards on `crypto` and on `fetch`: CORRECTED.** My test "r8" removes `crypto`, and
my test "r10" removes `fetch`. Neither one gives an error to the page, and neither one sends a
request. See MINOR 10 for a side effect.
**MINOR 5, `stop()` before `start()`: CORRECTED** (my test "r13": no `getItem`, no `setItem`, no
`removeItem`, no listener, no timer). This fix creates MAJOR 3 below.
**MINOR 6, the lint step and the format step: NOT CORRECTED, and I accept the rejection.** The
issue text asks for a small `devDependencies` list. A later issue can add the two tools across
the kit modules.
**MINOR 7, the dead branch in `flush()`: CORRECTED** (`src/index.ts:136-146`).
**MINOR 8, the integration note: CORRECTED** (`kit/tracker/README.md`). The file names the
consent gate, the two body fields, the four options, and the storage key.

## The six points of the second review

1. **The referrer.** PASS. See MAJOR 1 above.
2. **The `stop()` key.** PASS for the test, FAIL for one path. See MAJOR 2 and MAJOR 3.
3. **The MINOR findings.** PASS, except the rejection that I accept.
4. **The new try/catch blocks.** PASS. My test "r7" makes the `headers()` callback throw and the
   `fetch` call reject in one run. It spies on `log`, `info`, `warn`, `error`, `debug`, and
   `trace`. None of the six runs, thus no session id and no element name reaches the console.
   The catch blocks in `src/index.ts:130`, `:158`, and `:172` hold a comment only.
5. **The supply chain.** PASS. `dependencies` is `{}`. The lock file has no change in this
   commit. It holds 115 packages; each one has an `integrity` value and a `resolved` value on
   `registry.npmjs.org`. The `devDependencies` list still holds `jsdom`, `typescript`, and
   `vitest` only. `npm audit` gives no finding. The new CI step runs `npm run typecheck`. It adds
   no `uses:`, no permission, and no secret. See MINOR 9 for the package contents.
6. **The body.** PASS. My test "r6" clicks a button with a text and a `value` attribute. The body
   holds only `sessionId` and `clicks`, and the entry holds only `element` and `ageMs`. The raw
   body holds no text of the element, no form value, no URL, and no user agent. My test "r14"
   confirms that a bad `data-octo` value still never reaches a request. My test "r15" confirms
   that `ageMs` comes from `performance.now()`: a jump of the system clock does not change it.

## New findings

### MAJOR 3 — `stop()` keeps the session id of a previous page view, thus a re-consent reuses it

`kit/tracker/src/index.ts:90-94`. The fix of MINOR 5 removes the key only when `sessionId` is not
`null` in this instance. `sessionStorage` survives a reload and a navigation in the same tab.
The flush creates `sessionId`, thus a page view without a flush keeps `sessionId` at `null`.

My test "r4": the previous page view writes the key, this page view calls `start()`, the user
clicks one time, and the app calls `stop()` before the first flush. The key stays in
`sessionStorage`. My test "r5": a later `start()` reads that key, and the next request carries
the old session id. The consent withdrawal of the user does not break the link between the two
periods.

This breaks design decision D24 ("`stop()` empties the queue and removes the `sessionStorage`
key"), the acceptance criterion of issue #13 with the same words, and the promise of
`kit/tracker/README.md` ("removes the session id from `sessionStorage`").

Fix: keep a flag for the life of the instance, and remove the key when the app started the
tracker one time.

```ts
  let everStarted = false;
```

In `start()`, after `started = true;`, add `everStarted = true;`. In `stop()`, replace the block
at line 90:

```ts
    sessionId = null;
    if (everStarted) {
      everStarted = false;
      removeStoredSessionId();
    }
```

MINOR 5 stays correct: a `stop()` before the first `start()` still touches no storage. Please add
one test with a key that this instance never wrote, and one test that a `start()` after that
`stop()` gives a new session id.

### MINOR 9 — the npm package now holds the test file

`kit/tracker/package.json:16-19` adds `src` to `files`, for the source map. `npm pack --dry-run`
gives 8 files, and `src/index.test.ts` (21.3 kB) is one of them. The test code then ships to each
app that consumes the package. Fix: name the one file that the map needs.

```json
  "files": [
    "dist",
    "src/index.ts",
    "README.md"
  ],
```

### MINOR 10 — a failed flush stalls the queue for the life of the page

`kit/tracker/src/index.ts:136-142`. `flush()` calls `getOrCreateSessionId()` before the line
`queue = [];`. An error from `generateUuid()` leaves the queue full. `enqueue()` starts a timer
only when the queue length is 1, thus no later click starts a new timer. My test "r9" proves it:
after one failed flush, the tracker sends nothing again, also when `crypto` comes back. The
clicks stay in the memory of the page. Fix: take the entries and empty the queue before the call
to `getOrCreateSessionId()`.

### MINOR 11 — the comment and the README state the removal without a condition

`kit/tracker/src/index.ts:28-32` and the section "The consent gate" of `kit/tracker/README.md`
state that `stop()` removes the session id. MAJOR 3 shows the exception. The fix of MAJOR 3 makes
the two texts true again.

## Other checks

The diff of `1abee2a` holds no secret and no text with the form of a credential. The language of
the new README, of the new comments, and of the commit body follows ASD-STE100. I found one point
only: the README sentence "The tracker keeps it under the `sessionStorage` key `octo_session_id`,
for the life of one browser tab" is correct, and it becomes fully correct with the fix of
MAJOR 3.

xamcross:
Correction round 2. Commit `7ce9293` holds each fix. `npm ci`, `npm run
typecheck`, `npm test` (40 of 40), `npm run build`, and `npm pack
--dry-run` all pass. `gh pr checks 95` is green, `Tracker` included.

## MAJOR (both reviews)

**`stop()` kept an earlier session id.** Fixed with a new `everStarted`
flag, `kit/tracker/src/index.ts:65`, `:77`, `:94-99`. `stop()` now removes
the key each time `start()` ran before, also with no flush. A `stop()`
before the first `start()` still touches no storage. Proof: I put the old
guard back, ran the new test "stop() removes a session id of an earlier
page load, also before a flush", and it failed ("expected 'an-older-id' to
be null"). I restored the fix. The old wrong test is replaced by two tests
at `src/index.test.ts:249` and `:257`.

**The `Content-Type` guard covered one spelling.** Fixed at
`kit/tracker/src/index.ts:169-176`: the tracker deletes each key whose
name equals `content-type` in any case, before it sets its own. Proof: I
removed the loop and ran the new test "drops each spelling of the
content-type header name…"; it failed with a real `Headers` object merge
("text/plain, application/json"). I restored the fix.

## MINOR with a real effect

**A failed flush stalled the queue.** Fixed at `src/index.ts:141-149`: the
queue empties before the `getOrCreateSessionId()` call, so a throw there
drops one batch only. Proof: I moved the session id call back above the
queue reset, ran the new test "does not throw and does not stall the queue
…", and it failed with 0 requests after the crypto came back. I also
removed the `try`/`catch` around `flush()` and the same test failed with
an uncaught `Error`. I restored both fixes.

## MINOR, fixed

- `package.json:16-19`: `files` now names `dist`, `src/index.ts`, and
  `README.md`. `npm pack --dry-run` lists 7 files, no test file.
- `tsconfig.build.json` sets `removeComments: true` for the JS build.
  `removeComments` also strips a `.d.ts` in this TypeScript version, so I
  added a second config, `tsconfig.build.types.json`, for a
  declaration-only pass with no `removeComments`. `npm run build` now runs
  both. `dist/index.d.ts` keeps every doc comment; `dist/index.js` keeps
  none, only the required source-map pragma line.
- `README.md` and the `stop()` doc comment
  (`kit/tracker/src/index.ts:28-33`) now state the exact rule: `stop()`
  removes the key only when `start()` ran at least one time before.
- Added the `endpoint` `TypeError` behaviour to the README options table.

## Bundle size

The build changed shape. New size, from a clean `npm ci` and
`npm run build`: `dist/index.js` is 6202 bytes raw, 1866 bytes gzip. A new
comment on issue #13 records this.

xamcross:
**Review (privacy and security engineer, third review): APPROVE**

I made a new clone at `7ce9293`, and I ran each command with the portable Node 24.21.0. I wrote 9
new tests of my own in my clone (`src/privacy.review3.test.ts`, not committed). All 9 pass. No
BLOCKER and no MAJOR finding remains. I have no new MINOR finding.

| Command | Result |
| --- | --- |
| `npm ci` | `found 0 vulnerabilities` |
| `npm run typecheck` | No error |
| `npm test` | `Tests 40 passed (40)` |
| `npm run build` | 6202 bytes, 1866 bytes after gzip. The two numbers agree with the new comment on issue #13 |
| `npm audit` | `found 0 vulnerabilities` |

1. **MAJOR 3: CORRECTED.** `src/index.ts:65`, `:77`, `:94-99` hold the `everStarted` flag. My
   test "t1": a key of an earlier page view is absent after `stop()`, also with no flush before.
   My test "t2": a later `start()` makes a new UUID, not the old id. My test "t3": `stop()`
   before the first `start()` calls no `getItem`, no `setItem`, and no `removeItem`, and it keeps
   a foreign key. I also made a mutation: I commented out `everStarted = true;` in `start()`.
   Three tests failed, the two new ones included. I restored the file, and 40 tests pass again.
2. **MINOR 9: CORRECTED.** `npm pack --dry-run` lists 7 files: `README.md`, `package.json`,
   the four `dist` files, and `src/index.ts`. No test file ships.
3. **MINOR 10: CORRECTED.** `flush()` empties the queue before the session id call
   (`src/index.ts:142-149`). My test "t4": after a flush with no Web Crypto API, the next click
   starts a timer and the tracker sends again. My test "t5": the dropped batch never appears in a
   later request, after five more flushes.
4. **MINOR 11: CORRECTED.** The `stop()` doc comment (`src/index.ts:28-33`) and
   `README.md:10-12` state the exact rule: `stop()` removes the key only when `start()` ran at
   least one time before.
5. **The build and the supply chain: PASS.** The build runs two `tsc` passes.
   `dist/index.js` holds no comment except the source-map pragma on the last line, and no test
   code, no `console` call, no `Math.random`, no `localStorage`, no `document.cookie`, and no
   internal note. `dist/index.d.ts` keeps the doc comments, as a consumer needs. The source map
   holds no `sourcesContent`. `dependencies` is `{}`. The lock file has no change in this commit;
   it holds 115 packages, each with an `integrity` value, and 0 production packages. The
   workflow file has no change. The commit holds no secret.

The privacy checks of my first review stay true at this commit: `referrerPolicy: 'no-referrer'`,
the default `credentials` of `same-origin`, a body with `sessionId` and `clicks` only, an entry
with `element` and `ageMs` only, the C4 check before the queue, and a silent console on a failed
flush (my tests "t6" to "t9"). The header guard now drops each spelling of the content type
name, thus a real `Headers` merge gives `application/json` only.

I approve this pull request for the privacy role and the security role. I cannot use
`gh pr review --approve`, because the pull request comes from the same GitHub account.

xamcross:
**Review (third review, TypeScript library engineer): APPROVE**

New clone, `issue-13-tracker-core` at `7ce9293`, Node 24.21.0. `npm ci`, `npm run typecheck` (zero
errors), `npm test` (**40 of 40**), and `npm run build` all pass. `gh pr checks 95` is green.
`dist/index.js` = 6202 bytes raw, 1866 bytes gzip, the two numbers of the response comment.

## The four points

**1. MAJOR 4 — CORRECTED.** I put `octo_session_id` in `sessionStorage`, started the tracker,
recorded one click, and called `stop()` before the timer. The key is `null`. A `stop()` before the
first `start()` makes **0** calls of `getItem`, `setItem`, and `removeItem`; I counted them with
three spies. A second cycle of `start()` and `stop()` also removes a stale key, so the
`everStarted` flag resets correctly (`src/index.ts:65`, `:77`, `:94-99`). I put the old guard back
in my clone: the test "stop() removes a session id of an earlier page load, also before a flush"
fails. I restored the file.

**2. MAJOR 5 — CORRECTED.** I gave a `headers()` that returns two spellings at one time,
`content-type: text/plain` and `CONTENT-TYPE: text/html`, plus `X-App`. The header object becomes
`{"X-App":"demo","Content-Type":"application/json"}`, and a real `Headers` object gives
`[["content-type","application/json"],["x-app","demo"]]`. Exactly one content type, and the app
header survives (contract rule C12). A revert of the loop at `src/index.ts:169-176` makes its test
fail.

**3. The stall — CORRECTED.** `flush()` empties the queue before the session id call
(`src/index.ts:141-149`). I removed the global `crypto`, so `getOrCreateSessionId()` throws. Nothing
reached the page, and the tracker dropped that one batch. The next click gave **1** pending timer
and **1** request. Two reverts each make the test "does not throw and does not stall the queue when
getOrCreateSessionId() throws" fail: the move of the session id call above the queue reset, and the
removal of the `try`/`catch` around `flush()`. My MINOR 1 and MINOR 2 of round 2 are both closed.

**4. The build and the package — CORRECT.** The two `tsc` passes do what they must.
`dist/index.js` holds one comment line only, the source-map pragma on line 204. `dist/index.d.ts`
holds 23 doc comment lines. `npm pack --dry-run` lists 7 files, 7.1 kB: `README.md`,
`dist/index.js`, `dist/index.js.map`, `dist/index.d.ts`, `dist/index.d.ts.map`, `src/index.ts`, and
`package.json`. No test file. I unpacked the real tarball and checked each target of `exports` and
`types`: all four exist. The packed `dist/index.js` imports, and `start()` plus `stop()` run.

## The other checks

- The round-2 diff touches six files under `kit/tracker` only. It changes no CI job, so the pinned
  action SHAs and the `permissions` block stay as I checked them before. No secret, no build output.
- The README and the `stop()` doc comment now state the exact rule. Both agree with the code.
- The README options table names the `TypeError` of `createTracker()`. My MINOR 3 of round 2 is
  closed. `files` no longer ships the test file; my MINOR 4 is closed.
- `ageMs` still holds: after a backward clock jump of one hour the body is
  `{"sessionId":"<uuid>","clicks":[{"element":"save","ageMs":1000}]}`.

## One open point, not a finding

Please open the follow-up issue for ESLint and Prettier in the `tracker` job. The `frontend` job
runs `npm run lint` and `npm run format:check`, and the two Node projects should match. That work
sits outside issue #13, whose step 6 names only `npm ci`, `npm test`, and `npm run build`.

Good work. I have no BLOCKER and no MAJOR finding left. From the view of the TypeScript library,
the browser behaviour, the tests, and the package, this pull request is ready to merge.

Issue #58, opened 2026-09-21
Test the store against a corrupt, a locked, and a full database


xamcross:
**Blocked by #28.** Each test runs against the SQLite store of #9, and the lock test needs the status `ERROR` of #28. The free disk space field of `/api/health` can start now.

**Goal.** The monitor keeps its data and gives a clear error when `octometer.db` is corrupt, locked, or full.

**Implementation steps.**
1. Run `PRAGMA quick_check` at the start. A failed check gives exit code 2.
2. Make a second monitor instance refuse to start. Use the lock file `octometer.lock` in `dataDir` with `FileChannel.tryLock()` (D3).
3. Set the status `ERROR` when a write waits longer than `busy_timeout`.
4. Roll back the page transaction on `SQLITE_FULL`.
5. Add `databaseBytes` and `freeDiskBytes` to `GET /api/health`.
6. Write the tests of the acceptance criteria.

**Acceptance criteria.**
- [ ] Random bytes in `octometer.db` give exit code 2, one clear log line, and an unchanged file.
- [ ] `PRAGMA quick_check` runs at the start.
- [ ] A second process with the write lock gives `busy_timeout`, then the status `ERROR`.
- [ ] A second monitor instance refuses to start.
- [ ] `PRAGMA max_page_count` simulates `SQLITE_FULL`, the page transaction rolls back, and the cursor stays.
- [ ] `/api/health` returns the database size as `databaseBytes` and the free disk space as `freeDiskBytes`.
- [ ] `./gradlew :monitor:backend:test` passes.
- [ ] A comment on this issue records the test result.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, D3, D4, D8, and D13.

xamcross:
Body edit on 2026-09-21: #9 closed. The first line now names only the open blockers: #28.

xamcross:
Note on 2026-09-22 from the maintainer, from the second security review of #125 (https://github.com/xamcross/octometer/pull/125#issuecomment-5768041389): the application start of #15 removes each orphan secret from `secrets/apps.json` before this issue adds the lock `octometer.lock`. Thus two monitor processes that start at the same time can both write the secret file. When this issue adds the lock, take the lock BEFORE `MonitorServices.open` runs the sweep.

Issue #61, opened 2026-09-21
Delete the events of one user id in the monitor


xamcross:
**Goal.** One request deletes each event of one user id of one app from the monitor store. It also deletes each row with `user_id IS NULL` of a session of that user (rules E1 and E2).

**Implementation steps.**
1. In `monitor/backend`, read the `session_id` list of one `app_id` and one `user_id`. The user id is a bound parameter.
2. Add a repository function that deletes the rows with that `user_id`, and the rows with `user_id IS NULL` and a `session_id` of that list.
3. Run `PRAGMA wal_checkpoint(TRUNCATE)` on the writer connection after the delete commits.
4. Add the route `DELETE /api/apps/{appId}/events?userId=<id>` that calls the function. An unknown app id gets 404.
5. Return 200 with the JSON field `deleted`, which holds the row count.
6. Add the tests with `testApplication`.

**Acceptance criteria.**
- [x] The rows of that user id are gone, and the rows of each other user id stay.
- [x] A row with `user_id IS NULL` of a session of that user is gone.
- [x] A row of a second user in the same session stays.
- [x] A row with `user_id IS NULL` of another session stays.
- [x] The field `deleted` counts the two row groups.
- [x] `PRAGMA wal_checkpoint(TRUNCATE)` runs after the delete.
- [x] The user id is a bound parameter and appears in no line of the captured log.
- [x] The captured log holds no query string.
- [x] The response is 200 with the field `deleted`, and an unknown app id gives 404.
- [x] `./gradlew :monitor:backend:test` passes.
- [x] A comment on this issue records the test output.

**Related.**
- #5 protects the route.
- #31 owns the format of the call log.
- #64 gives the `curl` command with the `Origin` header, because version 1 has no UI control.
- #35 uses the same rule in the kit stores.
- The design change of 2026-09-21 added rules E1 and E2 to this issue.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, D13, D15.

xamcross:
Body edit on 2026-09-21: #9 closed. No open blocker remains, thus the first line and the label `blocked` are removed. The work can start.

xamcross:
Body edit of 2026-09-21: rules E1 and E2 of the design change. The erasure reads the session list of the user first, then it deletes the rows of that user and the anonymous rows of those sessions. A row of a second user in the same session stays. The issue keeps 6 steps.

The maintainer made this edit. The source is the design brief in the first comment of #101.

[...]

Issue #56, opened 2026-09-21
Add the upgrade script


xamcross:
**Blocked by #54 and #39.** The script switches the launcher and the task of #54, restores the backup of #55, and downloads the release of #39. The `schemaVersion` field of step 1 can start now.

**Goal.** `scripts/upgrade.ps1 -Version X.Y.Z` installs a release of the monitor and restores the old version on a failure.

**Implementation steps.**
1. Add `schemaVersion` to `GET /api/health`. Read the value from `PRAGMA user_version`.
2. Create `scripts/upgrade.ps1` with the parameter `-Version X.Y.Z`. It gets the zip with `gh release download X.Y.Z --repo xamcross/octometer --pattern "octometer-monitor-*.zip"`.
3. Make the script expand the zip into `app\X.Y.Z\`.
4. Add the switch to the script. It stops the task `Octometer`, changes the launcher `octometer.cmd` to the new version, and starts the task.
5. Make the script read `/api/health` until it shows the new version, for a maximum of 60 seconds. A timeout starts the failure path.
6. Add the failure path. The script restores the launcher and the pre-migrate backup.

**Acceptance criteria.**
- [ ] The script stops the task, switches the launcher, starts the task, and waits until `/api/health` shows the new version.
- [ ] When `/api/health` does not show the new version in 60 seconds, the script restores the launcher and the pre-migrate backup.
- [ ] `/api/health` returns `schemaVersion`, and `./gradlew :monitor:backend:test` passes.
- [ ] A comment on this issue records one real upgrade run of the installed monitor.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, D36 (second part) and D13.

xamcross:
Body edit on 2026-09-22: #55 closed. The first line now names only the open blockers: #54 and #39.

Issue #51, opened 2026-09-21
Serve the level 3 totals


xamcross:
**Goal.** `GET /api/apps/{appId}/elements` returns one row for each element that one user touched: the clicks, the sessions, and the last interaction timestamp.

**Implementation steps.**
1. Add the route `GET /api/apps/{appId}/elements` to `monitor/backend`. The response has no pages.
2. Accept exactly one of `userId=<id>` and `anonymous=true`. Each other request gets 400.
3. Write the level 3 statement of section 6. It filters with `user_id IS ?`, it groups by `element`, and it holds the literal `kind = 0`.
4. Return `lastInteractionAt` as UTC ISO 8601 with milliseconds.
5. Return 404 for an unknown app id.
6. Write the tests of the acceptance criteria.

**Acceptance criteria.**
- [x] A request with exactly one of `userId` and `anonymous=true` passes. Each other request gets 400. #113 adds the third filter `sessionId`.
- [x] The SQL uses `user_id IS ?`.
- [x] The anonymous rows are correct.
- [x] The response holds `rows` with `element`, `clicks`, `sessions`, and `lastInteractionAt`.
- [x] `lastInteractionAt` is UTC ISO 8601 with milliseconds.
- [x] The order is `clicks DESC, element ASC`.
- [x] An unknown app id gives 404.
- [x] The captured log holds no query string.
- [x] `./gradlew :monitor:backend:test` passes.
- [x] A comment on this issue records the test result.

**Related.** #31 owns the format of the call log. #109 gives the column `kind` and the partial index `event_agg`. #113 adds the third filter `sessionId=<uuid>`.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, D13, D15, and section 6.

xamcross:
Body edit on 2026-09-21: #9 closed. No open blocker remains, thus the first line and the label `blocked` are removed. The work can start.

xamcross:
Body edit of 2026-09-21: rules M4 and M7 of the design change. The elements route gets a third filter `sessionId` from #113, and the level 3 statement gets the literal `kind = 0` from #111. The criterion and the Related line name the two issues. The steps stay as they are.

The maintainer made this edit. The source is the design brief in the first comment of #101.

xamcross:
[...]

PR #96, opened 2026-09-21
feat(monitor): add the SQLite store with versioned migrations

### LINKED ISSUE #9: Add the SQLite store with versioned migrations
**Goal.** The monitor has a SQLite store with versioned migrations. One transaction commits a page plus its cursor.

**Implementation steps.**
1. Confirm the current `sqlite-jdbc` version with Maven Central before you use it.
2. Add the migration file `db/001_init.sql` to `monitor/backend` with the DDL of section 6.
3. Add a migration runner of about 30 lines with `PRAGMA user_version`.
4. Set these pragmas on each connection, before the migrations: `journal_mode=WAL`, `synchronous=NORMAL`, `busy_timeout=5000`, `foreign_keys=ON`, `secure_delete=ON`, `temp_store=MEMORY`, `cache_size=-65536`.
5. Open one writer connection on a single-thread dispatcher and one separate read connection. Use plain JDBC on the file `octometer.db` in `dataDir`.
6. Add one repository function that commits a page and the cursor in one `BEGIN IMMEDIATE` transaction (D4).

**Acceptance criteria.**
- [x] A test makes the insert fail and asserts that the cursor did not move.
- [x] A replay of the same page adds 0 rows.
- [x] A row with `element = NULL` gives an error. The store does not ignore it.
- [x] `PRAGMA foreign_keys` returns 1.
- [x] `PRAGMA secure_delete` returns 1.
- [x] A second start applies no migration.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, O3, D3, D4, section 6.

xamcross:
## Summary

This pull request adds the SQLite store of D3 and D4. It adds the migration
`db/001_init.sql` with the exact DDL of section 6. It opens one private
writer connection on a single-thread dispatcher, and one private,
read-only reader connection, each with the seven pragmas of step 4.
`EventStore.commitPage` commits one page of events and the cursor in one
`BEGIN IMMEDIATE` transaction.

## Confirm: the sqlite-jdbc version

- Source: `https://repo1.maven.org/maven2/org/xerial/sqlite-jdbc/maven-metadata.xml`.
- Date: 2026-09-21.
- Result: the release version is `3.53.4.0`. `gradle/libs.versions.toml` now
  holds this version.

## Acceptance criteria

- [x] A test makes the insert fail and asserts that the cursor did not
  move. Evidence: `EventStoreTest.a failed insert leaves the cursor as it
  was and rolls back the good event too`. The bad page holds a good event
  first, and a bad event with an empty user id second. The `CHECK`
  constraint of the `event` table breaks the insert. The whole `BEGIN
  IMMEDIATE` transaction rolls back. Neither event is present, and the
  cursor stays at its old value.
- [x] A replay of the same page adds 0 rows. Evidence:
  `EventStoreTest.a replay of the same page adds 0 rows and still moves
  the cursor`. The second `commitPage` call with the same events adds 0
  rows, because of `ON CONFLICT(app_id, event_id) DO NOTHING`, and the
  cursor still moves to the new value.
- [x] A row with `element = NULL` gives an error. Evidence:
  `EventStoreTest.a row with element NULL gives an error at the schema
  level`. A raw insert with a NULL element gives an `SQLException` with a
  `NOT NULL` message, straight from the schema.
- [x] `PRAGMA foreign_keys` returns 1. Evidence: `PragmaTest.the writer
  connection has each of the seven pragmas` and the matching reader test.
- [x] `PRAGMA secure_delete` returns 1. Evidence: the same two `PragmaTest`
  tests, one for the writer and one for the reader connection. Each test
  now asserts all seven pragmas, not only these two.
- [x] A second start applies no migration. Evidence:
  `MigrationRunnerTest.a second start applies no migration`. The second
  `SqliteDatabase.open` call on the same file does not throw, and a
  direct call to `MigrationRunner.run` after it returns an empty list.
- [x] `./gradlew :monitor:backend:test` passes. Evidence: a local run on
  2026-09-21 with `JAVA_HOME` set to the JDK 21 install: `BUILD
  SUCCESSFUL`, 27 tests, 0 failures, in the `store`, `config`, and root
  packages of `monitor/backend`.
- [x] A comment on the issue records the test output. Evidence: the
  comment on issue #9 after this pull request.

## Correction round 1

Two reviewers found 1 BLOCKER, 6 MAJOR, and 13 MINOR issues. Each fix and
each evidence sits in the response comment of this pull request. A short
list:

- The bad page of the rollback test now holds a good event first. The
  test failed with no transaction in the code, and passed once the
  transaction was back.
- The writer and the reader connections are now private.
  `SqliteDatabase.write` and `SqliteDatabase.read` are the only entry
  points, and each runs the block on the correct thread.
- The reader connection opens read-only.
- The migration runner stops with a clear error on a database that is
  newer than the code.
- `SqliteDatabase.open` closes each connection it already made when a
  later step fails.
- `BEGIN IMMEDIATE` sits inside a `try`. A `finally` rolls back when the
  commit did not happen. A failed `ROLLBACK` attaches to the first error.
- `PRAGMA journal_mode` reads its own answer back, and the store checks
  each of the seven pragmas.

## Notes

- The DDL of section 6 has no defect. This pull request uses it exactly,
  with `db/001_init.sql`.
- `.gitignore` already covers `*.db*`. It matches `*.db`, `*.db-wal`, and
  `*.db-shm`. No test writes to the real data folder. Each test opens its
  own temporary folder.
- This pull request does not add the poll cycle, a route, or a MongoDB
  reader. A later issue owns each of them.
- This pull request adds `kotlinx-coroutines-core` to the version catalog,
  for the single-thread dispatcher of the writer connection.
- The lock file `octometer.lock` and `PRAGMA quick_check` of D3 are still
  absent. Issue #58 is the closest owner in the backlog.

Closes #9

Generated with Claude Code


xamcross:
**Review (SQLite and data engineer): CHANGES REQUESTED** — first review.

I cloned the repository, checked out `issue-9-sqlite-store` at `404b644`, and ran each
check myself. I wrote ten probe tests in my clone. I deleted them after the run.

## Acceptance criteria

|---|---|---|
| A failed insert leaves the cursor | PASS | `EventStoreTest:56-68` passes. My probe repeats it. |
| A replay adds 0 rows | PASS | My probe: `rows=2 cursor=cursor-2` after two equal pages. The cursor also moves. |
| `element = NULL` gives an error | PASS | `EventStoreTest:72-89` passes. The `STRICT` table gives the `NOT NULL` error. |
| `PRAGMA foreign_keys` returns 1 | PASS | My probe reads `1` on the writer and on the reader. |
| `PRAGMA secure_delete` returns 1 | PASS | My probe reads `1` on the writer and on the reader. |
| A second start applies no migration | PASS | My probe: a third `MigrationRunner.run` call returns `[]`. |
| `./gradlew :monitor:backend:test` passes | PASS | `BUILD SUCCESSFUL in 13s`, 23 tests, 0 failures, 0 errors, `--rerun-tasks`. |
| A comment records the test output | PASS | The comment on issue #9 of 2026-09-21. |

## My own checks

1. **The DDL against section 6.** PASS. I put lines 405 to 427 of the design beside
   `001_init.sql` lines 5 to 26 and ran `diff`. The two texts are the same, character for
   character: each table, each column, each `CHECK`, each `STRICT`, both indexes, and both
   primary keys. The only difference is the markdown fence of the design.
2. **The migration runner.** PASS on the main path. I added a second migration file in my
   clone. The first start set `user_version=1`. The second start applied only file 002:
   `user_version=2`, the new table exists, and file 001 did not run again. I then made file
   002 fail in the middle. The runner threw, and the state stayed at `user_version=1` with no
   half result: the first statement of the failed file left no table. One transaction covers
   one file, and it covers `PRAGMA user_version` too. The newer-version guard is absent, see
   MAJOR-2.
3. **The seven pragmas on each connection.** PASS. I read each value back:
   `journal_mode=wal`, `synchronous=1`, `busy_timeout=5000`, `foreign_keys=1`,
   `secure_delete=1`, `temp_store=2`, `cache_size=-65536`. The values are the same on the
   writer and on the reader. `SqliteDatabase.kt:55-59` applies them before
   `MigrationRunner.run` at line 50.
4. **The page transaction.** PASS on the commit count. I read `PRAGMA data_version` on the
   reader before and after each page. A page of 3 rows moved it by 1. A page of 1000 rows
   also moved it by 1. One page is one commit. The SQL holds
   `ON CONFLICT(app_id, event_id) DO NOTHING`. `SQLITE_BUSY`: a rival connection held
   `BEGIN IMMEDIATE`; `commitPage` threw `org.sqlite.SQLiteException: [SQLITE_BUSY]` after
   5558 ms, thus the `busy_timeout` of 5000 ms works. The code does not retry. That is correct
   for the single-writer design of D3, and issue #58 owns the locked-database test.
5. **The reader connection.** FAIL, see MAJOR-1. A read during a writer transaction does not
   block: my probe read the table in 0 ms and saw the old snapshot while the writer held an
   open `BEGIN IMMEDIATE`. WAL works.
6. **Note on `secure_delete` plus WAL, for issue #61.** I measured this on the branch:
   - `secure_delete=ON` zeroes the row inside the page image that SQLite writes. In WAL mode
     that zeroed page goes into `octometer.db-wal`, never straight into `octometer.db`.
   - After the `DELETE`, and before a checkpoint, the old text is still in `octometer.db`.
     My probe found the marker there.
   - A `PRAGMA wal_checkpoint(PASSIVE)` copies the zeroed page over `octometer.db`. The
     marker leaves the main file. The `-wal` file keeps its size (16512 bytes in my run). The
     bytes of the old frames stay in the `-wal` file until a new frame writes over them.
   - A row that lived only in the WAL keeps its plain text in the `-wal` file until the
     checkpoint. My probe found the marker in `octometer.db-wal`.
   - `PRAGMA wal_checkpoint(TRUNCATE)` sets the `-wal` file to 0 bytes.
   - **Conclusion for #61: an erasure needs `DELETE` plus `PRAGMA wal_checkpoint(TRUNCATE)`.
     `secure_delete=ON` alone leaves the text in `octometer.db` or in `octometer.db-wal`.**
7. **The `sqlite-jdbc` version.** PASS. I read
   `https://repo1.maven.org/maven2/org/xerial/sqlite-jdbc/maven-metadata.xml` myself on
   2026-09-21: `<release>3.53.4.0</release>`, `lastUpdated 20260826100422`. The POM of
   3.53.4.0 answers 200. `gradle/libs.versions.toml:9` holds `3.53.4.0`. The native library
   loads on Windows: the CI run 35635205581, job "Monitor backend on Windows", is green, and
   each store test opens a real SQLite file.
8. **The diff.** PASS. 10 files, 480 lines, 0 deletions. No `.db`, no `-wal`, no `-shm`, no
   build output. `.gitignore:2` holds `*.db*` and `.gitignore:6` holds `build/`. Each test
   opens its own `Files.createTempDirectory` folder.
9. **Note on the later migration 002.** The runner of this pull request can apply it. I ran
   the exact shape in my clone against the schema of 001: `ALTER TABLE event ADD COLUMN path
   TEXT`, `ALTER TABLE event ADD COLUMN referrer_host TEXT`, `ALTER TABLE event ADD COLUMN
   kind INTEGER NOT NULL DEFAULT 0`, `DROP INDEX event_session`, and a partial index with a
   `WHERE` clause. Each statement ran, and the `event` table then had 9 columns. Three points
   for the author of 002:
   - A `STRICT` table accepts `ADD COLUMN`. The new column needs a type of the `STRICT` list.
   - `ADD COLUMN ... NOT NULL` needs a `DEFAULT`, which `kind INTEGER NOT NULL DEFAULT 0`
     has.
   - `executeUpdate` runs each statement of the file, thus a file with five statements works.
     The whole file stays in one transaction.

## Findings

### MAJOR-1. The reader connection can write.
`monitor/backend/src/main/kotlin/octometer/monitor/store/SqliteDatabase.kt:49`, with
`openConnection` at lines 55 to 59.

The reader uses the same plain `DriverManager.getConnection(url)` as the writer. My probe ran
`INSERT INTO gap (app_id, from_ts, to_ts) VALUES (1, 1, 2)` on `database.reader`, and the row
went in. Step 5 asks for a read connection, and D3 puts each write on the single-thread
dispatcher. A later issue that holds the reader can write off that thread and break the
single-writer rule.

Fix. Open the reader read-only. I tested this exact code in my clone, and it works:

```kotlin
private fun openConnection(url: String, readOnly: Boolean = false): Connection {
    val config = org.sqlite.SQLiteConfig()
    config.setReadOnly(readOnly)
    val connection = DriverManager.getConnection(url, config.toProperties())
    SqlitePragmas.applyTo(connection)
    return connection
}
```

Then call `openConnection(url)` for the writer and `openConnection(url, readOnly = true)` for
the reader. Keep the order: the writer first, because a read-only connection cannot create the
file or the `-shm` file. My measurements on the read-only connection: each of the seven
pragmas reads back the correct value (`journal_mode=wal`, `synchronous=1`,
`busy_timeout=5000`, `foreign_keys=1`, `secure_delete=1`, `temp_store=2`,
`cache_size=-65536`), a `SELECT` works, a read during a writer transaction takes 0 ms, and an
`INSERT` gives `[SQLITE_READONLY] Attempt to write a readonly database`. Add a test that
asserts the `SQLException` on a write.

### MAJOR-2. The runner starts on a database that is newer than the code.
`monitor/backend/src/main/kotlin/octometer/monitor/store/MigrationRunner.kt:21-31`.

`run` skips each migration with a version at or below `user_version`, and it reports nothing
when `user_version` is above the last known version. My probe set `PRAGMA user_version = 7`
and started the code again. No error came. The message was: "NO ERROR. The code started on
user_version=7 with 5 tables." The owner has `install.ps1` and `upgrade.ps1`, thus a
downgrade is a real path. Old code then reads and writes a newer schema. That risks wrong
totals or lost data.

Fix. Add the guard at the top of `run`:

```kotlin
fun run(connection: Connection): List<Int> {
    val current = userVersion(connection)
    val latest = MIGRATIONS.maxOf { it.version }
    check(current <= latest) {
        "The database is at user_version $current, and this build knows $latest. " +
            "Install a newer monitor build, or restore a backup."
    }
    ...
}
```

Add a test that sets `PRAGMA user_version` above the last version and asserts the message.

### MINOR-1. A pragma read skips `next()`.
`MigrationRunner.kt:48-51`. `statement.executeQuery("PRAGMA user_version").use { it.getInt(1) }`
reads column 1 without a `next()` call. `PRAGMA user_version` always gives one row, thus the
code works today. My probe proves the trap: the same shape on `SELECT id FROM app WHERE id =
-1` returned `0` and threw nothing. A future query with no row would give a silent `0`, and
the runner would apply migration 1 again. Fix: call `next()` first, and fail when it returns
`false`. The same shape is in `MigrationRunnerTest.kt:47-50` and `PragmaTest.kt:38-41`.

### MINOR-2. The insert of a page does not use a JDBC batch.
`EventStore.kt:46-58`. The loop calls `insert.executeUpdate()` once for each row. The commit
count is correct, one commit for a page of 1000 rows, and the transaction is correct. The
round trips are not needed. Fix: call `insert.addBatch()` in the loop and
`insert.executeBatch()` after it. Keep the `ON CONFLICT DO NOTHING` clause, because
`executeBatch` then reports 0 for a duplicate row.

### MINOR-3. The replay test does not assert the cursor.
`EventStoreTest.kt:45-52`. The test asserts 2 rows, and it does not read the cursor. D4 asks
for the cursor move on a replay too. My probe shows the code is correct: the cursor moved from
`cursor-1` to `cursor-2`. Fix: add `assertEquals("cursor-2", readCursor(database, appId))`.

### MINOR-4. `PragmaTest` covers two of the seven pragmas.
`PragmaTest.kt:33-36`. The two acceptance criteria pass, and the other five pragmas of step 4
have no test. I read each of the seven myself, and each is correct. Fix: assert
`journal_mode=wal`, `synchronous=1`, `busy_timeout=5000`, `temp_store=2`, and
`cache_size=-65536` in the same helper.

### MINOR-5. A failed migration leaks the two connections.
`SqliteDatabase.kt:48-52`. `MigrationRunner.run(writer)` at line 50 can throw. The writer, the
reader, and the executor then have no owner. My probe called `open` three times against a
failing migration, and `octometer.db-wal` plus `octometer.db-shm` stayed in the folder,
because the connections were still open. On Windows the open handle blocks a repair or a
restore of the file. D3 asks for "leaves the file as it is" after a failed check. Fix: put
lines 48 to 52 in a `try`, and close the writer, the reader, and `writerDispatcher` in the
`catch` before you throw again.

### MINOR-6. The result of `PRAGMA journal_mode=WAL` is not checked.
`SqlitePragmas.kt:13`. `journal_mode` is the one pragma that stays in the file, and it is the
one pragma that can fail without an exception. On a network share SQLite keeps the old mode
and answers `delete`. The store then loses the reader that does not block. Fix: read the
answer of that one statement, and throw when it is not `wal`.

### MINOR-7. `updateCursor` drops the update count.
`EventStore.kt:60-66`. `update.executeUpdate()` gives 0 when the `app` row is absent, and
`commitPage` still reports success. With an empty page the caller then believes the cursor
moved. Fix: `check(update.executeUpdate() == 1) { "No app row $appId" }`.

## Scope and language

The scope is correct. The pull request adds the store, and no route, no poll cycle, and no
MongoDB reader. The `octometer.lock` lock file and the `PRAGMA quick_check` of D3 are absent,
and that is correct here: issue #58 "Test the store against a corrupt, a locked, and a full
database" owns them. The language of each comment, of each document, and of both commit bodies
follows ASD-STE100. No secret is in the diff.

Summary: 2 MAJOR, 7 MINOR, 0 BLOCKER. Fix MAJOR-1 and MAJOR-2, and I approve.

xamcross:
**Review (Kotlin backend engineer): CHANGES REQUESTED**

I reviewed the Kotlin code, the threads, the lifecycle, the error handling, and the build.
A second reviewer owns the SQLite and the data model. I worked in my own clone at the head of
`issue-9-sqlite-store` (`404b644`), with `JAVA_HOME` on the JDK 21 install. I wrote my own
probe tests, I read their output, and I deleted them again.

---

## Acceptance criteria

| Criterion | Result | Evidence |
|---|---|---|
| A test makes the insert fail and asserts that the cursor did not move | **FAIL** | See BLOCKER 1. The test passes with no transaction at all. |
| A replay of the same page adds 0 rows | PASS | I deleted `ON CONFLICT(app_id, event_id) DO NOTHING` from `EventStore.kt:18`. `EventStoreTest > a replay of the same page adds 0 rows() FAILED`. |
| A row with `element = NULL` gives an error | PASS | `EventStoreTest.kt:73-88`. The schema rejects it. |
| `PRAGMA foreign_keys` returns 1 | PASS | I deleted `"PRAGMA foreign_keys=ON"` from `SqlitePragmas.kt`. Both `PragmaTest` tests FAILED. My own probe read `foreign_keys writer=1 reader=1`. |
| `PRAGMA secure_delete` returns 1 | PASS | My probe read `secure_delete writer=1 reader=1`. |
| A second start applies no migration | PASS | I changed `if (migration.version > current)` to `if (true)` at `MigrationRunner.kt:25`. `MigrationRunnerTest > a second start applies no migration() FAILED`. |
| `./gradlew :monitor:backend:test` passes | PASS | `BUILD SUCCESSFUL in 14s` with `--rerun-tasks`. 23 tests, 0 failures. |
| A comment on the issue records the test output | PASS | The comment on #9 is present. |

My probe also confirmed the seven pragmas on each connection: `journal_mode=wal`,
`synchronous=1`, `busy_timeout=5000`, `foreign_keys=1`, `secure_delete=1`, `temp_store=2`,
`cache_size=-65536`. The schema after `open()` holds `app`, `event`, `skipped_event`, `gap`,
`event_agg`, and `event_session`. The multi-statement `executeUpdate` of the migration works.

## Scope

The scope is correct. The pull request adds no poll cycle, no route, and no MongoDB code.

**The store is not a part of the application start.** `grep` finds `SqliteDatabase` and
`EventStore` only inside the package `octometer.monitor.store`. `Application.kt` is not in the
diff. **Thus pull request #5 has no conflict in the application module.** The issue that wires
the store must add the `open()` call, the `close()` call on stop, the lock file `octometer.lock`
of D3, and `PRAGMA quick_check` of D3.

## The version of kotlinx-coroutines (step 5 of my task)

I confirmed `1.11.0` myself against
`https://repo1.maven.org/maven2/org/jetbrains/kotlinx/kotlinx-coroutines-core/maven-metadata.xml`
on 2026-09-21: `<release>1.11.0</release>`, `<latest>1.11.0</latest>`.

`./gradlew :monitor:backend:dependencies --configuration runtimeClasspath` shows one version
only on the class path. Ktor 3.6.0 brings the same version:

```
+--- io.ktor:ktor-server-core:3.6.0
|    \--- io.ktor:ktor-server-core-jvm:3.6.0
|         +--- org.jetbrains.kotlinx:kotlinx-coroutines-core:1.11.0
```

Ktor forces no version upward. Each new version sits in `gradle/libs.versions.toml`, and the
build file holds `implementation(libs.kotlinx.coroutines.core)` and
`implementation(libs.sqlite.jdbc)` only. **PASS.**

## The database path (step 6 of my task)

`SqliteDatabase.kt:42` builds the path `<dataDir>/octometer.db`. The code creates the folder
`dataDir` and its parents, and nothing else. Each store test opens its own folder from
`Files.createTempDirectory`. No test reads `dataDir` from the config. **No test touches the real
data folder of the owner. PASS.**

---

# Findings

## BLOCKER 1. The first acceptance criterion has no test that can fail

`monitor/backend/src/test/kotlin/octometer/monitor/store/EventStoreTest.kt:58-68`

**Evidence.** I deleted the whole transaction from `EventStore.commitPage`: the
`BEGIN IMMEDIATE` of line 34, the `COMMIT` of line 38, and the `rollbackQuietly` call of line
40. Then I ran the test class:

```
> ./gradlew :monitor:backend:test --tests "octometer.monitor.store.EventStoreTest"
BUILD SUCCESSFUL in 7s
```

All four tests stay green with no transaction at all. The reason: `badPage` on line 61 holds
one event. `insertEvents` throws on that one event, thus `updateCursor` never runs. The cursor
cannot move, with a transaction or without one. The test proves the order of two calls. It does
not prove the rollback of D4.

The comment on lines 55-56 and the pull request text both claim "the whole page rolls back".
The test shows no such thing.

**The exact fix.** Give the bad page a good event first and a bad event second. Then the
assertion on the event count proves the rollback:

```kotlin
val badPage = listOf(sampleEvent("e2"), sampleEvent("e3", userId = ""))
assertFailsWith<SQLException> {
    store.commitPage(appId, badPage, cursor = "cursor-2")
}

assertEquals("cursor-1", readCursor(database, appId))
assertEquals(1, countEvents(database, appId))
```

**I verified this fix.** With the fix and the present code: `BUILD SUCCESSFUL`. With the fix and
the transaction removed:

```
EventStoreTest > a failed insert leaves the cursor as it was() FAILED
```

Please add a second test as well: a good page, then a failure inside the cursor update, to prove
that an inserted event rolls back too.

## MAJOR 2. Each use of the writer connection runs off the single-thread dispatcher

`monitor/backend/src/main/kotlin/octometer/monitor/store/SqliteDatabase.kt:19` and `:50`

`val writer: Connection` is public. Nothing holds a caller to the dispatcher. Step 5 of the
issue and D3 ask for the opposite rule.

**Evidence 1.** `MigrationRunner.run(writer)` on line 50 runs on the caller thread. My probe
printed `openCalledOn=Test worker`, while the dispatcher thread is `octometer-sqlite-writer`.
The first use of the writer connection is already off the dispatcher.

**Evidence 2.** The tests of this pull request use the writer connection from the JUnit thread:
`EventStoreTest.kt:75`, `:101`, `:110`, `MigrationRunnerTest.kt:39`, `:49`, `PragmaTest.kt:25`.

**Evidence 3.** A foreign write is not only untidy, it breaks the page transaction. My probe
opened `BEGIN IMMEDIATE`, then wrote one `app` row from a second thread through
`database.writer`, then ran `ROLLBACK`:

```
PROBE4 result: foreignRowsAfterRollback=0 (0 means the foreign write was lost)
```

The foreign write joined the page transaction and died with it. Issue #15 adds the app registry,
which writes to the same `app` table. A registry insert that lands inside a failed page
transaction disappears with no error.

**The exact fix.** Make the connection private. Give the class the only entry point:

```kotlin
class SqliteDatabase private constructor(
    private val writer: Connection,
    private val reader: Connection,
    private val writerDispatcher: ExecutorCoroutineDispatcher,
) : AutoCloseable {

    /** Each write of the store runs here, on the one writer thread (D3, step 5). */
    suspend fun <T> write(block: (Connection) -> T): T =
        withContext(writerDispatcher) { block(writer) }
```

`EventStore.commitPage` then calls `database.write { writer -> ... }`. Run the migration on the
same thread inside `open()`:

```kotlin
runBlocking(writerDispatcher) { MigrationRunner.run(writer) }
```

Each test moves to `write { }` too.

## MAJOR 3. A failed `open()` leaks both connections and blocks the delete of the folder

`monitor/backend/src/main/kotlin/octometer/monitor/store/SqliteDatabase.kt:39-53`

Line 48 opens the writer. Line 49 opens the reader. Line 50 runs the migration. A throw on line
49 or line 50 leaves each open connection open forever.

**Evidence.** My probe created `octometer.db` with a conflicting `app` table, then called
`SqliteDatabase.open`:

```
PROBE3 result: thrown=org.sqlite.SQLiteException: [SQLITE_ERROR] SQL error or missing database
               (table app already exists) folderDeleted=false
PROBE3 leftOver=octometer.db, octometer.db-shm, octometer.db-wal
```

On Windows the open handle blocks the delete of the folder. A failed start therefore leaves a
locked file, and the backup of D3 and the upgrade script of #56 cannot move it.

The happy path is clean. My probe opened and closed the store 50 times:
`writerThreadsBefore=0 writerThreadsAfter=0 foldersNotDeleted=0`.

**The exact fix.** Close each part that `open()` already made:

```kotlin
fun open(dataDir: String): SqliteDatabase {
    val folder = File(dataDir)
    check(folder.mkdirs() || folder.isDirectory) {
        "The data folder '$dataDir' is not available."
    }
    val url = "jdbc:sqlite:" + File(folder, DATABASE_FILE_NAME).absolutePath

    val writer = openConnection(url)
    try {
        val reader = openConnection(url)
        try {
            val writerDispatcher = Executors.newSingleThreadExecutor { runnable ->
                Thread(runnable, "octometer-sqlite-writer")
            }.asCoroutineDispatcher()
            runBlocking(writerDispatcher) { MigrationRunner.run(writer) }
            return SqliteDatabase(writer, reader, writerDispatcher)
        } catch (error: Throwable) {
            runCatching { reader.close() }
            throw error
        }
    } catch (error: Throwable) {
        runCatching { writer.close() }
        throw error
    }
}
```

## MAJOR 4. `BEGIN IMMEDIATE` sits outside the `try`, and a failed `ROLLBACK` stays silent

`monitor/backend/src/main/kotlin/octometer/monitor/store/EventStore.kt:34-42` and `:68-70`

Line 34 opens the transaction outside the `try`. A throw there runs no recovery. Line 69 hides
each failure of the `ROLLBACK`, because `runCatching` drops the result. A connection that stays
inside a transaction therefore stays that way, and each later page fails.

**Evidence.** My probe left one transaction open on the writer connection, then called
`commitPage` two times:

```
PROBE5 first:  [SQLITE_ERROR] SQL error or missing database (cannot start a transaction within a transaction)
PROBE5 second: [SQLITE_ERROR] SQL error or missing database (cannot start a transaction within a transaction)
```

The store never recovers. `catch (error: Exception)` on line 39 also lets each `Error` pass with
no rollback.

**The exact fix.** Use a `finally`, and report a failed rollback:

```kotlin
writer.createStatement().use { it.execute("BEGIN IMMEDIATE") }
var committed = false
try {
    insertEvents(writer, appId, events)
    updateCursor(writer, appId, cursor)
    writer.createStatement().use { it.execute("COMMIT") }
    committed = true
} finally {
    if (!committed) {
        writer.createStatement().use { it.execute("ROLLBACK") }
    }
}
```

A `finally` covers an `Error` too. Attach a failed `ROLLBACK` to the first error with
`addSuppressed`, so the caller sees both.

## MAJOR 5. The store gives the later API issues no safe read path

`monitor/backend/src/main/kotlin/octometer/monitor/store/SqliteDatabase.kt:20`

`val reader: Connection` is public, and the class offers no read function. Issues #18, #50, and
#51 each add a Ktor route that queries this one connection.

**What the code does today.** I ran 3200 reads from 8 coroutines on `Dispatchers.IO` against the
one reader connection:

```
PROBE1 result: exceptions=0 wrongResults=0 of 3200 reads
PROBE8 result (expect 40000 each): [w4=40000, w2=40000, w3=40000, w1=40000]
```

sqlite-jdbc locks the connection, thus a parallel call gives a correct result. **The store does
not break.** It serialises:

```
PROBE9 oneQueryMs=16978 fourParallelMs=51527
```

Four equal queries in parallel take about three times the time of one.

**What #18, #50, and #51 need.**
1. A `suspend` read function, because a route must not block a Ktor worker thread with a JDBC
   call. Today each route must remember `withContext(Dispatchers.IO)` on its own.
2. More than one read connection. #18 measured 0.33 s for the level 1 query at 1 million events.
   With one connection, ten parallel requests make the last one wait about 3.3 s. WAL allows
   parallel readers, so a small pool removes the wait.

**The exact fix.** Add a read API to `SqliteDatabase`, and open a pool of 2 to 4 read
connections:

```kotlin
suspend fun <T> read(block: (Connection) -> T): T =
    withContext(Dispatchers.IO) { readPool.borrow { connection -> block(connection) } }
```

A first, smaller step keeps one connection and adds the `suspend` wrapper plus a `Mutex`. Please
make `reader` private in each case, so a route cannot take the raw connection.

---

# MINOR findings

1. **`ResultSet.getInt` without `next()`.** `MigrationRunner.kt:50`, `MigrationRunnerTest.kt:50`,
   `PragmaTest.kt:40`. JDBC puts the cursor before the first row, thus `getInt` must throw.
   sqlite-jdbc accepts it today (my probe read `withoutNext=7 withNext=7`). An empty result
   returns 0 with no error (`emptyResultWithoutNext=0`). A driver change therefore makes the
   runner read 0 and run `001_init.sql` again, which fails with "table app already exists". Fix:
   `statement.executeQuery("PRAGMA user_version").use { check(it.next()); it.getInt(1) }`.
2. **`SqliteDatabase.kt:41` drops the result of `mkdirs()`.** A folder that the process cannot
   create gives an unclear driver message later. The fix sits in MAJOR 3.
3. **A cancel does not roll back.** `EventStore.kt:32`. `withContext` cannot interrupt a
   blocking JDBC call. My probe cancelled the caller during a page of 60000 events:
   `PROBE6 cursorAfterCancel=cursor-cancelled eventsAfterCancel=60000 nextCommit=OK`. The page
   commits, the caller still gets a `CancellationException`, and the connection stays outside a
   transaction. The store does not break, and `commitPage` does not swallow the
   `CancellationException`, because line 41 throws it again. Please add one line of KDoc that
   states this behaviour, so the poller of #17 handles it.
4. **The error type is the raw driver type.** `EventStore.kt`. The caller receives
   `org.sqlite.SQLiteException`. D3 asks the poller to treat `SQLITE_FULL` in a special way, and
   #28 and #58 need that. A small `StoreException` with the SQLite result code helps them. The
   store writes no log line, thus no user id and no event body reach a log. **PASS on that point.**
5. **Two parts of D3 have no owner.** The lock file `octometer.lock` with `FileChannel.tryLock()`
   and `PRAGMA quick_check` at the start. The steps of #9 do not name them, and no backlog issue
   owns them. Please name an owner.
6. **The language of three comments.** `MigrationRunner.kt:14-16`: "It compares each migration
   version with `PRAGMA user_version` and runs only the versions above it, thus a second start of
   the same file applies no migration again." That is 27 words. The limit is 25.
   `SqliteDatabase.kt:36-37` joins three verbs with "and". The pull request text has the same
   defect: "An empty user id breaks the `CHECK` constraint of the `event` table, the whole
   `BEGIN IMMEDIATE` transaction rolls back, and the cursor stays at its old value." That is 27
   words, and BLOCKER 1 shows the claim is wrong. Each other comment reads well. Please split
   each long sentence.

---

**Summary: 1 BLOCKER, 4 MAJOR, 6 MINOR.** The design of the store is correct and the pragmas,
the migration guard, and the schema each work. The transaction of D4 has no test that proves it,
the writer connection has no thread guard, a failed `open()` leaks a file handle, and the later
API issues need a read path. Please fix the BLOCKER and the four MAJOR findings.

xamcross:
## Response to correction round 1

Commit `794068b` fixes each BLOCKER and each MAJOR finding, and each MINOR
finding that this response does not reject. The schema, the pragma list,
and the migration list keep the shape that both reviewers already passed.

### BLOCKER 1 (Kotlin backend engineer): the rollback test could not fail

`monitor/backend/src/test/kotlin/octometer/monitor/store/EventStoreTest.kt:60-71`.

Fix: the bad page now holds a good event (`e2`) first, and the bad event
(`e3`, an empty user id) second. The test asserts `countEvents == 1`, so
only `e1` from the first page survives.

Proof, done in this task: I removed the transaction from
`EventStore.commitPage` (`BEGIN IMMEDIATE`, `COMMIT`, the rollback), ran
`./gradlew :monitor:backend:test --tests "octometer.monitor.store.EventStoreTest"`,
and saw:

```
EventStoreTest > a failed insert leaves the cursor as it was and rolls back the good event too() FAILED
EventStoreTest > the store commits again after one failed page() FAILED
5 tests completed, 2 failed
```

I then restored `EventStore.kt` to the version of commit `794068b` and
confirmed the diff against the restored file was empty. A full rerun of
`./gradlew :monitor:backend:test --rerun-tasks` then gave `BUILD
SUCCESSFUL`.

### MAJOR 1 (SQLite and data engineer): the reader connection could write

`monitor/backend/src/main/kotlin/octometer/monitor/store/SqliteDatabase.kt:96-102`.

Fix: `openConnection` takes a `readOnly` flag and calls
`SQLiteConfig().setReadOnly(readOnly)`. `open()` passes `readOnly = true`
for the reader (`SqliteDatabase.kt:65`).

Test: `PragmaTest.kt:39-50`, "a write on the reader connection fails".
Each of the seven pragmas on each connection: `PragmaTest.kt:52-60`,
run from both `write` (line 27-29) and `read` (line 31-34).

### MAJOR 2 (SQLite and data engineer): no guard for a newer database

`monitor/backend/src/main/kotlin/octometer/monitor/store/MigrationRunner.kt:23-29`.

Fix: `run` computes `latest` and calls `check(current <= latest)` with a
message that names both numbers.

Test: `MigrationRunnerTest.kt:55-71`, "a database that is newer than the
code stops with a clear error".

### MAJOR (Kotlin backend engineer): the writer ran off its own thread

`SqliteDatabase.kt:19-41`. `writer` and `reader` are now `private`
(lines 26-27). `write` (lines 34-35) and `read` (lines 38-41) are the only
entry points. `open()` now runs the migration on the writer dispatcher
with `runBlocking(writerDispatcher) { MigrationRunner.run(writer) }`
(line 71). `EventStore.commitPage` calls `database.write { writer -> ...
}` (`EventStore.kt:36`). Each test now reaches the connections only
through `write`/`read` (for example `EventStoreTest.kt:96`,
`MigrationRunnerTest.kt:24`, `PragmaTest.kt:28`).

### MAJOR (Kotlin backend engineer): a failed open() leaked both connections

`SqliteDatabase.kt:56-87`. Each opened part now sits inside its own
`try`. A throw from the migration closes the dispatcher (lines 72-75), a
throw after that closes the reader (lines 77-82), and a throw after that
closes the writer (lines 83-86).

Test: `SqliteDatabaseTest.kt` (new file), "a failed open closes each
connection it already made". It pre-creates a conflicting `app` table,
calls `SqliteDatabase.open`, and asserts `tempDir.deleteRecursively()`
returns `true`.

### MAJOR (Kotlin backend engineer): BEGIN IMMEDIATE outside the try, a silent ROLLBACK

`EventStore.kt:35-54`. `BEGIN IMMEDIATE` now sits before a
`try`/`catch`/`finally`. The `finally` (lines 48-52) rolls back only when
`committed` is still `false`. `rollback` (lines 80-88) attaches a failed
`ROLLBACK` to the first error with `addSuppressed`, and it never swallows
the original `Throwable` (the `catch` at line 45 always rethrows at line
47, `CancellationException` included).

Tests: `EventStoreTest.kt:60-71` (the rollback itself, see BLOCKER 1) and
`EventStoreTest.kt:73-89`, "the store commits again after one failed
page": a good page, a failed page, then a third good page, with the
cursor and the row count of both good pages asserted.

### MAJOR, with a decision (Kotlin backend engineer): no safe read path

`SqliteDatabase.kt:37-41`. `reader` is `private`. `read` runs the
block on `Dispatchers.IO`, inside `readMutex.withLock { }`, because one
JDBC connection is not safe for two parallel calls. This pull request
keeps one reader connection, per the issue and D3. A connection pool
waits for a later issue with a measurement, per your decision.

Tests: `PragmaTest.kt:31-34` and `:39-50` call `database.read { }`.
`EventStoreTest.kt` reads the row counts and the cursor through `read`
(lines 142-161).

### MINOR findings

1. **`next()` before `getInt`** (both reviewers). Fixed in
   `MigrationRunner.kt:57-63` (`check(result.next())`), and in each test
   helper that reads a pragma or `user_version`:
   `MigrationRunnerTest.kt:73-79`, `PragmaTest.kt:62-76`.
2. **A JDBC batch for the page insert** (SQLite and data engineer). Fixed
   in `EventStore.kt:56-70`: `insert.addBatch()` inside the loop,
   `insert.executeBatch()` once after it.
3. **The replay test does not assert the cursor** (SQLite and data
   engineer). Fixed in `EventStoreTest.kt:46-55`: the test now asserts
   `cursor-2` after the replay.
4. **`PragmaTest` covers two of seven pragmas** (SQLite and data
   engineer). Fixed in `PragmaTest.kt:52-60`: `assertPragmas` now checks
   `journal_mode`, `synchronous`, `busy_timeout`, `foreign_keys`,
   `secure_delete`, `temp_store`, and `cache_size`.
5. **A failed migration leaked the two connections** (SQLite and data
   engineer). Fixed by the same change as MAJOR "a failed open() leaked
   both connections" above (`SqliteDatabase.kt:56-87`); the migration
   throw is one of the three covered failure points.
6. **The result of `journal_mode=WAL` is not checked** (SQLite and data
   engineer). Fixed in `SqlitePragmas.kt:39-47`: `applyJournalMode` reads
   the row back and calls `check(mode.equals("wal", ignoreCase = true))`.
7. **`updateCursor` drops the update count** (SQLite and data engineer).
   Fixed in `EventStore.kt:72-78`: `check(update.executeUpdate() == 1)`.
8. **`ResultSet.getInt` without `next()`** (Kotlin backend engineer). Same
   fix as MINOR 1 above.
9. **`SqliteDatabase.kt` drops the result of `mkdirs()`** (Kotlin backend
   engineer). Fixed in `SqliteDatabase.kt:58-60`:
   `check(folder.mkdirs() || folder.isDirectory)`.
10. **A cancel does not roll back** (Kotlin backend engineer). This is a
    documentation ask, not a code defect: the store already rethrows
    `CancellationException`. Fixed with the KDoc at `EventStore.kt:29-31`.
11. **The error type is the raw driver type** (Kotlin backend engineer).
    **Rejected**, with evidence: the finding text itself reads "PASS on
    that point" for the logging concern, and frames `StoreException` as a
    future need for issues #28 and #58, which own the status mapping of
    D8. `org.sqlite.SQLiteException` already exposes `resultCode`, and
    issue #9 does not touch the poller or the status. I leave this for
    whichever of #28 or #58 first needs the mapping.
12. **Two parts of D3 have no confirmed owner** (Kotlin backend engineer).
    Named, not fixed, per the coordinator's decision: the lock file
    `octometer.lock` and `PRAGMA quick_check` stay out of this pull
    request. Issue #58, "Test the store against a corrupt, a locked, and
    a full database", is the closest owner in the backlog.
13. **Long sentences in three places** (Kotlin backend engineer). Fixed
    in `MigrationRunner.kt:13-17`, `SqliteDatabase.kt:18-24`, and in the
    pull request text (the "Confirm" and the acceptance criteria
    sections).

### Full test run after the fix

`./gradlew :monitor:backend:test --rerun-tasks` with `JAVA_HOME` set to
the JDK 21 install: `BUILD SUCCESSFUL`, 27 tests, 0 failures, 0 errors,
across `HealthRouteTest` (2), `ConfigLoaderTest` (13), `EventStoreTest`
(5), `MigrationRunnerTest` (3), `PragmaTest` (3), and the new
`SqliteDatabaseTest` (1).

xamcross:
**Review (second review, SQLite and data engineer): APPROVE**

I made a new clone, checked out `issue-9-sqlite-store` at `794068b`, and tested the code, not
the response comment. `./gradlew :monitor:backend:test` with JDK 21: `BUILD SUCCESSFUL`, 27
tests, 0 failures, 0 errors (`--rerun-tasks`). I wrote nine probe tests and two mutations, and
I deleted each of them after the run. The DDL file `001_init.sql` is unchanged since `404b644`.

## My round 1 findings

### MAJOR-1 — the reader connection can write. **CORRECTED.**
`SqliteDatabase.kt:96-102` opens the connection with `SQLiteConfig().setReadOnly(readOnly)`.
I ran four write shapes through `database.read {}`. Each one failed:
```
refused 'INSERT INTO gap (app_id, from_ts, to_ts) VALUES (1, 1, 2)' -> [SQLITE_READONLY] ...
refused 'CREATE TABLE sneaky (x INTEGER) STRICT'                    -> [SQLITE_READONLY] ...
refused 'PRAGMA user_version = 42'                                  -> [SQLITE_READONLY] ...
refused 'DELETE FROM app'                                           -> [SQLITE_READONLY] ...
```
`user_version` on the reader stayed at 1. The seven pragmas read back the correct value on
**both** connections: `journal_mode=wal`, `synchronous=1`, `busy_timeout=5000`,
`foreign_keys=1`, `secure_delete=1`, `temp_store=2`, `cache_size=-65536`.

A read during a writer transaction does not block. I opened `BEGIN IMMEDIATE` on the writer,
inserted a row, and then read from the reader: **0 rows in 0 ms**. The reader saw the old
snapshot and waited for nothing.

A new file works. I called `open` on an empty folder. The read-only connection attached to the
file that the writer made a moment before, on Windows: `journal_mode=wal`, `tables=5`,
`user_version=1`, with `octometer.db` at 4096 bytes, `octometer.db-wal` at 45352 bytes, and
`octometer.db-shm` present. The order in `open` makes this safe: the writer runs
`PRAGMA journal_mode=WAL` first, thus the header and the `-shm` file exist before the
read-only connection attaches. Keep that order.

### MAJOR-2 — the runner starts on a newer database. **CORRECTED.**
`MigrationRunner.kt:23-29`. I set `PRAGMA user_version = 7` on a good database and started the
code again:
```
java.lang.IllegalStateException: The database is at user_version 7, and this build knows 1.
Install a newer monitor build, or restore a backup.
```
The file stays as it is. I compared the SHA-256 of `octometer.db` before and after the failed
start: the two are equal. The folder held only `octometer.db` before and after, thus the
failed start left no `-wal` and no `-shm` file. The folder then deleted with no error, thus no
Windows handle stayed open. `MigrationRunnerTest.kt:56-71` covers it.

### MINOR-1 — a pragma read without `next()`. **CORRECTED.** `MigrationRunner.kt:57-63` calls
`check(result.next())` with a clear message.
### MINOR-2 — no JDBC batch. **CORRECTED.** `EventStore.kt:66-68` uses `addBatch` plus
`executeBatch`. The commit count is still one. I read `PRAGMA data_version` on the reader:
`3 -> 4 -> 5`. A page of 3 rows moved it by 1, and a page of 1000 rows also moved it by 1.
1003 rows landed, and the cursor reached `c-big`.
### MINOR-3 — the replay test skips the cursor. **CORRECTED.** `EventStoreTest.kt:47-55` now
asserts `cursor-2`. My probe agrees: 2 rows, cursor `cursor-2`.
### MINOR-4 — `PragmaTest` covers two pragmas. **CORRECTED.** `PragmaTest.kt:52-60` asserts all
seven, and `PragmaTest.kt:27-34` runs the same helper on the writer and on the reader.
### MINOR-5 — a failed migration leaks the connections. **CORRECTED.**
`SqliteDatabase.kt:63-87` closes each part that it already opened. My probe in MAJOR-2 shows
the folder deletes. `SqliteDatabaseTest.kt:17-35` covers it.
### MINOR-6 — the `journal_mode` answer is unchecked. **CORRECTED.** `SqlitePragmas.kt:39-47`
reads the answer and stops on a value other than `wal`.
### MINOR-7 — `updateCursor` drops the update count. **CORRECTED.** `EventStore.kt:76`. My
probe: `commitPage` on a missing app id gave
`java.lang.IllegalStateException: No app row for id 987654.`, and the writer worked again
after it.

## The transaction of `commitPage`

- One page is one commit, with the batch. See MINOR-2 above.
- A replay adds 0 rows and moves the cursor. See MINOR-3 above.
- The store works again after one failed page. My probe held a rival `BEGIN IMMEDIATE` from a
  second connection. `commitPage` threw
  `org.sqlite.SQLiteException: [SQLITE_BUSY] The database file is locked` after **5529 ms**,
  thus `busy_timeout=5000` works. The next page then committed, and the cursor moved to
  `c-after`. `EventStoreTest.kt:77-89` covers the same rule for a constraint failure.
- `BEGIN IMMEDIATE` sits at `EventStore.kt:37`, one line **above** the `try`. That is correct,
  and I prefer it: a failed `BEGIN` leaves no transaction, thus a `ROLLBACK` in the `finally`
  would only add a second error. My `SQLITE_BUSY` probe proves the path: one error reached the
  caller, and the connection stayed clean. See the MINOR below on the commit text.
- The rollback runs in the `finally` at `EventStore.kt:48-52`, and a failed `ROLLBACK` goes to
  `addSuppressed` at line 86. No error stays silent.

## The mutations (proof that the tests are not always green)

1. I removed `BEGIN IMMEDIATE`, `COMMIT`, and the rollback from `commitPage`. Two tests
   failed: `a failed insert leaves the cursor as it was and rolls back the good event too`
   (`expected: <1> but was: <2>`) and `the store commits again after one failed page`
   (`expected: <2> but was: <3>`). Without the transaction the good event of the bad page
   stays, because `executeBatch` in autocommit mode commits the rows before the bad one.
2. I changed the cursor statement to `UPDATE app SET cursor = COALESCE(cursor, ?) WHERE id = ?`,
   so the cursor moves once and never again. The corrected replay test failed:
   `expected: <cursor-2> but was: <cursor-1>`. The new cursor assertion is real.

I restored the file after each mutation. `git status` is clean.

## The planned migration 002, through `write {}`

I ran the planned shape through the new API, inside one transaction on the writer thread:
three `ALTER TABLE event ADD COLUMN` statements (`path TEXT`, `referrer_host TEXT`,
`kind INTEGER NOT NULL DEFAULT 0`), `DROP INDEX event_session`, one partial index with a
`WHERE` clause, and one plain index. Each statement ran. The result: `event` has 9 columns, 3
indexes with the prefix `event`, and `user_version=2`. The `write {}` API adds no limit here.
A `STRICT` table accepts `ADD COLUMN` with a type of the `STRICT` list, and a `NOT NULL` column
needs the `DEFAULT` that `kind` has.

## The one new MINOR

**MINOR-8. The commit body of `794068b` does not match the code.** The body says "Move BEGIN
IMMEDIATE inside a try". `EventStore.kt:37` keeps it one line above the `try`. The code is
correct, and the text is not. Fix: on the next commit of this branch, write "Keep BEGIN
IMMEDIATE above the try, and roll back in a finally". This does not block the merge.

## The rest

The scope is correct: the store only, with no route, no poll cycle, and no MongoDB reader. The
diff holds 11 files, and no `.db`, no `-wal`, no `-shm`, and no build output. Each test uses
its own `Files.createTempDirectory` folder. `001_init.sql` still equals section 6, character
for character. `sqlite-jdbc` stays at `3.53.4.0`, the release on Maven Central. The Windows CI
job is green. The language of each comment and of the commit body follows ASD-STE100.

My note of round 1 for issue #61 stays true: `secure_delete=ON` alone is not an erasure. An
erasure needs `DELETE` plus `PRAGMA wal_checkpoint(TRUNCATE)`.

Result: 0 BLOCKER, 0 MAJOR, 1 new MINOR. Both MAJOR findings of round 1 are corrected, and each
of my 7 MINOR findings is corrected. I approve the merge.

xamcross:
**Review (second review, Kotlin backend engineer): APPROVE**

I made a new clone at the head of `issue-9-sqlite-store` (`794068b`). I checked each finding
against the code, not against the response comment. I ran `./gradlew :monitor:backend:test` with
JDK 21: `BUILD SUCCESSFUL`, **27 tests, 0 failures**. I wrote probe tests, I read their output,
and I deleted them. The working tree is clean.

---

## The first findings

| Finding | Result |
|---|---|
| BLOCKER 1. The cursor test cannot fail | **CORRECTED** |
| MAJOR 2. The writer connection runs off the dispatcher | **CORRECTED** |
| MAJOR 3. A failed `open()` leaks the connections | **CORRECTED** |
| MAJOR 4. `BEGIN IMMEDIATE` outside the `try`, a silent `ROLLBACK` failure | **CORRECTED** (one small residual below) |
| MAJOR 5. No safe read path | **CORRECTED**, against the decision of the maintainer |
| MINOR 1. `getInt` without `next()` | **CORRECTED** |
| MINOR 2. The result of `mkdirs()` | **CORRECTED** |
| MINOR 3. A cancel does not roll back | **CORRECTED** (the KDoc states it) |
| MINOR 4. A custom `StoreException` | **REJECTED by the author. I accept the reason.** |
| MINOR 5. No owner for the lock file and `quick_check` | **RESOLVED.** #58 owns both. |
| MINOR 6. Three long comment sentences | **CORRECTED** |

### BLOCKER 1 — CORRECTED

`EventStoreTest.kt:61-71`. The bad page now holds a good event first. I removed
`BEGIN IMMEDIATE`, `COMMIT`, and the `rollback` call from `EventStore.kt`, and I ran the class:

```
EventStoreTest > a failed insert leaves the cursor as it was and rolls back the good event too() FAILED
EventStoreTest > the store commits again after one failed page() FAILED
```

The test can fail now. I restored the file.

### MAJOR 2 — CORRECTED

`SqliteDatabase.kt:26-27` makes both connections private. `write` on line 34 is the one path to
the writer. Line 71 runs the migration on the writer thread.

My probe ran 480 `write` blocks from 8 coroutines on `Dispatchers.IO`:

```
PROBE_A overlaps=0 threads=[octometer-sqlite-writer]
```

Each block ran on the one writer thread. No two blocks overlapped, because `block` is not a
suspend lambda, thus a block cannot yield midway. A second thread can no longer reach the
connection: the field is private and has no getter. The read-only reader (line 65) adds a second
guard at the driver:

```
PROBE_F readerWrite=[SQLITE_READONLY] Attempt to write a readonly database
PROBE_F readerIsReadOnly=true
```

### MAJOR 3 — CORRECTED

`SqliteDatabase.kt:63-86` closes each opened part on a throw. I removed the two
`runCatching { ... close() }` calls, and the new test failed:

```
SqliteDatabaseTest > a failed open closes each connection it already made() FAILED
```

I restored the file. On the present code the folder deletes with no locked handle.

### MAJOR 4 — CORRECTED

`EventStore.kt:48-52` rolls back in a `finally`. `EventStore.kt:82-88` attaches a failed
`ROLLBACK` to the first error with `addSuppressed`, thus no failure stays silent.
`catch (error: Throwable)` on line 45 throws the error again on line 47, thus a
`CancellationException` is not swallowed.

I removed only the `rollback` call, and the new test failed:

```
EventStoreTest > the store commits again after one failed page() FAILED
```

My probe ran three failed pages, each with one good page after it:

```
PROBE_E round=0 bad=SQLiteException good=OK
PROBE_E round=1 bad=SQLiteException good=OK
PROBE_E round=2 bad=SQLiteException good=OK
PROBE_E autoCommit=true events=3 (3 means every bad page rolled back)
```

Two `commitPage` calls in sequence after one failure both work.

A cancel behaves as the KDoc on lines 29-31 now states:

```
PROBE_D jobCancelled=true cursorAfterCancel=cursor-cancelled events=80000 nextCommit=OK
```

The page finishes on the writer thread, the caller still gets the `CancellationException`, and
the connection stays outside a transaction.

### MAJOR 5 — CORRECTED against the decision of the maintainer

The decision: one read connection, a `Mutex`, `Dispatchers.IO`, and no pool, because D3 names one
read connection. `SqliteDatabase.kt:31` and `:38-41` give exactly that form. I judge it against
that decision, and it passes.

1. **No two coroutines use the connection at the same time.** 480 read blocks from 8 coroutines:
   `PROBE_B overlaps=0 errors=0`.
2. **The block runs off the caller thread.**
   `IOPROBE caller=Test worker read=DefaultDispatcher-worker-1 write=octometer-sqlite-writer`.
   `Dispatchers.IO` shares the worker names of the default pool, so the name is correct.
3. **A cancelled waiter leaves the mutex free.** I held the lock, cancelled a second caller while
   it waited, then took the lock with a third caller:
   `PROBE_C waiterCancelled=true afterCancelAcquire=acquired`.
4. **A `read` inside a `write` cannot give a deadlock.** The guard is better than a KDoc line: it
   is the compiler. `write` takes `(Connection) -> T`, which is not a suspend lambda, thus a
   `read` call inside it does not compile. I proved it:

   ```
   e: DeadlockProbe.kt:11:22 Suspension functions can only be called within coroutine body.
   ```

   The same rule blocks a `write` inside a `read`. Please keep both parameters non-suspend.

**Accepted limit.** The mutex serialises each read. #18 measured 0.33 s for the level 1 query at
1 million events, thus ten parallel requests make the last one wait about 3.3 s. The maintainer
accepts this, and a later issue adds a pool with a measurement.

### MINOR 4 — I accept the rejection

The author keeps the driver exception and names #28 and #58 as the owners of the status mapping
of D8. **The reason holds.** `org.sqlite.SQLiteException.getResultCode()` already returns
`SQLITE_FULL`, thus #58 can separate a full disk with no new type. The module already depends on
`sqlite-jdbc`, thus the import adds no coupling. I withdraw the finding.

### MINOR 5 — resolved

#58 step 1 owns `PRAGMA quick_check`, and step 2 owns `octometer.lock` with
`FileChannel.tryLock()`. I read the issue body. Both have an owner.

---

# New MINOR findings

None of these prevents the approval.

1. **`EventStore.kt:37`: `BEGIN IMMEDIATE` still sits outside the `try`.** The commit message
   says "Move BEGIN IMMEDIATE inside a try". The code does not do that. The substance of MAJOR 4
   is correct, because the `finally` covers each failure after the `BEGIN`. One narrow path
   stays: a `ROLLBACK` that fails leaves the transaction open, thus each later `commitPage`
   throws at line 37 with no recovery. Suggested fix: when `rollback` fails, set a `broken` flag
   and throw a clear message on the next call, instead of the opaque "cannot start a transaction
   within a transaction". Please correct the commit message of a later round.
2. **No test holds the writer-thread rule.** My probe proves it today, and the private field
   makes a breach hard. A small test locks the rule in:
   `assertEquals("octometer-sqlite-writer", database.write { Thread.currentThread().name.substringBefore(" @") })`.
3. **`EventStore.kt:76` adds a second exception type.** `check(update.executeUpdate() == 1)`
   throws `IllegalStateException`, while each other failure of `commitPage` is a `SQLException`.
   A caller that writes `catch (error: SQLException)` misses a missing `app` row. Please name both
   types in the KDoc of `commitPage`.
4. **`MigrationRunner.kt:47-52`: `catch (error: Exception)` with `finally { autoCommit = true }`.**
   An `Error` skips the `catch`, and the `finally` then commits the part of the DDL that already
   ran. Use `catch (error: Throwable)`, as `EventStore` now does.
5. **Language.** `SqliteDatabase.kt:18-24` holds seven sentences in one paragraph. The limit is
   six. `SqlitePragmas.kt:38` joins two instructions with "and": "Read the answer, and stop on a
   bad value." Each other new comment reads well, and the long sentence of
   `MigrationRunner.kt` is now three short sentences.

---

**Summary: 0 BLOCKER, 0 MAJOR, 5 MINOR.** The BLOCKER and each of my four MAJOR findings are
corrected, and I proved each one with a break test or a probe. The rejected MINOR has a sound
reason, and I withdraw it. The store is ready for #17, #18, #50, and #51. **APPROVE.**

Issue #38, opened 2026-09-21
Bundle the frontend into the monitor distribution


xamcross:
**Goal.** One distribution zip holds the monitor server and the Angular build. It starts on Windows with only a JDK 21.

**Implementation steps.**
1. Add a Gradle task that builds the Angular app of `monitor/frontend` into the static resources of `monitor/backend`.
2. Make `installDist` and `distZip` depend on that task.
3. Set the class path of the Windows start script to the wildcard `lib\*`.
4. Serve the Angular build at `/`, with `index.html` as the response to a deep link such as `/apps/1/users`.
5. Extend the `monitor-windows` job in `.github/workflows/ci.yml`. The job calls `installDist`, starts the script, reads `/api/health`, and opens `/`.

**Acceptance criteria.**
- [x] The zip of `./gradlew :monitor:backend:distZip` starts on Windows with only a JDK 21.
- [x] `/` serves the Angular app.
- [x] A deep link such as `/apps/1/users` returns `index.html`.
- [x] The Host check covers the static files: `Host: evil.example` gets 403 on `/`.
- [x] A browser opens `/` and the console shows no CSP error.
- [x] `./gradlew :monitor:backend:test` passes.
- [x] The `monitor-windows` CI job is green.
- [x] A comment on this issue records the URL of the CI run.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, D35 (first part), D39.

xamcross:
Body edit on 2026-09-21: #6 closed. The first line now names only the open blockers: #5 and #7.

xamcross:
Body edit on 2026-09-21: #7 closed. The first line now names only the open blockers: #5.

xamcross:
Body edit on 2026-09-21: #5 closed. No open blocker remains, thus the first line and the label `blocked` are removed. The work can start.

xamcross:
Closed on 2026-09-22 through pull request #163 (squash merge, commit 7be3c40). The maintainer ticked each acceptance criterion on the evidence of the three review rounds:
- The zip of `./gradlew :monitor:backend:distZip` starts on Windows with only a JDK 21: each of the two reviewers started `bin\backend.bat` from a path with a space, with JDK 21 only on the PATH.
- `/` serves the Angular app: `GET /` gave 200 with the page and the CSP header of D12.
- A deep link returns `index.html`: the reviewers opened `/apps/1/users` in a real Chrome and through the route tests.
[...]

PR #97, opened 2026-09-21
feat(monitor): reject a request with a wrong Host or Origin header

### LINKED ISSUE #5: Reject a request with a wrong Host or Origin header
**Goal.** The monitor rejects a request with a wrong `Host` or `Origin` header.

**Why it blocks production.** Without the checks a hostile web page reads personal data through a DNS rebind attack, or changes the registry through CSRF.

**Implementation steps.**
1. Add a Host check on each route.
2. Add an `Origin` check on each request with a method other than GET and HEAD.
3. In dev mode, allow the `ng serve` port 4200 in the two checks.
4. Require `Content-Type: application/json` on each such request that holds a body. A content type that the server cannot parse gets 415.
5. Send the three headers of D12, with the CSP `default-src 'self'; style-src 'self' 'unsafe-inline'; object-src 'none'; base-uri 'self'; frame-ancestors 'none'`.
6. Refuse a non-loopback bind address with a clear error.

**Acceptance criteria.**
- [x] `Host: evil.example` gets 403 on each route. `localhost:<port>` and `127.0.0.1:<port>` pass.
- [x] A POST, PUT, PATCH, or DELETE without an allowed `Origin` header gets 403 and changes no data. A route with a different method, for example `FOO`, gets the same check.
- [x] A DELETE with an allowed `Origin` header and without a body passes the content type rule.
- [x] A POST with `Content-Type: text/plain` gets 415.
- [x] `http://localhost:4200` passes in dev mode and gets 403 in prod mode.
- [x] No response holds `Access-Control-Allow-Origin`.
- [x] Each response holds the CSP of step 5 and `X-Content-Type-Options: nosniff`. Each API response holds `Cache-Control: no-store`, also for the paths `//api/health` and `/%61pi/health`.
- [x] A non-loopback bind address stops the server with a clear error.
- [x] `./gradlew :monitor:backend:test` passes with `testApplication` cases.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, D12.

xamcross:
## Summary

The monitor now rejects a request with a wrong `Host` or `Origin` header, per
D12. One interceptor runs before each route. It checks the `Host` header on
each route, the `Origin` header on each request with a method other than GET
and HEAD (an allow-list, thus a later route on a new method gets the check
too), and the `Content-Type` header on a request that holds a body. It adds
the CSP, `X-Content-Type-Options`, and `Cache-Control` headers. The server
installs no CORS plugin.

This text replaces the first version, after correction round 1 (the two
reviews on pull request 97). The response comment on this pull request lists
each finding with its fix.

## Decisions

- **The IPv6 loopback form.** D12 names two Host forms, `localhost:<port>`
  and `127.0.0.1:<port>`, plus the `ng serve` port in dev mode. It names no
  `[::1]:<port>` form. Decision: the monitor refuses `[::1]:<port>`. Source:
  D12, `docs/superpowers/specs/2026-09-21-octometer-design.md`. The security
  reviewer's raw-socket probe confirms a connect to `[::1]:<port>` gets
  `Connection refused`, because the server binds `127.0.0.1` only.
- **The bind address of step 6.** D2 names five config keys, and none of
  them is a bind address. The code has no bind-address config value: the
  constant `HOST` in `Application.kt` stays at `"127.0.0.1"`, with no way to
  set a different value today. Step 6 is enforced by
  `requireLoopbackBindAddress(HOST)` in `main()`, unit-tested directly
  (`BindAddressTest`). Correction: the accepted set now holds `127.0.0.1`
  only, not `localhost` too, because a bind on the name `localhost` can
  resolve to the IPv6 loopback address first on Windows (MINOR 2 of the
  security review).
- **The plugin form (MINOR-1 of the Ktor review): rejected, with evidence.**
  The reviewer proposed `createApplicationPlugin` with `onCall`, the
  documented Ktor 3 form, and called the present form "correct... thus this
  is a preference only." I built a throwaway probe with
  `createApplicationPlugin` and confirmed `call.respond()` inside `onCall`
  does stop the route handler, with no explicit `finish()` call needed, on
  Ktor 3.6.0. I did not adopt it, for one reason: `onCall` always attaches to
  the `Plugins` phase, and the security review's MINOR 1 asks for the
  earliest phase, `Setup`, so that a plugin a later change installs cannot
  run before the guard. Only the raw `intercept(phase)` call, kept in this
  pull request, can choose that phase. I moved the interceptor to
  `ApplicationCallPipeline.Setup`.

## Acceptance criteria (issue #5, after the maintainer's edit of 2026-09-21)

- [x] `Host: evil.example` gets 403 on each route. `localhost:<port>` and
  `127.0.0.1:<port>` pass. Evidence: `RequestGuardTest` — "a wrong Host
  header gets 403 on the health route", "the Host localhost with the
  configured port passes", "the Host 127-0-0-1 with the configured port
  passes", "an upper-case Host passes...", "an absent Host header gets 403",
  "an unknown route still gets the Host check...", and "a route added only
  in this test still gets the Host check, with no new code at the route".
- [x] A POST, PUT, PATCH, or DELETE without an allowed `Origin` header gets
  403 and changes no data. A route with a different method, for example
  `FOO`, gets the same check. Evidence: `RequestGuardTest` — "a POST without
  an allowed Origin header gets 403 and changes no data" (a counter stays at
  0), "a PUT, a PATCH, and a DELETE without an allowed Origin header each
  get 403", and "a route on a custom method without an allowed Origin
  header gets 403" (a `FOO` route, `handlerRan` stays `false`). The check is
  now an allow-list of the safe methods (GET, HEAD), per MAJOR 3 of the
  security review.
- [x] A DELETE with an allowed `Origin` header and without a body passes the
  content type rule. Evidence: `RequestGuardTest` — "a DELETE with an
  allowed Origin and without a body passes the content type rule"
  (`handlerRan` becomes `true`), and "a DELETE with an allowed Origin and a
  text-plain body gets 415". This criterion, and the content-type-with-a-body
  rule behind it, is the maintainer's decision after MAJOR (Ktor review) /
  MAJOR 4: a DELETE with no body needs no `Content-Type` header.
- [x] A POST with `Content-Type: text/plain` gets 415. Evidence:
  `RequestGuardTest` — "a POST with Content-Type text-plain gets 415", and
  "a POST with application-x-www-form-urlencoded gets 415". A malformed
  value (for example `)(`) also gets 415, with no echo of the header value
  in the body: `RequestGuardRawSocketTest` — "a malformed Content-Type gets
  415, and the body does not echo the header value" (MAJOR 2 of the security
  review; the Ktor test client itself refuses to send a malformed value,
  thus this test starts a real server and uses a raw socket).
- [x] `http://localhost:4200` passes in dev mode and gets 403 in prod mode.
  Evidence: `RequestGuardTest` — "the ng serve origin passes in dev mode",
  and "the ng serve origin gets 403 in prod mode".
- [x] No response holds `Access-Control-Allow-Origin`. Evidence:
  `RequestGuardTest` — "no response holds Access-Control-Allow-Origin", and
  "an OPTIONS request gets no Access-Control-Allow-Origin, also with an
  allowed Origin" (MAJOR 3 of the security review named this case). The
  server installs no CORS plugin.
- [x] Each response holds the CSP of step 5 and `X-Content-Type-Options:
  nosniff`. Each API response holds `Cache-Control: no-store`, also for the
  paths `//api/health` and `/%61pi/health`. Evidence: `RequestGuardTest` —
  "each response holds the CSP and X-Content-Type-Options, and the API
  response holds Cache-Control no-store", "a rejected request holds the
  CSP, X-Content-Type-Options, and Cache-Control too", "a percent-encoded
  api segment still gets Cache-Control no-store", and "a non-api path gets
  no Cache-Control header". `RequestGuardRawSocketTest` — "a repeated slash
  in the path still gets Cache-Control no-store", and "a path with a bad
  percent escape fails closed...". (MAJOR 1 of the security review, MAJOR-2
  of the Ktor review; both edge cases need a raw socket, because the Ktor
  test client rewrites `//api/health` into a different host and refuses to
  send `/%zz` unchanged.)
- [x] A non-loopback bind address stops the server with a clear error.
  Evidence: `BindAddressTest` — "the loopback address of D12 passes", "the
  name localhost is not the address of D12, thus it gives a clear error",
  and "a non-loopback address gives a clear error".
- [x] `./gradlew :monitor:backend:test` passes with `testApplication` cases.
  Evidence: run on 2026-09-21 with `JAVA_HOME` set to the JDK 21 install,
  after `git merge origin/main` (issue #9 merged first; no shared file, no
  conflict). Result: `BUILD SUCCESSFUL`, 60 tests, 0 failures, 0 errors.
  The tests of this issue: `HealthRouteTest` 2, `ConfigLoaderTest` 13,
  `BindAddressTest` 3, `RequestGuardTest` 26, `RequestGuardRawSocketTest` 4
  (48 total). The other 12 tests are the SQLite store tests of issue #9,
  merged from `main`.
- [x] A comment on this issue records the test output. Evidence: see the
  issue comment linked to this pull request, and the correction-round
  comment for this update.

## Design document

`docs/superpowers/specs/2026-09-21-octometer-design.md`, D12: one sentence
for the allow-list of the safe methods (GET, HEAD), and one sentence for the
content type rule on a request that holds a body, per the maintainer's
decision 5.

## Confirm steps

The issue names no "Confirm" step.

Closes #5

Generated with Claude Code


xamcross:
**Review (Ktor backend engineer, first review): CHANGES REQUESTED**

## How I checked

I cloned the repository to my own folder and checked out the head of
`issue-5-host-origin-checks` (`edf8df0`). I ran
`./gradlew :monitor:backend:test` with `JAVA_HOME` on the JDK 21 install.
Result: `BUILD SUCCESSFUL`, 30 tests, 0 failures (`HealthRouteTest` 2,
`ConfigLoaderTest` 13, `BindAddressTest` 2, `RequestGuardTest` 13).

I broke the production code four times, and each break made the correct test
fail. I restored the file after each break.

| The break | The failed tests |
| --- | --- |
| `RequestGuard.kt:54`, the Host set test removed | 2 tests |
| `RequestGuard.kt:49`, the API path prefix changed | 1 test |
| `RequestGuard.kt:93`, the dev mode condition removed | 1 test |
| `RequestGuard.kt:62`, the missing Origin accepted | 2 tests |

I also ran my own probe tests against the branch. I deleted each probe file
after the run. The branch and the clone hold no change from me.

## The acceptance criteria

1. **`Host: evil.example` gets 403 on each route. PASS.** `RequestGuardTest:37`
   and `RequestGuardTest:64`. My probe: `GET /no-such-path` with
   `Host: evil.example` gives 403, thus a path with no route gets the check too.
2. **A mutating method without an allowed Origin gets 403 and changes no data.
   PASS.** `RequestGuardTest:81` asserts the counter stays at 0.
   `RequestGuardTest:104` covers PUT, PATCH, and DELETE.
3. **A POST with `Content-Type: text/plain` gets 415. PASS.**
   `RequestGuardTest:154`.
4. **`http://localhost:4200` passes in dev mode and gets 403 in prod mode.
   PASS.** `RequestGuardTest:173` and `RequestGuardTest:192`. The two tests use
   the two modes of the config loader.
5. **No response holds `Access-Control-Allow-Origin`. PASS.**
   `RequestGuardTest:211`. The module installs no CORS plugin.
6. **Each response holds the CSP and `X-Content-Type-Options`. Each API
   response holds `Cache-Control: no-store`. PASS on a normal path, FAIL on an
   encoded path.** `RequestGuardTest:220` proves the normal path. My probe:
   `GET /%61pi/health` returns the health JSON with status 200 and no
   `Cache-Control` header. See MAJOR-2.
7. **A non-loopback bind address stops the server with a clear error. PASS as a
   unit check.** `BindAddressTest` checks the exact message. See MINOR-2.
8. **`./gradlew :monitor:backend:test` passes with `testApplication` cases.
   PASS.** My own run gives `BUILD SUCCESSFUL`. No test binds a real port,
   because `testApplication` uses the in-process engine.
9. **A comment on the issue records the test output. PASS.**

## The Ktor design: what is correct

- **The phase is correct.** `ApplicationCallPipeline.Plugins` runs before the
  `Call` phase, thus before the routing plugin resolves a route. My probes
  confirm four cases: a path with no route gets the 403, a 404 of an allowed
  Host holds the security headers, a route that the test adds gets the check
  with no new code, and a rejected request never reaches a handler, thus the
  server reads no body.
- **An error response keeps the headers.** I added a route that throws. Without
  a `StatusPages` plugin the response is 500 and it holds the CSP, `nosniff`,
  and `no-store`. With a `StatusPages` plugin that answers 500 the result is the
  same. The reason: the guard writes the headers into the same response object,
  and the engine commits that object one time.
- **`call.respond` plus `finish()` is the correct pair here.** Without
  `finish()` the routing plugin answers a second time and the call fails.
- **The IPv6 decision is sound.** The server binds only `127.0.0.1`, thus a
  literal `http://[::1]:<port>` URL never connects. The refusal of
  `[::1]:<port>` costs nothing.
- **The scope is clean.** The pull request changes 6 files, all in
  `monitor/backend`. It adds no dependency, no CORS plugin, no secret, and no
  build output. Pull request #96 (issue #9) changes `gradle/libs.versions.toml`,
  `monitor/backend/build.gradle.kts`, and files under
  `monitor/backend/src/**/octometer/monitor/store/`. **The two pull requests
  share no file, thus the merge order is free.**
- **The language is correct.** Each new comment and the pull request text follow
  ASD-STE100. I found no defect.

## The findings

### MAJOR-1 — a DELETE without a body gets 415

`monitor/backend/src/main/kotlin/octometer/monitor/security/RequestGuard.kt:67`

A DELETE with an allowed Host, an allowed Origin, and no body gets 415. My
probe returns `415` with the body
`{"error":"Set the Content-Type header to application/json."}`. The reason:
`call.request.contentType()` returns `*/*` when the header is absent.
`RequestGuardTest:104` hides this, because its DELETE fails at the Origin check
first.

D13 names `DELETE /api/apps/{id}` and
`DELETE /api/apps/{appId}/events?userId=<id>`. Neither one carries a body. The
Angular `HttpClient.delete(url)` sends no `Content-Type` header, thus each
delete of the UI will fail with 415.

The code follows the text of step 4 and of D12, thus the conflict is in the
text, not in the work of the author. The Content-Type rule stops a CSRF form
post, because an HTML form cannot send `application/json`. An HTML form sends
only GET or POST, thus the rule adds no protection on DELETE.

Exact fix (I applied it in my clone; the 30 tests stay green):

```kotlin
val hasBody = (call.request.header(HttpHeaders.ContentLength)?.toLongOrNull() ?: 0L) > 0L ||
    call.request.header(HttpHeaders.TransferEncoding) != null
if (hasBody && call.request.contentType().withoutParameters() != ContentType.Application.Json) {
```

Add two tests: a DELETE with an allowed Origin and no body reaches the route; a
DELETE with a `text/plain` body gets 415. **The maintainer decides.** My
recommendation: take the fix, then record the exception in D12 and in issue #5,
in one sentence: "A request without a body needs no Content-Type header."

### MAJOR-2 — the API rule reads the raw path

`monitor/backend/src/main/kotlin/octometer/monitor/security/RequestGuard.kt:49`

`call.request.path()` returns the path before the decode, and the routing plugin
matches the path after the decode. The prefix test also matches a path that is
not the API. Two probe results show the two defects:

- `GET /%61pi/health` returns the health JSON, status 200, and no
  `Cache-Control` header. A browser or a proxy may store that API response.
- A route `/apidocs.html` gets `Cache-Control: no-store`. A static file route of
  issue #38 must not get the header, because D12 puts `no-store` on the API
  only.

Exact fix (I applied it in my clone; the 30 tests stay green, and both probes
give the correct result):

```kotlin
import io.ktor.http.decodeURLPart
// ...
val path = call.request.path().decodeURLPart()
if (path == "/api" || path.startsWith("/api/")) {
    call.response.header(HttpHeaders.CacheControl, "no-store")
}
```

Add a test for a path outside the API with
`assertNull(response.headers[HttpHeaders.CacheControl])`. That test guards the
static file route of issue #38.

### MINOR-1 — the plugin form of Ktor 3

`RequestGuard.kt:46`. `createApplicationPlugin` with `onCall` is the documented
form of Ktor 3. I measured it: a plugin that answers inside `onCall` stops the
call, and the route does not run. The form gives a name in a stack trace, a
config class, and the hooks such as `on(ResponseSent)`. The present form works
and it is correct, thus this is a preference only.

### MINOR-2 — the bind-address check drops its return value

`Application.kt:38` and `Application.kt:47`. `requireLoopbackBindAddress(HOST)`
returns the address, and the code ignores it. `embeddedServer` then reads the
constant again. A later bind-address key can reach `embeddedServer` and miss the
check. Fix: delete the call at line 38 and write `host = requireLoopbackBindAddress(HOST)`
at line 47. Then the check sits on the value that the server uses.

### MINOR-3 — the error text of a bind address names the config

`Application.kt:41`. The catch block logs "The config is invalid." for a bind
address that comes from a constant, not from the config. Add a second catch, or
move the call out of the `try` block.

### MINOR-4 — a test name promises two results

`RequestGuardTest:173`. The name says "passes in dev mode and gets 403 in prod
mode". The body tests the dev mode only. The prod mode has its own test at line
192. Rename the test to "the ng serve origin passes in dev mode".

### MINOR-5 — a weak assertion on a rejected request

`RequestGuardTest:236`. The test checks `X-Content-Type-Options` only. Add the
CSP assertion and the `Cache-Control` assertion, because a 403 of an API path
must hold the three headers too.

### MINOR-6 — the mode literal

`RequestGuard.kt:84` and `RequestGuard.kt:93` compare with the string `"dev"`.
The enum `Mode` holds the value. Write `config.mode == Mode.DEV.value`.

### MINOR-7 — the duplicated config helpers

`RequestGuardTest:244` and `HealthRouteTest` hold the same `devConfig()` and
`prodConfig()` helpers. Move them to one test helper file. Add a helper for the
allowed Host header there, because each later test must send that header.

### MINOR-8 — a port of 80 or 443 breaks the Host check

`RequestGuard.kt:83`. A browser omits the port in the `Host` header for port 80
and for port 443. A user who sets `OCTOMETER_PORT=80` then gets 403 on each
request. Add the bare host form for those two ports, or name the limit in the
run guide.

## Summary

2 MAJOR findings and 8 MINOR findings. The MAJOR findings need a change. The
MINOR findings do not prevent an approval. The tests are real: each break of the
production code made the correct test fail.

xamcross:
**Review (web security specialist): CHANGES REQUESTED**

First review. I used my own clone `C:/Users/xamcr/octo-wt/review-97-security`, at the head `edf8df0` of `issue-5-host-origin-checks`.

`./gradlew :monitor:backend:test` gives `BUILD SUCCESSFUL`. The JUnit XML reports show 30 tests, 0 failures, 0 errors: `HealthRouteTest` 2, `ConfigLoaderTest` 13, `BindAddressTest` 2, `RequestGuardTest` 13.

I checked the test-first claim. At the commit `5310963` the test compile fails with `Unresolved reference 'requireLoopbackBindAddress'`. The tests are red before the production code.

I then attacked the guard with my own probes:

- A set of `testApplication` probes for the Host, the Origin, the method, the content type, a static file route, and a 404.
- A raw-socket probe against a real Netty server. It sends what the test client cannot send: no Host header, two Host headers, an absolute request target, and a folded header.

I deleted both probe files after the run. The branch is unchanged.

---

## Acceptance criteria

**1. `Host: evil.example` gets 403 on each route. `localhost:<port>` and `127.0.0.1:<port>` pass. — PASS.**

Each of these 23 values gets 403: `evil.example`, `localhost.evil.example`, `127.0.0.1.evil.example`, `localhost:9999`, `localhost`, `127.0.0.1`, `127.0.0.1:7431.`, `localhost:7431.`, `[::1]:7431`, `::1`, an empty value, `a@localhost:7431`, `localhost:7431@evil.example`, `evil.example@localhost:7431`, `localhost:07431`, `localhost:7431/x`, `localhost:4200` in prod mode, `127.0.0.1:4200` in prod mode, `0x7f000001:7431`, `2130706433:7431`, `127.1:7431`, and a value folded over two lines. An absent Host header gets 403 on HTTP/1.1 and on HTTP/1.0. Two Host headers give 400 or 403, never a pass. An absolute request target `http://localhost:<port>/api/health` with `Host: evil.example` gets 403. `localhost:7431`, `127.0.0.1:7431`, and `LOCALHOST:7431` pass; the upper-case form is correct, because a host name is case-insensitive.

The check covers a route that my probe adds, a static file route, and a 404. The set test is exact, thus no prefix trick and no suffix trick works. The Host check is the only protection on a GET that returns personal data, and it holds.

**The `[::1]` decision is correct.** My raw probe connects to `[::1]:<port>` and gets `Connection refused`. The server binds `127.0.0.1` only, thus no browser can ever send `Host: [::1]:<port>` to it. The refusal removes a name from the allow-list and costs nothing. It agrees with D12, which names two Host forms.

**2. A POST, PUT, PATCH, or DELETE without an allowed `Origin` gets 403 and changes no data. — PASS.**

Each of these gets 403: an absent Origin, `null`, `NULL`, an empty value, `https://localhost:7431`, `http://localhost:7431.evil.example`, `http://evil.example`, `http://localhost:7431/`, `http://localhost:7431/path`, `http://localhost:9999`, `http://localhost`, `http://a@localhost:7431`, and `http://localhost:4200` in prod mode. A `Referer` header alone never passes. `http://localhost:7431`, `http://127.0.0.1:7431`, and `HTTP://LOCALHOST:7431` pass; the upper-case form is correct per RFC 6454. My probe counts the handler calls, and the counter stays at 0.

**3. A POST with `Content-Type: text/plain` gets 415. — PASS.**

These get 415: `text/plain`, `text/plain;charset=UTF-8`, `application/x-www-form-urlencoded`, `multipart/form-data; boundary=x`, no header at all, an empty value, `application/jsonx`, `application/json+evil`, `application/*`, `*/*`, and `application/json, text/plain`. These pass: `application/json`, `application/json; charset=utf-8`, `APPLICATION/JSON`, and `Application/Json; Charset=UTF-8`. A simple form POST from a hostile page stops at the Origin check, one step earlier.

**4. `http://localhost:4200` passes in dev mode and gets 403 in prod mode. — PASS.** `http://127.0.0.1:4200` behaves the same way. `http://localhost:4200.evil.example` gets 403 in dev mode.

**5. No response holds `Access-Control-Allow-Origin`. — PASS.** I checked a 200, a 403, a 404, a 415, a static file response, and an OPTIONS preflight with `Access-Control-Request-Method: POST`. No response holds the header.

**6. Each response holds the CSP and `X-Content-Type-Options`. Each API response holds `Cache-Control: no-store`. — PARTIAL FAIL.**

The CSP text matches D12 character for character, and `X-Content-Type-Options: nosniff` is present. Both are on every response that I made: the 200, the 403, the 404, the 415, the 500, and the static file response.

`Cache-Control: no-store` is absent on an API response that a hostile page can force. See MAJOR 1.

**7. A non-loopback bind address stops the server with a clear error. — PASS.**

`requireLoopbackBindAddress` refuses `0.0.0.0`, `::`, `0:0:0:0:0:0:0:0`, `192.168.1.10`, `127.0.0.2`, `::1`, `[::1]`, `0`, and an empty value, each with the exact message. A fixed constant plus the guard meets step 6, because the constant is the strongest form of the rule: no config key can set another value. My raw probe confirms the result. A connect to `[::1]:<port>` gets `Connection refused`, and a connect to the LAN address fails. See MINOR 2 and MINOR 3.

**8. `./gradlew :monitor:backend:test` passes with `testApplication` cases. — PASS.** 30 tests, 0 failures, 0 errors.

**9. A comment on the issue records the test output. — PASS.**

**8 of the criteria pass. Criterion 6 fails in part.**

---

## Findings

### MAJOR 1 — a non-canonical path drops `Cache-Control: no-store` from an API response

`monitor/backend/src/main/kotlin/octometer/monitor/security/RequestGuard.kt:49`

`call.request.path()` returns the raw request target. Ktor routing decodes the target and collapses a repeated slash, but this test does not. The two forms give different answers:

```
GET //api/health      -> 200 OK, the real API body, no Cache-Control header
GET /%61pi/health     -> 200 OK, the real API body, no Cache-Control header
GET /api/health       -> 200 OK, the real API body, Cache-Control: no-store
```

A hostile page can force the first two forms with one element, for example ``. The Host header is correct in that request, thus the Host check passes and the handler runs. The attacker cannot read the body, because no response holds `Access-Control-Allow-Origin`. The body still reaches the browser disk cache, which `no-store` exists to prevent. The defect grows when a later issue adds the D13 routes, because `/api/apps` and `/api/apps/{appId}/users` return personal data.

Exact fix. Test the decoded and normalised path, and fail closed on a bad escape:

```kotlin
import io.ktor.http.URLDecodeException
import io.ktor.http.decodeURLPart

// A raw target can hide the /api prefix. Ktor routing decodes the target
// and collapses a repeated slash, thus this test must do the same.
private fun isApiPath(rawPath: String): Boolean {
    val decoded = try {
        rawPath.decodeURLPart()
    } catch (badEscape: URLDecodeException) {
        return true
    }
    val normalised = Regex("/+").replace(decoded, "/")
    return normalised == "/api" || normalised.startsWith("/api/")
}
```

Then change line 49 to `if (isApiPath(call.request.path())) {`. Add a test for `//api/health` and for `/%61pi/health`.

An alternative that also closes the hole: send `Cache-Control: no-store` on every response. The monitor is a loopback app, thus a cache of the UI file gains nothing.

### MAJOR 2 — a malformed `Content-Type` gives 500, and the body echoes the header value

`monitor/backend/src/main/kotlin/octometer/monitor/security/RequestGuard.kt:67`

`call.request.contentType()` throws `BadContentTypeFormatException` on a value that does not parse. The interceptor does not catch it. My raw probe sends `Content-Type: )(` on a POST with an allowed Host and an allowed Origin:

```
HTTP/1.1 500 Internal Server Error
Bad Content-Type format: )(
```

Two defects. First, the rule of step 4 says 415, and the answer is 500. Second, the body holds the attacker value word for word. The 403 body and the 415 body hold no echo, and this one breaks the pattern. A hostile page cannot reach this line today, because the Origin check runs first and stops it. A later issue that adds a plugin, or a local client that is not a browser, reaches it.

Exact fix:

```kotlin
import io.ktor.http.BadContentTypeFormatException
...
val requestContentType = try {
    call.request.contentType().withoutParameters()
} catch (badFormat: BadContentTypeFormatException) {
    null
}
if (requestContentType != ContentType.Application.Json) {
    call.respond(
        HttpStatusCode.UnsupportedMediaType,
        RejectionBody("Set the Content-Type header to application/json."),
    )
    finish()
    return@intercept
}
```

Add a test that sends a malformed value and asserts 415.

### MAJOR 3 — the method list is a deny-list, thus a later route with another method gets no Origin check

`monitor/backend/src/main/kotlin/octometer/monitor/security/RequestGuard.kt:26` and `:60`

`MUTATING_METHODS` names four methods. Each other method skips the Origin check and the content type check. My probe adds a route for `FOO`, `OPTIONS`, `TRACE`, and `PROPPATCH`, and each one reaches the handler:

```
FOO,       Host ok, Origin http://evil.example -> 204, the handler ran
OPTIONS,   Host ok, Origin http://evil.example -> 204, the handler ran
TRACE,     Host ok, Origin http://evil.example -> 204, the handler ran
PROPPATCH, Host ok, Origin http://evil.example -> 204, the handler ran
```

No hostile page can use this today. The Fetch standard preflights each of those methods, the preflight gets no `Access-Control-Allow-Origin`, thus the browser stops the real request. I confirmed that the preflight carries no CORS header. The defect is still real, for two reasons. The KDoc at line 32 promises "A later route gets the same protection, with no new code at that route", and that promise is false. A later issue that adds a route with another method loses the CSRF protection in silence.

Exact fix. Turn the test into an allow-list. Keep the content type rule on the four methods of D12:

```kotlin
// Only GET and HEAD are safe. Each other method needs an allowed Origin.
private val SAFE_METHODS = setOf(HttpMethod.Get, HttpMethod.Head)
```

```kotlin
if (call.request.httpMethod !in SAFE_METHODS) {
    val origin = call.request.header(HttpHeaders.Origin)?.lowercase()
    if (origin == null || origin !in allowedOrigins) {
        call.respond(HttpStatusCode.Forbidden, RejectionBody("The Origin header is missing or not allowed."))
        finish()
        return@intercept
    }
}
if (call.request.httpMethod in MUTATING_METHODS) {
    // the content type check of step 4, unchanged
}
```

Add a test with a route on a custom method. Assert 403 without an allowed Origin.

---

### MINOR 1 — the guard runs on the `Plugins` phase

`monitor/backend/src/main/kotlin/octometer/monitor/security/RequestGuard.kt:46`

A plugin that a later issue installs on the `Setup` phase or the `Monitoring` phase runs before the guard. Use `ApplicationCallPipeline.Setup`, thus the guard is always first.

### MINOR 2 — the bind-address set holds `localhost`, and the test is case-sensitive

`monitor/backend/src/main/kotlin/octometer/monitor/security/BindAddress.kt:7`

D12 names `127.0.0.1`. The set adds `localhost`, and `localhost` resolves to `::1` first on Windows. A bind on `localhost` can then listen on the IPv6 loopback only, and the Host check refuses `[::1]:<port>`, thus the UI breaks. The set also refuses `LOCALHOST`, while the Host check lowercases its input. Keep only `127.0.0.1`, or lowercase the input first.

### MINOR 3 — the bind guard is dead code today

`monitor/backend/src/main/kotlin/octometer/monitor/Application.kt:38`

`HOST` is a compile-time constant, thus `requireLoopbackBindAddress(HOST)` can never throw. The decision in the pull request text is correct, and I accept it. Add one comment line at the call site, so a reader sees why the call stays.

### MINOR 4 — a test name promises more than the test body checks

`monitor/backend/src/test/kotlin/octometer/monitor/security/RequestGuardTest.kt:173`

The name is `the ng serve origin passes in dev mode and gets 403 in prod mode`. The body checks the dev result only. The next test checks the prod result. Rename the first test to `the ng serve origin passes in dev mode`.

### MINOR 5 — the test of the headers on a rejected request checks one header

`monitor/backend/src/test/kotlin/octometer/monitor/security/RequestGuardTest.kt:236`

`a rejected request holds the security headers too` asserts `X-Content-Type-Options` only. Add the CSP assertion and the `Cache-Control` assertion. My probe shows all three are present, thus the test passes after the change.

### MINOR 6 — correct behaviour with no test to hold it

These answers are right today, and no test locks them in: `Host: LOCALHOST:7431` passes, an absent Host header gets 403, two Host headers never pass, `Origin: null` gets 403, `Origin: http://localhost:7431/` gets 403, `Content-Type: application/x-www-form-urlencoded` gets 415, a 404 keeps the security headers, and a static file route gets the Host check.

### MINOR 7 — the KDoc "Order:" sentence needs a vertical list

`monitor/backend/src/main/kotlin/octometer/monitor/security/RequestGuard.kt:36`

The sentence joins three steps with "then" and "and". ASD-STE100 rule 8 asks for a vertical list. The rest of the language in the diff is correct: short sentences, the active voice, articles, and no `-ing` form as a noun.

---

## Other checks

- **Scope.** The diff holds six files: three production files and three test files. It adds no route and no config key. The change does what the issue asks.
- **Secrets.** The diff holds no credential, no database file, and no build output.
- **The order of the checks.** A rejected request runs no handler. My probe counts the handler calls, and the counter stays at 0 on a 403 and on a 415.
- **CORS.** The module installs no CORS plugin. I grep the module and find no `install(CORS)` call.
- **The 403 body.** It holds no echo of the hostile header value. I sent a Host value with an HTML payload, and the body stayed the fixed text.

---

## Summary

The Host check and the Origin check are exact and strict. I found no input that lets a hostile page read data or change data. The three MAJOR findings are a cache header that a non-canonical path drops, a 500 with an echo on a malformed content type, and a deny-list of methods that a later route will fall through. Fix the three, and I approve.

xamcross:
Correction round 1: each finding, with its fix, its file, and its line
(line numbers are in the branch head after this round, commits `bca3ca0`,
`030aaaa`, `b575342`, merged with `origin/main` at `84319f1`).

## MAJOR findings

**MAJOR 1 (security review) — Cache-Control: no-store dropped on a
non-canonical path.** Fixed. `RequestGuard.kt:119` adds `isApiPath`: it
decodes the path, collapses a repeated slash, and matches `/api` or a
`/api/` prefix; a bad escape fails closed (returns `true`). Called at
`RequestGuard.kt:67`. Tests: `RequestGuardTest.kt` — "a percent-encoded api
segment still gets Cache-Control no-store", "a non-api path gets no
Cache-Control header"; `RequestGuardRawSocketTest.kt` — "a repeated slash in
the path still gets Cache-Control no-store", "a path with a bad percent
escape fails closed...". The last two need a real server and a raw socket:
the Ktor test client rewrites `//api/health` into a request for host `api`,
and it will not send `/%zz` unchanged (I confirmed both with a throwaway
probe before writing the fix).

**MAJOR 2 (both reviews) — a malformed Content-Type gave 500 with an echo.**
Fixed. `RequestGuard.kt:86-99` calls `hasBody` first, and only then tries
`call.request.contentType()` inside a try block, catching
`BadContentTypeFormatException` at `RequestGuard.kt:89` and treating the
value as not JSON. The body stays the fixed text of `RejectionBody`, at
`RequestGuard.kt:41`. Test: `RequestGuardRawSocketTest.kt` — "a malformed
Content-Type gets 415, and the body does not echo the header value" (a raw
socket again, because the Ktor test client's own `HttpPlainText` plugin
parses the header and throws before the request leaves the client, even
with a `ByteArray` body — I confirmed this too before moving to a raw
socket).

**MAJOR 3 (security review) — the method list was a deny-list.** Fixed.
`RequestGuard.kt:38` replaces `MUTATING_METHODS` with `SAFE_METHODS` (GET,
HEAD). `RequestGuard.kt:78` checks `!in SAFE_METHODS`, thus a route on any
other method gets the Origin check, OPTIONS and a custom method included.
Tests: `RequestGuardTest.kt` — "a route on a custom method without an
allowed Origin header gets 403" (a `FOO` route), "an OPTIONS request gets no
Access-Control-Allow-Origin, also with an allowed Origin".

**MAJOR, Ktor review — a DELETE without a body got 415.** Fixed by the
maintainer's decision: the content type check now runs only when
`hasBody(call.request)` is true (`RequestGuard.kt:86`, `hasBody` defined at
`RequestGuard.kt:107`, on `Content-Length` or `Transfer-Encoding`). Tests:
`RequestGuardTest.kt` — "a DELETE with an allowed Origin and without a body
passes the content type rule", "a DELETE with an allowed Origin and a
text-plain body gets 415".

**MAJOR-2, Ktor review — same defect as MAJOR 1, one fix.** Covered above.

## MINOR findings

**MINOR 1 (security) — the guard should run before any other plugin.**
Fixed. `RequestGuard.kt:64` moved from `ApplicationCallPipeline.Plugins` to
`ApplicationCallPipeline.Setup`, the first phase.

**MINOR 2 (security) — the bind-address set held `localhost` too.** Fixed.
`BindAddress.kt:9` now holds one constant, `"127.0.0.1"`. Test:
`BindAddressTest.kt` — "the name localhost is not the address of D12, thus
it gives a clear error".

**MINOR 3 (security) — the bind guard is dead code with no comment.**
Fixed. `Application.kt:43-46` explains why the call stays, at the call site.

**MINOR 4 (both reviews) — a test name promised two results.** Fixed.
Renamed to "the ng serve origin passes in dev mode"
(`RequestGuardTest.kt`); the prod-mode case keeps its own test.

**MINOR 5 (both reviews) — a weak assertion on a rejected request.** Fixed.
"a rejected request holds the CSP, X-Content-Type-Options, and
Cache-Control too" now checks all three (`RequestGuardTest.kt`).

**MINOR 6 (security) — correct behaviour with no test.** Mostly fixed.
Added: an upper-case Host, an absent Host header, an Origin with a trailing
slash, `application/x-www-form-urlencoded`, an unknown route keeping the
security headers, and two Host headers (the last one needed a raw socket:
`RequestGuardRawSocketTest.kt` — "two Host headers never give a passing
response"). Not added: a folded header line and an absolute request target;
these need a raw socket too and the round is already large. I can add them
on request.

**MINOR 7 (security) — the KDoc needed a vertical list.** Fixed.
`RequestGuard.kt:49-54`.

**MINOR-1 (Ktor) — switch to `createApplicationPlugin`.** Rejected, with
evidence; see the "plugin form" decision in the pull request text. I built
and ran a throwaway probe confirming `call.respond()` inside `onCall` does
stop the route handler on Ktor 3.6.0, so the form is not unsafe. I kept the
raw `intercept()` call instead, because it is the only way to run on the
`Setup` phase that MINOR 1 of the security review asks for; `onCall` always
attaches to the later `Plugins` phase.

**MINOR-2 (Ktor) — the bind check dropped its return value.** Fixed.
`Application.kt:47-52`: `host` now holds the checked value, and
`embeddedServer` at `Application.kt:56` uses `host`, not the constant again.

**MINOR-3 (Ktor) — the bind error logged "The config is invalid."** Fixed.
`Application.kt:50` logs "The bind address is invalid." in its own catch
block.

**MINOR-6 (Ktor) — the `"dev"` string literal.** Fixed.
`RequestGuard.kt:138` and `:151` now compare with `Mode.DEV.value`.

**MINOR-7 (Ktor) — duplicated test config helpers.** Fixed. New file
`TestConfigs.kt` holds `devConfig()`, `prodConfig()`, and `allowedHost()`;
`HealthRouteTest.kt` and `RequestGuardTest.kt` no longer hold their own
copy.

**MINOR-8 (Ktor) — a port of 80 breaks the Host check.** Fixed.
`RequestGuard.kt:134` and `:147` add the bare Host and Origin form when
`config.port == 80`. I did not add a bare form for port 443, because the
server always speaks plain HTTP (D1, D12), and a browser omits a port only
when it matches the scheme's own default; port 443 is the default for
`https`, never reached here. Tests: `RequestGuardTest.kt` — "a Host on the
bare default port 80 passes...", "an Origin on the bare default port 80
passes...".

## Design document

`docs/superpowers/specs/2026-09-21-octometer-design.md`, D12: one sentence
for the allow-list of the safe methods, one sentence for the content type
rule on a request that holds a body, per the maintainer's decision 5.

## Verification

`./gradlew :monitor:backend:test`, `JAVA_HOME` on the JDK 21 install, run
after `git merge origin/main` (issue #9's SQLite store; no shared file, no
conflict). Result: `BUILD SUCCESSFUL`, 60 tests, 0 failures, 0 errors (48 of
them from this issue).

xamcross:
**Review (second review, web security specialist): APPROVE**

New clone `C:/Users/xamcr/octo-wt/review-97-security-2`, at the head `84319f1` of `issue-5-host-origin-checks`.

`./gradlew :monitor:backend:test` with JDK 21 gives `BUILD SUCCESSFUL`. The JUnit XML reports show 60 tests, 0 failures, 0 errors: `HealthRouteTest` 2, `ConfigLoaderTest` 13, `BindAddressTest` 3, `RequestGuardRawSocketTest` 4, `RequestGuardTest` 26, `EventStoreTest` 5, `MigrationRunnerTest` 3, `PragmaTest` 3, `SqliteDatabaseTest` 1.

I attacked the new code with a raw-socket probe against a real Netty server: 44 requests over the new path rule, the new body rule, the new method allow-list, and the `Setup` phase. I deleted the probe file after the run. The branch is unchanged.

The commits keep the test-first order: `bca3ca0` adds the tests, `030aaaa` adds the fix, `b575342` updates D12. The branch diff holds nine files. It holds no secret, no database file, and no build output. The D12 text now matches the new steps 2 and 4 of the issue, word for word.

---

| # | Finding | Status |
|---|---------|--------|
| MAJOR 1 | `Cache-Control: no-store` dropped on a non-canonical path | **CORRECTED** |
| MAJOR 2 | A malformed `Content-Type` gave 500 with an echo | **CORRECTED** |
| MAJOR 3 | The method list was a deny-list | **CORRECTED** |
| MINOR 1 | The guard ran on the `Plugins` phase | **CORRECTED** |
| MINOR 2 | The bind-address set held `localhost` | **CORRECTED** |
| MINOR 3 | The bind guard was dead code with no comment | **CORRECTED** |
| MINOR 4 | A test name promised two results | **CORRECTED** |
| MINOR 5 | A weak assertion on a rejected request | **CORRECTED** |
| MINOR 6 | Correct behaviour with no test | **MOSTLY CORRECTED**, two cases declined |
| MINOR 7 | The KDoc needed a vertical list | **CORRECTED** |

---

## 1. The decoded and normalised path — CORRECTED

`isApiPath` at `RequestGuard.kt:119` agrees with the routing plugin in each of the 12 forms that I sent:

```
/api/health          -> 200, no-store       /api%2Fhealth        -> 404, no-store
//api/health         -> 200, no-store       /%2561pi/health      -> 404, no header
///api//health       -> 200, no-store       /%zz/health          -> 400, no-store
/%61pi/health        -> 200, no-store       /%61pi/%zz           -> 400, no-store
/api/../api/health   -> 404, no-store       /apidocs.html        -> 200, no header
/API/health          -> 404, no header      /%61pidocs.html      -> 200, no header
```

The rule fails closed. A bad escape returns `true`, and Ktor itself answers 400 on the same target, thus no body reaches the client. A double escape gives 404 and no header, which is right: the routing plugin decodes one time too, thus `/%2561pi/health` is not an API path for either side. `/apidocs.html` now gets no header, thus the old prefix defect is gone. `decodeURLPart` decodes `%2F` where routing keeps it literal; that difference only adds the header, thus it stays on the safe side.

## 2. The content type rule with `hasBody()` — the relaxation is safe

The rule behaves as step 4 asks:

```
POST body, no Content-Type              -> 415
POST Transfer-Encoding chunked + text/plain -> 415
POST Content-Length: 0 + text/plain     -> 204   (no body, thus no rule)
POST no body at all                     -> 204
POST body + text/plain                  -> 415
POST Content-Length 100, 2 bytes sent   -> 415   (the guard rejects before it reads the body)
DELETE without a body                   -> 204
DELETE with a text/plain body           -> 415
```

**The risk of the new rule, and the proof.** A bodyless POST is a simple request under the Fetch standard. A hostile page sends it with no preflight, thus the `Origin` check is the only protection. I sent the three forms that a hostile page can produce, and each one gets 403 before any handler:

```
POST, no body, Origin: http://evil.example                       -> 403
POST, no body, no Origin header                                  -> 403
POST, Content-Length: 0, x-www-form-urlencoded, evil Origin      -> 403
```

I accept the relaxation. A request without a body carries no attacker data, and the `Origin` check stops each cross-origin form. The content type rule now works as a second layer for a request whose `Origin` already passed, which is the correct order of the two layers.

A malformed or a duplicated `Content-Length` gives 400 from Netty, before the Ktor pipeline. See MINOR N2.

## 3. The allow-list of safe methods — CORRECTED

```
FOO,       evil Origin  -> 403     TRACE,     evil Origin -> 403
OPTIONS,   evil Origin  -> 403     PROPPATCH, no Origin   -> 403
PUT,       no Origin    -> 403     OPTIONS,   good Origin -> 204, no Access-Control-Allow-Origin
```

An OPTIONS preflight with `Access-Control-Request-Method: POST` gets 403 and no CORS header. A later route on any method other than GET and HEAD gets the check with no new code, thus the KDoc promise at `RequestGuard.kt:44` is now true.

A HEAD request and a GET request against a route that changes data still run the handler. That is correct: step 2 and D12 name GET and HEAD as the safe methods, as RFC 9110 does. It puts one rule on each later issue: a GET route and a HEAD route must never change data. A hostile page can force a cross-origin GET with the right Host header, for example with an `img` element.

## 4. A malformed `Content-Type` — CORRECTED

`POST` with `Content-Type: )(` and a body gives 415, with the fixed body `{"error":"Set the Content-Type header to application/json."}`. No 500, and no echo of the header value.

## 5. The two declined cases — I accept the decline, and I reject one half of the reason

The reason is "these need a raw socket too and the round is already large". The first half no longer holds: `RequestGuardRawSocketTest.kt` now exists, with the helper `withRawSocketServer`, thus each case costs about eight lines. The second half is fair, and the decline blocks nothing.

I confirmed both behaviours on the new head:

```
folded Host header (evil.example + a continuation line)  -> 403
absolute target http://localhost:<port>/... , Host evil  -> 403
absolute target http://evil.example/... , Host good      -> 404
```

**A folded header needs no test.** Netty decides the folding, not this code, thus a regression would come from a Netty upgrade, and six tests already hold the Host check.

**An absolute request target should get a test**, for a reason that my first review did not name. `isApiPath` returns false for `http://evil.example/api/health`, thus the response loses `Cache-Control: no-store`. Today the routing plugin answers 404 for each absolute target, thus no API body is ever served that way. One test holds that, and a later route change cannot break it in silence. This is MINOR N3.

## 6. The bind address — CORRECTED, and the reason is right

`BindAddress.kt:9` now holds one value, `127.0.0.1`. The comment names the real cause: on Windows a bind on the name `localhost` can take the IPv6 loopback address first, and the Host check refuses the form `[::1]:<port>`. The two rules would then disagree, and the UI would break. The set now matches the first sentence of D12 exactly.

`Application.kt:47-52` now keeps the return value, and `embeddedServer` binds `host`, not the constant a second time. The check and the bind can no longer diverge. My raw probe confirms the result: a connect to `[::1]:<port>` gets `Connection refused`.

## 7. The `Setup` phase and the headers — CORRECTED

Each response that reaches the Ktor pipeline holds the CSP and `X-Content-Type-Options`, and each API response holds `Cache-Control: no-store`:

```
200 OK        | csp | nosniff | no-store        403 Forbidden | csp | nosniff | no-store
404 Not Found | csp | nosniff | no-store        415           | csp | nosniff | no-store
500 (a handler that throws) | csp | nosniff | no-store
```

No response holds `Access-Control-Allow-Origin`, the OPTIONS preflight included.

---

## New MINOR findings

**N1 — the `Transfer-Encoding` half of `hasBody` has no test.** `monitor/backend/src/main/kotlin/octometer/monitor/security/RequestGuard.kt:109`. I removed `|| request.header(HttpHeaders.TransferEncoding) != null` and ran the module: all 60 tests still pass. A later change can delete that half in silence, and a chunked `text/plain` body then reaches a handler. Add one raw-socket test: a POST with `Transfer-Encoding: chunked`, `Content-Type: text/plain`, and an allowed Origin must give 415. My probe shows 415 today.

**N2 — a Netty protocol error carries no security header.** A malformed `Content-Length` (`abc`, `-1`) and two `Content-Length` headers each give a bare 400 with no CSP, no `X-Content-Type-Options`, and no `Cache-Control`. Netty answers below the Ktor pipeline, thus the guard cannot run. The body is empty, and no browser sends such a request, thus I accept it. Record it as a known limit of acceptance criterion 7.

**N3 — an absolute request target drops `Cache-Control: no-store`.** `RequestGuard.kt:67`. See section 5. Add one raw-socket test.

**N4 — a 500 body holds the exception message.** My probe adds a route that throws `RuntimeException("boom-marker")`, and the 500 body holds `boom-marker`. No route throws today, thus this is out of scope for issue #5. A later API issue should install `StatusPages` with a fixed 500 body, because a driver message can hold a host name or a path (D11).

**N5 — one comment sentence is too long.** `RequestGuard.kt:115-118`. The sentence that starts "A bad escape fails closed" runs to about 40 words and joins two clauses with "and". ASD-STE100 rule 1 asks for 25 words in a description. Split it in two. The rest of the language in the diff is correct: short sentences, the active voice, articles, and no `-ing` form as a noun.

---

## Verdict

Each MAJOR finding and each MINOR finding of my first review is corrected. I attacked the new path rule, the new body rule, and the new method allow-list, and I found no input that lets a hostile page read data or change data. The five new findings are all MINOR, thus they do not prevent the approval. I approve the branch from the web security side.

xamcross:
**Review (Ktor backend engineer, second review): APPROVE**

## How I checked

I made a new clone and checked out the head `84319f1`. I ran
`./gradlew :monitor:backend:test` with `JAVA_HOME` on the JDK 21 install.
Result: `BUILD SUCCESSFUL`, 60 tests, 0 failures. The count of this issue is
48: `RequestGuardTest` 26, `ConfigLoaderTest` 13, `RequestGuardRawSocketTest`
4, `BindAddressTest` 3, `HealthRouteTest` 2. The store of issue #9 adds 12.

I broke the new production code three times. Each break made the correct test
fail. I restored the file after each break.

| The break | The failed tests |
| --- | --- |
| `isApiPath` reads the raw path again | 3 |
| `hasBody` returns `true` always | 1 |
| Only POST, PUT, PATCH, DELETE get the Origin check | 1 |

I ran my own probes against the branch, and I deleted each probe file after the
run. The clone holds no change from me.

## My first findings

| The finding | The result |
| --- | --- |
| MAJOR-1, a DELETE without a body got 415 | **CORRECTED** |
| MAJOR-2, the API rule read the raw path | **CORRECTED** |
| MINOR-1, the plugin form of Ktor 3 | NOT CORRECTED, rejected; I accept the decision, see below |
| MINOR-2, the bind check dropped its return value | CORRECTED (`Application.kt:47-56`) |
| MINOR-3, the wrong error text | CORRECTED (`Application.kt:50`) |
| MINOR-4, a test name promised two results | CORRECTED (`RequestGuardTest.kt:294`) |
| MINOR-5, a weak assertion on a rejected request | CORRECTED (`RequestGuardTest.kt:446`) |
| MINOR-6, the `"dev"` literal | CORRECTED (`RequestGuard.kt:138`, `:151`) |
| MINOR-7, the duplicated test helpers | CORRECTED (`TestConfigs.kt`) |
| MINOR-8, a port of 80 breaks the Host check | CORRECTED (`RequestGuard.kt:134`, `:147`) |

### MAJOR-1 — CORRECTED

`RequestGuard.kt:86` calls `hasBody`, and `RequestGuard.kt:107` reads
`Content-Length` and `Transfer-Encoding`. `RequestGuardTest.kt:376` proves a
DELETE with an allowed Origin and no body reaches the handler; the test asserts
the handler flag, not the status alone. `RequestGuardTest.kt:398` proves a
DELETE with a `text/plain` body gets 415. My break of `hasBody` made the first
test fail.

**The guard reads no body.** The code touches only `header()`, `contentType()`,
`httpMethod`, and `path()`; it never calls `receive()`. My raw socket probe
confirms it: a POST that announces `Content-Length: 5000000` and sends 5 bytes
gets `HTTP/1.1 415` after 243 ms, thus the guard answers before the body
arrives.

**A malformed `Content-Length` cannot pass as "no body".** `toLongOrNull()`
gives 0 for a bad value, thus `hasBody` returns false. Netty refuses such a
request first. My probe: `Content-Length: abc` gives 400, `Content-Length: -1`
gives 400, and `Content-Length: 5` gives 415. The gap is closed.

### MAJOR-2 — CORRECTED

`RequestGuard.kt:119` decodes the path, collapses a repeated slash, and matches
`/api` or the prefix `/api/`. A bad escape returns `true`, thus the rule fails
closed. `RequestGuardTest.kt:465` covers `/%61pi/health`,
`RequestGuardTest.kt:474` covers `/apidocs.html`, and the raw socket tests cover
`//api/health` and `/%zz`. My break of `isApiPath` made those 3 tests fail.

## The `Setup` phase, and the rejected MINOR-1

I accept the decision. I reject one sentence of the reason.

The decision is correct. `ApplicationCallPipeline.Setup` is the first phase,
thus the guard runs before each plugin, before routing, before a body read, and
before a static file route. The phase keeps each header. My probes:

- a 404 holds the CSP and `X-Content-Type-Options`,
- a 500 from a route that throws holds the CSP, `X-Content-Type-Options`, and
  `Cache-Control: no-store`,
- a 500 from a `StatusPages` handler holds the same three headers,
- a rejection still gets the JSON body `{"error":"..."}` with
  `Content-Type: application/json`, thus `ContentNegotiation` transforms a
  response from the `Setup` phase.

The sentence "it is the only way to run on the `Setup` phase" is wrong.
`createApplicationPlugin` reaches that phase with the hook `on(CallSetup)`, not
with `onCall`. My probe: a plugin that answers inside `on(CallSetup)` gives 403,
and the route does not run. The claim about `onCall` alone is correct.

The finding stays MINOR, and the present form is correct: `intercept` plus
`finish()` states the end of the call in the code, and a hook has no `finish()`.
Please correct the sentence in the pull request text.

## The raw socket tests

`RequestGuardRawSocketTest.kt` is good work. It depends on no fixed port:
`ServerSocket(0)` asks the operating system for a free port. Each socket sits in
a `use` block, and each server stops in a `finally` block with a timeout of 1000
ms. Each read has a socket timeout of 5000 ms, thus a lost response fails the
test and never hangs the job.

I ran the class 5 times: 4 tests, 0 failures each time, and a class time of
2.35 s, 2.42 s, 2.46 s, 2.61 s, and 2.65 s. The cost on the Windows job is about
2.5 seconds. That is acceptable.

## The design document

The D12 change holds two sentences, one for the safe-method allow-list and one
for the body rule. `git diff` on the design file shows the D12 block only. No
other design text changed. The new text matches the code: `SAFE_METHODS` holds
GET and HEAD, and the content type check runs only when the request holds a
body.

## New findings, both MINOR

### MINOR-9 — the free port can go to another process

`RequestGuardRawSocketTest.kt:91`. The helper opens a `ServerSocket(0)`, reads
the port, closes the socket, and binds the port again with Netty. Another
process can take the port in that window. Many agents work on this machine, thus
the window is real. The 5 runs give no failure, and the risk stays small. Fix,
if you want it: catch a `BindException` from `server.start` and retry the whole
block one time with a new port.

### MINOR-10 — a semicolon joins two statements in D12

`docs/superpowers/specs/2026-09-21-octometer-design.md:252-255`. Each new
sentence joins two statements with a semicolon. ASD-STE100 asks for one
statement in one sentence. The content is correct, thus this is style only.

## Summary

The 2 MAJOR findings of my first review are corrected, and I verified each one
with a mutation and with a probe. Of the 8 MINOR findings, 7 are corrected, and
1 is rejected with a reason that I accept in its result. 2 new MINOR findings
stay open. A MINOR finding does not prevent an approval, thus I approve.

PR #98, opened 2026-09-21
feat(monitor): show the refresh bar before each table

xamcross:
## Summary

This pull request adds the component `RefreshBar`. It shows the button
"Pause refresh" and the interval text. Each of the three table views
puts the bar directly after the `<h1>`, before the table content. The
paused state now lives in one root service, so it stays the same
after a navigation between the table routes.

## Decision: the text before the health answer

The text says "The app reads the refresh interval." before `GET
/api/health` gives its first answer. A press of the button shows "The
refresh is paused." at once, even before that answer. The paused
state does not depend on the interval value.

## Acceptance criteria

- [x] On each of the three table routes, `RefreshBar` stands after the
  `<h1>` and before the table content in the DOM order, inside
  `<main>`. Evidence: `apps.spec.ts`, `users.spec.ts`, and
  `elements.spec.ts` each check the DOM order of the heading, the
  bar, and the placeholder content.
- [x] The route `/manage` shows no `RefreshBar`. Evidence: the test
  "shows no refresh bar (#92)" in `manage.spec.ts`. The view
  "Page not found" holds the same test, for the same reason.
- [x] The paused state stays after a navigation between the three
  table routes. Evidence: `RefreshPauseState` is one root service.
  The test "shares the paused state between two poll stores, so it
  stays after a navigation" in `poll-store.spec.ts`, and the test
  "reads an already-paused state from a previous view" in
  `refresh-bar.spec.ts`, each build two component instances in one
  test, without a reset between them, and check the shared state.
- [x] The interval text shows the `refreshSeconds` value of `GET
  /api/health`, and it changes when the user pauses. Evidence:
  `refresh-bar.spec.ts`, the tests "shows the interval text once the
  health answer arrives" and "shows the paused text instead of the
  interval text once the user presses the button".
- [x] A press writes one message to the status region of #19, and
  `RefreshBar` adds no live region of its own. Evidence:
  `refresh-bar.spec.ts`, the tests "writes 'Refresh paused' to the
  announcer on a press", "writes 'Refresh resumed' ... on the second
  press", and "adds no role=\"status\" region of its own".
- [x] The accessible name of the button is "Pause refresh" in the two
  states. Evidence: `refresh-bar.spec.ts`, the test "sets
  aria-pressed to true, and keeps the accessible name, on a press".
  `RefreshBar` does not change `pause-refresh-button.html`.
- [x] `npm test` passes in `monitor/frontend`. Evidence below.
- [x] A comment on issue #92 records the test output (this pull
[...]

Issue #92, opened 2026-09-21
Show the refresh bar before each table


xamcross:
**Goal.** Each table view shows the button "Pause refresh" and the refresh interval between its `<h1>` and its table.

**Evidence.** Observed on 2026-09-21: #25 builds the button and mounts it in no view, and no open issue mounts it. The second accessibility review of #90 found the gap: https://github.com/xamcross/octometer/pull/90#issuecomment-5763574578. A screen reader in browse mode does not move the DOM focus, thus the focus rule of D29 does not protect that user. The button is then the only protection (WCAG 2.2.2). D28 moves the focus to the `<h1>`, and the user reads down from there, thus the button must stand after the `<h1>` and before the table.

**Implementation steps.**
1. In `monitor/frontend`, build the component `RefreshBar` with the button "Pause refresh" of #25 and one interval text.
2. Let the text follow the state: "The table refreshes each <n> seconds." while the refresh runs, and "The refresh is paused." while the user pauses it.
3. At each press, write "Refresh paused" or "Refresh resumed" to the `role="status"` region of #19.
4. Put `RefreshBar` directly after the `<h1>` in each of the three table views of #19.

**Acceptance criteria.**
- [x] On each of the three table routes, `RefreshBar` stands after the `<h1>` and before the table content in the DOM order, inside `<main>`.
- [x] The route `/manage` shows no `RefreshBar`.
- [x] The paused state stays after a navigation between the three table routes.
- [x] The interval text shows the `refreshSeconds` value of `GET /api/health`, and it changes when the user pauses.
- [x] A press writes one message to the status region of #19, and `RefreshBar` adds no live region of its own.
- [x] The accessible name of the button is "Pause refresh" in the two states.
- [x] `npm test` passes in `monitor/frontend`.
- [x] A comment on this issue records the test output.

**Related.**
- #25 gives the button and the poll store.
- #20, #52, and #53 replace the three placeholder views, and each one keeps the bar.
- #67 does the check with a real screen reader. An axe scan does not find this type of defect.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, D28, D29, D33.

xamcross:
Body edit on 2026-09-21: #25 closed. The first line now names only the open blockers: #19.

xamcross:
Body edit on 2026-09-21: #19 closed. No open blocker remains, thus the first line and the label `blocked` are removed. The work can start.

xamcross:
[...]

PR #99, opened 2026-09-21
feat(monitor): add the main nav and the skip link to the shell

### LINKED ISSUE #93: Add the main navigation and the skip link to the shell
**Goal.** The owner reaches `/manage` and `/apps` from each page through a main navigation, and a skip link goes to the main content.

**Evidence.** Observed on 2026-09-21: no open issue adds a permanent link to `/manage`. The only link is in the empty state of #20, thus an owner with one registered app must type the address. The second accessibility review of #91 names the skip link and its target `id="main-content"`: https://github.com/xamcross/octometer/pull/91#issuecomment-5763348702

**Implementation steps.**
1. In `monitor/frontend`, add `<nav aria-label="Main">` to the shell with the links "Apps" (`/apps`) and "Manage apps" (`/manage`).
2. Set `aria-current="page"` on the active link.
3. Add the link "Skip to the main content" (`href="#main-content"`) as the first element of the page that can get the focus.
4. Make the skip link visible on focus, with a visible focus state.

**Acceptance criteria.**
- [x] Each of the five routes shows the two links, and the active link has `aria-current="page"`.
- [x] The label "Main" makes the navigation different from the navigation "Breadcrumb".
- [x] After a fresh page load, the first Tab stop is the skip link, and its target exists.
- [x] The skip link is visible on focus.
- [x] Each target is a minimum of 24 by 24 CSS px.

**Related.**
- #19 gives the shell, the breadcrumb, and the target `id="main-content"`.
- #32 gives the page behind the link "Manage apps".

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, D28, D33.

xamcross:
## Summary

The shell shows a nav labelled Main. It holds the links Apps and Manage
apps. A skip link "Skip to the main content" is the first focusable
element of the page. It stays hidden until it gets the focus. The nav and
the skip link stay outside the main landmark, next to the breadcrumb nav
and the banner of #19.

## Decision: the Apps link on a route below /apps

The Apps link opens `/apps`. A user list (`/apps/:appId/users`) and an
element list (`/apps/:appId/elements`) sit below `/apps` in the route
table. The Apps link keeps a visual `current-section` state on each of
the two views, and it now also sets `aria-current="true"` there.

WAI-ARIA 1.2, section 6.6, defines the token `page` as "the current page
... in a set of pages" (https://www.w3.org/TR/wai-aria-1.2/#aria-current).
A user list is a different page from `/apps`, so `aria-current="page"`
there would name the wrong page to a screen reader. The token `true`
names the current item of a set when no more specific token applies. One
computed signal picks the token: `page` on the exact `/apps` view, `true`
on a view below it, and no attribute on each other route. A screen-reader
user now hears the current section on the two views below `/apps`, not
only a sighted user.

## The skip link and the base href

`src/index.html` sets `<base href="/">`. A browser resolves a
fragment-only href against that base, not against the present page. The
first version of this link let the browser follow its native `href`, so
a press started a full page load of `/` and the router then redirected
to `/apps`. The address of the page changed on every route except `/apps`.

The click handler now stops the default action first, then moves the
focus to `<main id="main-content" tabindex="-1">` in script. The link
keeps its native `href`, so it stays a real link for a screen reader, but
the browser never follows it. The address of the page does not change.

## Acceptance criteria

- [x] Each of the five routes shows the two links, and the active link
  has `aria-current="page"`. Evidence: `app.spec.ts`, the describe block
  "the main nav on each route", one test per route. On the two views
  below `/apps`, the Apps link gets `aria-current="true"` and the visual
  `current-section` state, per the decision above.
- [x] The label "Main" makes the navigation different from the
  navigation "Breadcrumb". Evidence: `app.spec.ts`, the test "shows a nav
  landmark labelled 'Main', different from the nav labelled 'Breadcrumb'".
- [x] After a fresh page load, the first Tab stop is the skip link, and
  its target exists. Evidence: `app.spec.ts`, the describe block "the
  skip link", the tests "is the first focusable element after a fresh
  load" and "links to the main content, which exists".
- [x] The skip link is visible on focus. Evidence: `app.spec.ts`, the
  test "becomes visible on focus, by translating the link back onto the
  screen". A further test confirms the link is not `display:none` nor
  `visibility:hidden`. The test "stops the default action, moves the
  focus to main, and starts no router navigation" confirms the fix
  above: the browser keeps the present page, and the focus lands on
  `<main>`, not on the `<h1>`.
- [x] Each target is a minimum of 24 by 24 CSS px. Evidence: `app.spec.ts`
  test "gives the skip link a minimum target size", `main-nav.spec.ts`
  test "gives each link a minimum target size of 24 by 24 CSS px".
- [x] `npm test` passes in `monitor/frontend`. Evidence: 113 tests, 15
  files, all green. The full output is in the issue comment.
- [x] A comment on this issue records the test output. This comment
  follows this pull request.

## Notes

- The banner of #19 now sits inside a `<header>` landmark, before the
  main nav. `banner.css` is unchanged, and the reserved height of the
  banner slot stays 3rem.
- `<main id="main-content">` gets `tabindex="-1"`. The skip link needs
  this to move the focus there. The `<h1 tabindex="-1">` focus rule of
  D28 is unchanged, and its own tests still cover it. The focus rule now
  strips a query string and a fragment from the compared path, so a
  navigation that only adds or removes a fragment keeps the focus in
  place.
- This change touches only the shell files and the `main-nav` folder. It
  does not touch `src/app/poll/` nor a table view component.

## Correction round

Two reviewers requested changes on the first version. This round fixes
the 1 BLOCKER, the 4 MAJOR findings, and the MINOR findings that a small
change could reach without a new decision. The response comment on this
pull request lists each finding with its fix, its file, and its line.

Closes #93

Generated with Claude Code


xamcross:
**Review (accessibility): CHANGES REQUESTED**

First review, accessibility specialist. I made a fresh clone at `C:/Users/xamcr/octo-wt/review-99-a11y`.
I ran `gh pr checkout 99`. I ran each command myself with the portable Node v24.21.0 and npm 11.19.0.
I built the app. I served `dist/frontend/browser` on the port 4331 with a static server and an
index fallback. I drove Google Chrome 153.0.8010.52 through the DevTools Protocol. I measured each
value in that browser. I read the Chromium accessibility tree, because NVDA reads that tree. I ran
axe-core on each of the five routes. I changed no file of the branch. I stopped my server.

The shell never sets `monitorApiError` yet, so the banner shows no text in the running app. For
point 4 I added the banner paragraph to the DOM with the same view-encapsulation attribute. The CSS
and the geometry are then the real ones.

I found **1 BLOCKER**, **3 MAJOR** findings, and **6 MINOR** findings.

## Acceptance criteria

| # | Criterion | Result |
|---|---|---|
| 1 | Each of the five routes shows the two links, and the active link has `aria-current="page"`. | **PASS**, with MAJOR 2 |
| 2 | The label "Main" makes the navigation different from the navigation "Breadcrumb". | **PASS** |
| 3 | After a fresh page load, the first Tab stop is the skip link, and its target exists. | **PASS** |
| 4 | The skip link is visible on focus. | **PASS** for the visibility, but see BLOCKER 1 |
| 5 | Each target is a minimum of 24 by 24 CSS px. | **PASS** |
| 6 | `npm test` passes in `monitor/frontend`. | **PASS** |
| 7 | A comment on the issue records the test output. | **PASS** |

**1.** The two links appear on each of the five routes. `aria-current="page"` sits on the Manage
apps link on `/manage`, and on the Apps link on `/apps`. No link carries it on `/apps/7/users`, on
`/apps/7/elements?userId=42`, and on `/no-such-route`. The criterion holds. The visual state of the
Apps link on the two child routes reaches no screen reader; see MAJOR 2.

**2.** `document.querySelectorAll('nav')` gives the labels `["Main","Breadcrumb"]` on each of the
five routes. The accessibility tree holds two `navigation` nodes with the names `Main` and
`Breadcrumb`. PASS.

**3.** `document.activeElement` is `BODY` after a fresh load of each route. The first Tab press puts
the focus on `A.skip-link "Skip to the main content"`. `#main-content` exists on each route. The
limit is correct: after a router navigation the focus sits on the `<h1>` (D28), and I measured the
next Tab stop as the skip link through a wrap of the tab order, not as the first stop.

**5.** Measured boxes on `/apps/7/users`:

| Target | 1280 px | 320 px | 200 % text zoom |
|---|---|---|---|
| Apps | 59.58 x 32.00 | 59.58 x 32.00 | 119.16 x 53.00 |
| Manage apps | 108.86 x 32.00 | 108.86 x 32.00 | 217.72 x 53.00 |
| The skip link | 188.42 x 40.00 | 188.42 x 40.00 | 320.00 x 106.00 |

Each value passes WCAG 2.5.8.

**6.** `npm ci`, then `npm test`:
```
Test Files  15 passed (15)
     Tests  110 passed (110)
```
`npm run build`: "Application bundle generation complete." `npm run lint`: "All files pass
linting." `npm run format:check`: "All matched files use Prettier code style!"

## The seven points of my role

**1. The landmarks.** The DOM order is the same on each of the five routes:

```
a.skip-link > nav[Main] > div.banner-slot > nav[Breadcrumb] > main > div[role=status]
```

Two navigation landmarks with two different names are correct. A screen-reader user opens the
landmark list and sees "Main navigation", "Breadcrumb navigation", and "main". The skip link comes
first, which is correct. One point is weaker than the design sequence: the banner slot sits after
the main nav, not before it. See MINOR 3.

**2. `aria-current`.** I agree with the author about the token `page`. WAI-ARIA 1.2, section 6.6,
binds `page` to the exact current page. A user list is a different page from `/apps`, so `page`
there would name the wrong page. The author is right to remove it. The author is **not** right to
leave the attribute unset. See MAJOR 2. The visual state passes WCAG 1.4.1: the state uses
`font-weight: 700` plus `text-decoration: none`, which is not colour. Both links keep the colour
`rgb(0, 0, 238)` on white, which measures 9.40:1.

**3. The skip link.** The keyboard behaviour fails in a real browser. See BLOCKER 1. The other
parts pass: `display` is `flex`, `visibility` is `visible`, the box moves from `top: -48px` to
`top: 0` on focus, and the text `#ffffff` on `#001f3f` measures **16.56:1**. With the default
action stopped, the next Tab after the skip reaches `A "Go to the app list."` inside `<main>` on
`/no-such-route`.

**4. The banner.** The focused skip link does **not** hide the banner text. Measured at 1280 px and
at 320 px with an error text present: the skip link box is `top 0, height 40`, the banner text box
starts at `top 40`, and the vertical overlap is **0 px**. `document.elementFromPoint` over the first
word of the banner text returns `P.banner-text`. The reserved height stays: the slot measures 48 px,
and the top of `<main>` stays at 127.44 px before the focus and after it. PASS.

**5. The target size and the reflow.** See the table above. The reflow at 320 px:
`document.documentElement.clientWidth` equals `scrollWidth` (320 = 320) at 100 % and at 200 % text
zoom. The nav makes no horizontal scroll bar. PASS. One defect appears at 320 px with 200 % text
zoom; see MAJOR 3.

**6. The forced-colours mode.** I set `forced-colors: active` plus `prefers-color-scheme: dark`
through `Emulation.setEmulatedMedia`, and `matchMedia('(forced-colors: active)')` reports true. The
Apps link keeps `font-weight: 700` and `text-decoration-line: none`, so the active state stays
visible. The focus ring becomes `rgb(26, 235, 255)` on the black canvas. The skip link becomes
`rgb(255, 255, 0)` on `rgb(0, 0, 0)`. PASS.

**7. axe-core.** Tag set: wcag2a, wcag2aa, wcag21a, wcag21aa, wcag22aa, best-practice.

| Route | Violations | Incomplete |
|---|---|---|
| `/manage` | **0** | 0 |
| `/apps` | **0** | 0 |
| `/apps/7/users` | **0** | 0 |
| `/apps/7/elements?userId=42` | **0** | 0 |
| `/no-such-route` | **0** | 0 |
| `/apps/7/users` with the banner error text | **1** (`region`, moderate, `app-banner`) | 0 |

## BLOCKER 1 — the skip link opens another page

`monitor/frontend/src/app/app.html:1`

The link uses `href="#main-content"`. `monitor/frontend/src/index.html:6` holds `<base href="/" />`.
A browser resolves a fragment-only href against the document base URL, not against the current URL.
The href therefore resolves to `http://host/#main-content`. That target is a different document
from `/apps/7/users`, so the browser starts a **full document load** of `/`, which the router
redirects to `/apps`.

I measured `document.querySelector('.skip-link').href` on each of the five routes. The value is
`http://127.0.0.1:4331/#main-content` every time. I then pressed Tab and Enter:

| Route | URL after the press | `<h1>` after the press |
|---|---|---|
| `/manage` | `/apps#main-content` | Apps |
| `/apps` | `/apps#main-content` | Apps |
| `/apps/7/users` | `/apps#main-content` | Apps |
| `/apps/7/elements?userId=42` | `/apps#main-content` | Apps |
| `/no-such-route` | `/apps#main-content` | Apps |

The frame navigations for each route are `["http://127.0.0.1:4331/#main-content",
"http://127.0.0.1:4331/apps#main-content"]`. A mouse click gives the same result. The Angular
execution context dies each time, which proves the full load.

A keyboard user on the Users view presses the first Tab stop, presses Enter, and lands on the Apps
view. The skip link does not bypass the repeated block. It replaces the page. This fails WCAG 2.4.1
Bypass Blocks. It also makes the pull-request text and the commit body wrong: "A press on it moves
the focus to main, and it does not start a router navigation."

**Fix.** Stop the default action. The app needs script for every view, so the native href gives no
benefit here.

`monitor/frontend/src/app/app.html:1`:
```html
<a class="skip-link" href="#main-content" (click)="onSkipLinkActivated($event)">
  Skip to the main content
</a>
```

`monitor/frontend/src/app/app.ts:71-79`:
```ts
  /**
   * Moves the focus to the main content on a skip-link activation.
   * The method stops the default action. `<base href="/">` makes a fragment
   * href point at the root page, not at the present page.
   */
  protected onSkipLinkActivated(event: Event): void {
    event.preventDefault();
    this.hostElement.nativeElement.querySelector<HTMLElement>('#main-content')?.focus();
  }
```

Add this test to `monitor/frontend/src/app/app.spec.ts`:
```ts
    it('stops the default action, so the browser keeps the present page', async () => {
      fixture.detectChanges();
      await router.navigateByUrl('/apps/7/users');
      await fixture.whenStable();

      const compiled = fixture.nativeElement as HTMLElement;
      const skipLink = compiled.querySelector('.skip-link') as HTMLAnchorElement;
      const event = new MouseEvent('click', { bubbles: true, cancelable: true });
      skipLink.dispatchEvent(event);

      expect(event.defaultPrevented).toBe(true);
    });
```

Correct the sentence in the pull-request text and in the commit body of the fix commit. I confirmed
the fix in the browser: with the default action stopped, `/no-such-route` keeps its URL, the focus
lands on `MAIN#main-content`, and the next Tab reaches `A "Go to the app list."` inside `<main>`.

## MAJOR 2 — the current-section state reaches no screen reader

`monitor/frontend/src/app/main-nav/main-nav.html:4-12`, `monitor/frontend/src/app/main-nav/main-nav.ts:35-39`

On `/apps/7/users` and on `/apps/7/elements` the Apps link shows `font-weight: 700` and
`text-decoration-line: none`. A sighted user reads "you are in the Apps section". The accessibility
tree of the nav on `/apps/7/users` holds no state at all:

```
[navigation] "Main"
  [link] "Apps"          <- no current state
  [link] "Manage apps"
```

The same tree on `/apps` also shows no `aria-current` in the snapshot, but the attribute is present
there, so NVDA speaks "current page". On the two child routes a screen-reader user gets nothing.
The information exists in the presentation and not in the code. That fails **WCAG 1.3.1 Info and
Relationships** (Level A).

The token `true` is the correct answer. WAI-ARIA 1.2, section 6.6, defines `true` as "the current
item within a container or set of related elements", and it is the fallback when no more specific
token applies. `true` does not claim a page. The breadcrumb keeps the one `aria-current="page"` of
the view, so the two do not conflict: one `page` marker for the exact view, plus one `true` marker
for the current item of the main nav. This is the pattern of the WAI navigation tutorial.

**Fix.** Put the whole rule in one computed signal, so one source decides the class and the
attribute.

`monitor/frontend/src/app/main-nav/main-nav.ts`, in place of `appsSectionActive`:
```ts
  /** 'page' on the exact Apps view, 'true' on a view below it, null elsewhere. */
  protected readonly appsCurrent = computed<'page' | 'true' | null>(() => {
    const path = this.currentPath();
    if (path === '/apps') {
      return 'page';
    }
    return path.startsWith('/apps/') ? 'true' : null;
  });
```

`monitor/frontend/src/app/main-nav/main-nav.html:4-12`:
```html
      <a
        routerLink="/apps"
        [attr.aria-current]="appsCurrent()"
        [class.current-section]="appsCurrent() !== null"
      >
        Apps
      </a>
```

Remove `routerLinkActive`, `ariaCurrentWhenActive` and `[routerLinkActiveOptions]` from the Apps
link only. Keep them on the Manage apps link. Then change the two child-route tests in
`main-nav.spec.ts:86-100` and the two in `app.spec.ts:313-339`: assert `'true'` in place of
`toBeNull()`.

## MAJOR 3 — the hidden skip link stays on screen at 320 px with 200 % text zoom

`monitor/frontend/src/app/app.css:14`

`top: -3rem` is a fixed offset of 48 px. The height of the link is not fixed. At a viewport of
320 CSS px with `html { font-size: 32px }`, which is a text zoom of 200 percent, the text wraps to
three lines and the box grows to 320 x 106 px. The offset stays at 96 px, so **10 px of the link
stay inside the viewport**, with `z-index: 100`:

| Viewport | Text zoom | Link box | Bottom edge | On screen |
|---|---|---|---|---|
| 1280 px | 100 % | 188.42 x 40 | -8 px | no |
| 320 px | 100 % | 188.42 x 40 | -8 px | no |
| 640 px | 200 % | 376.84 x 69 | -27 px | no |
| 320 px | 200 % | 320.00 x 106 | **+10 px** | **yes** |

A dark navy bar crosses the top of the page. It overlaps the Apps link of the nav by 2 px, and
`document.elementFromPoint(20, 4)` returns `A.skip-link`, so the bar also takes the pointer over
that strip. A user loses content and a target. This fails **WCAG 1.4.4 Resize Text** at a width
that **WCAG 1.4.10 Reflow** requires.

**Fix.** Move the link by its own height, not by a fixed length. Add the explicit focus ring at the
same time (see MINOR 1).

`monitor/frontend/src/app/app.css:12-28`:
```css
.skip-link {
  position: absolute;
  top: 0;
  left: 0;
  transform: translateY(-100%);
  z-index: 100;
  display: inline-flex;
  align-items: center;
  min-width: 24px;
  min-height: 24px;
  padding: 0.5rem 1rem;
  background-color: #001f3f;
  color: #ffffff;
}

.skip-link:focus {
  transform: translateY(0);
  outline: 3px solid #ffffff;
  outline-offset: -3px;
}
```

`translateY(-100%)` moves the link up by its own height at every zoom and at every wrap. Then change
the test `app.spec.ts:167-175`: read `transform` in place of `top`.

## MAJOR 4 — the test suite cannot see BLOCKER 1

`monitor/frontend/src/app/app.spec.ts:205-219`

The test "does not start a router navigation that moves the focus to the h1" spies on
`router.navigateByUrl`. jsdom follows no `href`, and the test document holds no `<base>` element.
The spy can never fire, with the defect or without it. The test stays green while the skip link
opens another page. The test at `app.spec.ts:188-203` has the same limit: it reads the focus that
the click handler sets, and jsdom never runs the browser navigation that follows.

**Fix.** Add the `defaultPrevented` test of BLOCKER 1. That assertion fails today and passes after
the fix, so it covers the real cause inside jsdom.

## MINOR findings (6)

**MINOR 1 — the focus ring of the skip link is the browser default, and the viewport clips it.**
`monitor/frontend/src/app/app.css:26-28`. No rule gives the link a focus style. Chrome draws
`outline: auto 1px rgb(16, 16, 16)` with `outline-offset: 1px`. The focused link sits at
`left 0, top 0`, so the ring on the top edge and on the left edge falls outside the viewport. Only
the right edge and the bottom edge stay visible. WCAG 2.4.7 still passes, because the link itself
moves on screen. Step 4 of the issue asks for a visible focus state, so the branch must own that
state. The fix of MAJOR 3 adds it.

**MINOR 2 — the banner text sits outside every landmark.** `monitor/frontend/src/app/app.html:7`.
With an error text present, axe reports one `region` violation on `app-banner`. The defect comes
from #19 and not from this branch, and `monitorApiError` is always null today. The issue that fills
the signal must wrap the shell head in `<header>`, or give the banner `role="region"` with a label.

**MINOR 3 — the banner slot sits after the main nav.** `monitor/frontend/src/app/app.html:5-7`. The
error text of the monitor API is the most urgent text of the page, and it now comes second. A
screen-reader user who jumps by landmark passes over it. Move `<app-banner />` above `<app-main-nav
/>`, or wrap both in a `<header>` landmark.

**MINOR 4 — a test name is not clear.** `monitor/frontend/src/app/app.spec.ts:146`. The name "stays
out of the tab order removal: it is not display:none nor visibility:hidden" is hard to read. Use
"keeps its place in the tab order, because it is not display:none nor visibility:hidden".

**MINOR 5 — `{ exact: true }` also demands an exact query-parameter match.**
`monitor/frontend/src/app/main-nav/main-nav.html:8` and `:19`. Angular maps `exact: true` to
`{ paths: 'exact', queryParams: 'exact', fragment: 'ignored', matrixParams: 'ignored' }`. A future
`/manage?tab=apps` drops `aria-current="page"` from the Manage apps link. Use the object form with
`queryParams: 'ignored'`.

**MINOR 6 — the current state removes the underline.**
`monitor/frontend/src/app/main-nav/main-nav.css:18-21`. The current link is the only nav item with
no underline, so it can read as plain text. Keep the underline, and mark the state with a border or
with `text-decoration-thickness`.

## Other checks

- **Scope.** The diff touches 8 files, all under `monitor/frontend/src/app/`. It adds the `main-nav`
  folder, and it changes the shell files. It touches no poll file and no table view. PASS.
- **Secrets.** The diff holds no connection string, no password, no `.db` file, and no build output.
  PASS.
- **Design.** D28 stays correct: the first load keeps the focus, and a later path change moves it to
  the `<h1>`. I measured both. D33 asks for targets of 24 by 24 CSS px, and each target passes. D33
  asks for one permanent `role="status"` region, and the count stays at 1 on each route.
- **Tests.** The 110 tests read the rendered DOM and assert a value. No test is empty. Two tests
  carry the limit of MAJOR 4.
- **Language.** The commit body and the pull-request text use Simplified Technical English. One
  content defect, not a language defect: the claim about the press on the skip link. See BLOCKER 1.

xamcross:
**Review (Angular engineer, first review): CHANGES REQUESTED**

I cloned the repository myself. I checked out `issue-93-main-navigation` at commit 2a70ab2. I ran each command with the portable Node 24.21.0. I changed nothing on the branch, and I merged nothing.

## The commands

| Command | Result |
| --- | --- |
| `npm ci` | exit code 0 |
| `npm test` | `Test Files  15 passed (15)`, `Tests  110 passed (110)` |
| `npm run format:check` | `All matched files use Prettier code style!` |
| `npm run build` | `Application bundle generation complete.`, output in `dist/frontend` |

## The acceptance criteria

I checked each criterion against the code and against a test. I also broke the code for four criteria in my clone. The correct test failed each time. I restored each file after the check.

1. **Each of the five routes shows the two links, and the active link has `aria-current="page"` — PASS.**
   Evidence: `main-nav.spec.ts:68-125`, `app.spec.ts:283-350`. Break test: I deleted `ariaCurrentWhenActive="page"` from `main-nav.html`. Four tests failed with `AssertionError: expected null to be 'page'`.
   Note: on `/apps/:appId/users` and on `/apps/:appId/elements` no link carries `aria-current`. The pull request gives the WAI-ARIA reason for it. I accept the decision as an Angular reviewer. The accessibility reviewer judges the token.
2. **The label "Main" makes the navigation different from the navigation "Breadcrumb" — PASS.**
   Evidence: `main-nav.html:1`, `app.spec.ts:270`, `main-nav.spec.ts:47`. Break test: I changed the label to `Primary`. Two tests failed.
3. **After a fresh page load, the first Tab stop is the skip link, and its target exists — PASS.**
   Evidence: `app.html:1`, `app.spec.ts:126` and `:135`. Break test: I moved `<app-main-nav />` above the skip link. The test at line 126 failed.
4. **The skip link is visible on focus — PASS.**
   Evidence: `app.css:12-28`. The rule `.skip-link:focus { top: 0 }` moves the link back on screen. MINOR 1 names the weakness of the test.
5. **Each target is a minimum of 24 by 24 CSS px — PASS.**
   Evidence: `app.css:19-20`, `main-nav.css:12-13`, `app.spec.ts:178`, `main-nav.spec.ts:60`. Break test: I set `min-width: 16px` in `main-nav.css`. The test failed with `expected '16px' to be '24px'`.
6. **`npm test` passes in `monitor/frontend` — PASS.** 15 files, 110 tests, exit code 0.
7. **A comment on the issue records the test output — PASS.**

## The active state on each route

I confirmed the state on each of the five routes and on "Page not found":

| Route | `aria-current` | class `current-section` |
| --- | --- | --- |
| `/apps` | Apps: `page` | Apps |
| `/manage` | Manage apps: `page` | Manage apps |
| `/apps/7/users` | none | Apps |
| `/apps/7/elements?userId=42` | none | Apps |
| `/no-such-route` | none | none |

The query parameter of the Elements route causes no problem, because `appsSectionActive()` strips the query string. The fragment is a different matter. See MAJOR 1 and MINOR 3.

## The focus rule of D28 (issue #19)

`npm test -- --include=src/app/app.spec.ts --filter="the focus rule"` gives exit code 0. The three tests of #19 still pass. The first load keeps the focus. A path change moves the focus to the `<h1>`. A change of only a query parameter keeps the focus.

## The skip link, the fragment, and the router

The pull request states that a press starts no router navigation. I checked that claim:

- The app does not use `withInMemoryScrolling`. `app.config.ts:9` passes only `withComponentInputBinding()`.
- A press moves the address to `<path>#main-content`. The browser fires `hashchange`, not `popstate`.
- `HistoryStateManager` of `@angular/router` reacts only to an event with `type === 'popstate'` (`_router-chunk.mjs:4356`). A fragment change therefore starts **no** router navigation. The claim of the pull request is correct.
- A reload of `/apps#main-content` opens the `/apps` route. A probe test confirms it.
- The fragment stays in the address, and two places of the code do not strip it. That is MAJOR 1.

## Findings

### MAJOR 1 — the code strips the query string, but not the fragment

`monitor/frontend/src/app/main-nav/main-nav.ts:30` and `:32`, `monitor/frontend/src/app/app.ts:89`.

This pull request puts `#main-content` into the address. Two places then read a wrong path.

1. `main-nav.ts` gives `currentPath() === '/apps#main-content'` after a reload of that address, or after a visit from a shared link. `appsSectionActive()` returns false. The Apps link loses the class `current-section`, and it keeps `aria-current="page"`. The visual state and the ARIA state disagree.
   Evidence: my probe navigates to `/apps#main-content` and prints `aria-current = page  class = ""`. The route `/apps/7/users#main-content` keeps the class, because the prefix test still matches.
2. `app.ts:89` puts the fragment into `path`. After a reload of `/apps#main-content`, a press on the Apps link goes to `/apps`. The path does not change for the user, but `pathChanged` is true, and the focus jumps to the `<h1>`. D28 allows the focus move only after a change of the path.
   Evidence: my probe navigates from `/apps` to `/apps#main-content`, and `document.activeElement` becomes the `<h1>`.

Exact fix. Split on both characters in the three places:

```ts
// main-nav.ts:30
map((event) => event.urlAfterRedirects.split(/[?#]/)[0]),

// main-nav.ts:32
{ initialValue: this.router.url.split(/[?#]/)[0] },

// app.ts:89
const path = event.urlAfterRedirects.split(/[?#]/)[0];
```

Add one test in `main-nav.spec.ts` for the route `/apps#main-content`. Add one test in `app.spec.ts` for a navigation from `/apps#main-content` to `/apps`.

An option, and not a replacement for the fix above: stop the default action of the link. The address then keeps no fragment, and the history gets no extra entry.

```html
<!-- app.html:1 -->
<a class="skip-link" href="#main-content" (click)="onSkipLinkActivated($event)">
```

```ts
// app.ts:77
protected onSkipLinkActivated(event: Event): void {
  event.preventDefault();
  this.hostElement.nativeElement.querySelector<HTMLElement>('#main-content')?.focus();
}
```

### MINOR 1 — the `:focus` test cannot prove that the link comes back on screen

`monitor/frontend/src/app/app.spec.ts:167-176`. The test compares only the two `top` values. I set `.skip-link:focus { top: -10rem }`. The link stays off screen, and the test still passed (exit code 0). The test is not always green: it failed for a deleted rule (`expected undefined to be truthy`), and it failed for an equal value (`expected '-3rem' not to be '-3rem'`). The read of the stylesheet is a correct method here, because jsdom gives no computed style for `:focus`.

Fix: assert the real value.

```ts
expect(focusTop).toBe('0');
```

### MINOR 2 — the spy on `navigateByUrl` covers only one path to a navigation

`monitor/frontend/src/app/app.spec.ts:205-219`. A navigation that the browser history starts goes through `Router.scheduleNavigation`, and never through `navigateByUrl`. In jsdom the address does not change on a click at all: my probe prints `href before = http://localhost:3000/  after = http://localhost:3000/`. The test proves less than its name states. It still catches one real regression: a `routerLink` on the skip link.

Fix: watch the router events.

```ts
const started: NavigationStart[] = [];
const sub = router.events.subscribe((e) => {
  if (e instanceof NavigationStart) {
    started.push(e);
  }
});
skipLink.dispatchEvent(new MouseEvent('click', { bubbles: true, cancelable: true }));
await fixture.whenStable();
expect(started).toEqual([]);
sub.unsubscribe();
```

### MINOR 3 — `{ exact: true }` also demands an exact query string

`monitor/frontend/src/app/main-nav/main-nav.html:8` and `:19`. Angular maps `exact: true` to `{ paths: 'exact', queryParams: 'exact', fragment: 'ignored', matrixParams: 'ignored' }`. On `/apps?x=1` the Apps link loses `aria-current="page"`. My probe prints `aria-current = null  class = current-section`. No route of D28 puts a query parameter on `/apps` or on `/manage` today, so the defect is latent.

Fix: name each option.

```html
[routerLinkActiveOptions]="{
  paths: 'exact',
  queryParams: 'ignored',
  fragment: 'ignored',
  matrixParams: 'ignored'
}"
```

### MINOR 4 — one test name is not clear, and the test is a duplicate

`monitor/frontend/src/app/app.spec.ts:146`. The name "stays out of the tab order removal: ..." is hard to read. The test at line 156 covers the same property with a real focus call. Keep the test at line 156. Rename the test at line 146, or delete it.

### MINOR 5 — `routerLinkActive` with no value

`monitor/frontend/src/app/main-nav/main-nav.html:6`. An empty directive input reads as a mistake. Write `routerLinkActive=""`, and add a short comment. An option without the hand-written computed: put `routerLinkActive="current-section"` on the `<li>` element, with no `exact` option. `RouterLinkActive` finds the link in its content, and the non-exact match covers the whole section. The computed in `main-nav.ts` then goes away.

### MINOR 6 — `MainNav` has no `ChangeDetectionStrategy.OnPush`

`monitor/frontend/src/app/main-nav/main-nav.ts:18-23`. No component of this project sets it, so the new component agrees with the project. I ask for no change here. A later issue can set it for each component at one time.

## The other checks

- **The component.** `MainNav` is standalone, the same as each component of this project. `toSignal` closes its subscription through the `DestroyRef` of the component, so no subscription stays open. The template calls only `appsSectionActive()`, and `main-nav.spec.ts` reaches that state through the router. The component holds no logic that a test cannot reach.
- **The scope.** The diff holds 8 files: 4 shell files and 4 files of the new `main-nav` folder. It holds no secret, no database file, and no build output.
- **The parallel pull request #98 (issue #92).** The two branches change no common file. In my clone `git merge --no-commit --no-ff origin/issue-92-refresh-bar` gives `Automatic merge went well; stopped before committing as requested`. I ran `git merge --abort` after it. #98 changes `apps/`, `users/`, `elements/`, `manage/`, `not-found/`, `poll/`, and it adds `refresh-bar/`. #99 changes `app.*` and it adds `main-nav/`. A merge of the two gives **no conflict**.
- **The language.** Each new comment, the commit body, and the pull request text follow ASD-STE100. I found no sentence above the word limit, no passive voice, and no `-ing` form as a noun. One claim of the commit body, "it does not start a router navigation", is correct, but only because Angular ignores a `hashchange`. A short note in the comment at `app.ts:71-76` helps the next reader.

---

## The flaky test of `src/app/poll/pause-refresh-integration.spec.ts`

This is **not** a finding against this pull request. This pull request does not change that file. The maintainer opens a separate bug issue.

### What I measured in my clone (20 CPU cores)

| Run | Failures |
| --- | --- |
| `npm test`, 15 times | 0 of 15 |
| `npm test -- --include=src/app/poll/pause-refresh-integration.spec.ts`, 15 times | 0 of 15 |
| `npm test` with **one** worker (`fileParallelism: false`), 6 times | **2 of 6** |

Each failure gives the reported text: `AssertionError: expected +0 to be 1` at line 58, and `AssertionError: expected +0 to be 2` at line 90.

### The cause: a leak of `document.hidden` between two test files

1. The Angular unit-test builder sets `isolate: false` for Vitest ("Defaults to false to align with the Karma/Jasmine experience", `@angular/build` `unit-test/schema.json`). Each test file of one worker therefore shares one jsdom `document`.
2. `poll-store.spec.ts:349` runs `Object.defineProperty(document, 'hidden', { value: true, configurable: true })`. The `afterEach` block at `poll-store.spec.ts:57-62` deletes the `activeElement` override, but not the `hidden` override. `document.hidden` stays `true` for each later file of the same worker.
3. `pause-refresh-integration.spec.ts` never sets `document.hidden`. `canPoll()` at `poll-store.ts:94` reads `!document.hidden`. The gate stays closed, `gateOpen$` gives `EMPTY`, the store sends no data request, and `calls` stays 0.

Evidence: I added one log line to the `beforeEach` of `pause-refresh-integration.spec.ts`. Each failed run printed:

```
PROBE document.hidden = true  own descriptor = {"value":true,"writable":false,"enumerable":false,"configurable":true}
```

That descriptor is exactly the one that `Object.defineProperty` writes. No passing run failed with `hidden = true`.

The order of the test files decides the result. The failure needs `poll-store.spec.ts` before `pause-refresh-integration.spec.ts` in one worker. A machine with many cores gives each file its own worker, so the failure stays hidden. `ubuntu-latest` gives 4 cores, so Vitest groups the 15 files into 3 workers. This pull request adds one test file and 204 lines of tests, thus the grouping changes, and the failure can start on this branch. `auditTime(0)` and the fake timers are not the cause: the file passes 15 times of 15 when it runs alone.

The run history of the repository holds no failed run on this branch. The only run, 35636356363, passed. So I could not read the CI log of the four reported failures.

### The smallest correct fix

In `monitor/frontend/src/app/poll/poll-store.spec.ts`, in the `afterEach` block, delete the `hidden` override too:

```ts
delete (document as unknown as Record<string, unknown>)['activeElement'];
delete (document as unknown as Record<string, unknown>)['hidden'];
```

I applied that one line in my clone. The suite with one worker then passed 6 times of 6. Without the line, the same command failed 2 times of 6.

A second, larger option for the bug issue: set `"isolate": true` in the `test` options of `angular.json`. That gives a fresh jsdom per file, and it removes this whole class of leak. It costs test time.

xamcross:
**Response to the two reviews.** I fixed the BLOCKER, the 4 MAJOR findings, and the MINOR findings that a small change could reach. Commit 3a87989 on branch `issue-93-main-navigation` holds each fix.

## BLOCKER

**1. The skip link opened another page.** `src/app/app.html:1`, `src/app/app.ts:76-79`. The click handler now calls `event.preventDefault()`, then moves the focus to `#main-content`. The link keeps its native `href`, so it stays a link for a screen reader, but the browser never follows it. The address of the page does not change. Test: `app.spec.ts`, "stops the default action, moves the focus to main, and starts no router navigation" (asserts `event.defaultPrevented`, the focused element, and zero started navigations). I wrote this test first and saw it fail on the old code, with `event.defaultPrevented` equal to `false`.

## MAJOR

**2. The test could not find the BLOCKER.** `src/app/app.spec.ts`. The rewritten test above asserts `defaultPrevented` and the focused element, and it watches `router.events` for a `NavigationStart`, in place of a spy on `navigateByUrl`.

**3. The current-section state reached no screen reader.** `src/app/main-nav/main-nav.ts:38-44`, `src/app/main-nav/main-nav.html:4-10`. One computed signal `appsCurrent` now returns `'page'` on the exact `/apps` view, `'true'` on a view below it, and `null` elsewhere. The template binds it with `[attr.aria-current]`. The visual `current-section` state stays. Tests: `main-nav.spec.ts`, "sets aria-current=\"true\" on the Apps link for a user list below /apps" and the matching element-list test; `app.spec.ts`, the two "shows both links on the ... route" tests, updated to expect `'true'`.

**4. The hidden skip link stayed on screen at 320 px with 200% text zoom.** `src/app/app.css:14-33`. The link now moves with `transform: translateY(-100%)` and back with `translateY(0)`, so it moves by its own height at each width and at each zoom, not by a fixed `top` offset. Test: `app.spec.ts`, "becomes visible on focus, by translating the link back onto the screen" (reads the two `transform` values from the stylesheet).

**5. The code stripped the query string but not the fragment.** `src/app/main-nav/main-nav.ts:32,34`, `src/app/app.ts:89`. All three places now use `.split(/[?#]/)[0]`. Tests: `main-nav.spec.ts`, "sets aria-current=\"page\" on the Apps link when the URL keeps a fragment"; `app.spec.ts`, "keeps the focus after a later navigation that only adds a fragment to the same path". Both tests failed on the old code before the fix.

## MINOR

- **Explicit focus ring, skip link and nav links.** `src/app/app.css:29-32` (`outline: 3px solid #ffffff`), `src/app/main-nav/main-nav.css:18-21` (`a:focus-visible`, `outline: 3px solid #001f3f`).
- **Weak `:focus` test.** `src/app/app.spec.ts`, rewritten to assert the exact `transform` value of the base rule and of the `:focus` rule (see MAJOR 4).
- **Spy on the wrong router method.** `src/app/app.spec.ts`, the BLOCKER test now watches `router.events` for `NavigationStart` (see MAJOR 2).
- **`{ exact: true }` drops `aria-current` on a query string.** `src/app/main-nav/main-nav.html:13-23`. The Manage apps link now names each option: `paths: 'exact'`, `queryParams: 'ignored'`, `fragment: 'ignored'`, `matrixParams: 'ignored'`. The Apps link no longer uses `routerLinkActive` at all; the computed signal of MAJOR 3 owns its state. Test: `main-nav.spec.ts`, "keeps aria-current=\"page\" on the Manage apps link when the URL keeps a query string".
- **Unclear, duplicate test name.** `src/app/app.spec.ts:164`. Renamed to "keeps its place in the tab order, because it is not display:none nor visibility:hidden".
- **`routerLinkActive` with no value.** `src/app/main-nav/main-nav.html:4-10`. The Apps link no longer carries `routerLinkActive`; the computed signal of MAJOR 3 sets the class directly.
- **`OnPush` for `MainNav`.** No change. `grep -r ChangeDetectionStrategy src/app` finds no match in this project, so `MainNav` stays consistent with each other component, per the Angular reviewer's own conclusion.
- **The underline on the current link.** `src/app/main-nav/main-nav.css:23-31`. The current link keeps its underline; `text-decoration-thickness: 0.2em` marks the state next to the bold weight, not through colour alone.

## Point 7 — the banner of #19

Both MINOR findings (the banner sits outside a landmark; the banner sits after the main nav) are fixed with one small change confined to `src/app/app.html`: `<app-banner />` now sits inside a `<header>` landmark, before `<app-main-nav />`, and before the change `<app-main-nav />` was first. `banner.css` is unchanged, and the reserved height of the banner slot stays 3rem.

## Verification

`npm ci`, `npm test` (113 tests, 15 files, all green), `npm run lint` ("All files pass linting."), `npm run format:check` ("All matched files use Prettier code style!"), `npm run build` ("Application bundle generation complete.") — all run in `monitor/frontend` on this branch after the fixes.

## The known flaky test

CI run 35642205639 failed once on `src/app/poll/pause-refresh-integration.spec.ts`, with the exact symptom the Angular reviewer described (`expected +0 to be 1`, `expected +0 to be 2`). This pull request does not touch that file or `poll-store.spec.ts`. I reran the failed jobs once with `gh run rerun 35642205639 --failed --repo xamcross/octometer`, and the rerun passed. All checks are green.

xamcross:
**Review (Angular engineer, second review): APPROVE**

I made a new clone at the head of `issue-93-main-navigation`, commit 3a87989. I ran each command with the portable Node 24.21.0. I changed nothing on the branch, and I merged nothing.

| Command | Result |
| --- | --- |
| `npm ci` | exit code 0 |
| `npm test` | `Test Files  15 passed (15)`, `Tests  113 passed (113)` |
| `npm run format:check` | `All matched files use Prettier code style!` |
| `npm run build` | `Application bundle generation complete.` |

## My first findings

### MAJOR 1 (the code keeps the fragment in the path) — CORRECTED

`main-nav.ts:32` and `:34`, and `app.ts:89`, now use `.split(/[?#]/)[0]`. The two new tests are there:

- `main-nav.spec.ts:126` — the Apps link holds `aria-current="page"` on `/apps#main-content`.
- `app.spec.ts:124` — the focus stays after a navigation that only adds a fragment.

Break test: I put `.split('?')[0]` back into `main-nav.ts:32`. The new test failed with `AssertionError: expected null to be 'page'`. I restored the file. The test locks the fix.

### MINOR 1 (the `:focus` test) — CORRECTED

`app.spec.ts:185-193` now asserts the exact values `translateY(-100%)` and `translateY(0)`. A rule that keeps the link off screen no longer passes. The `transform` in `app.css:18` and `:30` is also a better method than the old `top: -3rem`, because it follows the height of the link at each text zoom.

### MINOR 2 (the spy on the wrong router method) — CORRECTED

`app.spec.ts:222-247` now watches `router.events` for `NavigationStart`. It also asserts `event.defaultPrevented === true` and the focus on `<main>`.

Break test: I replaced `event.preventDefault()` in `app.ts:77` with `void event;`. The test failed with `AssertionError: expected false to be true`. I restored the file.

### MINOR 3 (`routerLinkActiveOptions`) — CORRECTED

`main-nav.html:17-22` names each option for the Manage link. The Apps link no longer uses `routerLinkActive`. It binds `[attr.aria-current]` to the computed signal. `main-nav.spec.ts:135` covers `/manage?tab=apps`.

### MINOR 4 (the test names) — CORRECTED

`app.spec.ts:164`, `:185` and `:222` read clearly now. The test at line 164 and the test at line 174 still cover the same property, but each name states its own method. I accept both.

### MINOR 5 (`routerLinkActive` with no value) — CORRECTED

The empty directive input is gone.

### MINOR 6 (no `OnPush`) — ACCEPTED

No component of this project sets `ChangeDetectionStrategy.OnPush`, so the new component agrees with the project. I accept the decision, and I ask for no change. A later issue can set it for each component at one time.

## The new work

### The skip link

`app.ts:76-79` stops the default action, and it moves the focus in code. The commit message names a defect that my first review did not find: `src/index.html:6` holds `<base href="/">`, so the browser resolves `href="#main-content"` against the root page. On `/apps` a press therefore opened `/#main-content`, which is a full page load. The author is right, and `preventDefault()` is the correct fix. The address now stays, and `NavigationStart` never arrives.

### The focus rule of D28

The four tests of the describe block "the focus rule" pass. The first load keeps the focus. A path change moves the focus to the `<h1>`. A change of only a query parameter keeps the focus. A change of only a fragment keeps the focus.

### The computed signal `appsCurrent`

`main-nav.ts:38-44` returns `'page'`, `'true'`, or `null`. My probe navigated to nine URLs and read both links:

| URL | Apps | Manage apps |
| --- | --- | --- |
| `/apps` | `page`, section | — |
| `/manage` | — | `page`, section |
| `/apps/7/users` | `true`, section | — |
| `/apps/7/elements?userId=42` | `true`, section | — |
| `/no-such-route` | — | — |
| `/apps?x=1` | `page`, section | — |
| `/apps#main-content` | `page`, section | — |
| `/manage?tab=apps` | — | `page`, section |
| `/apps/7/users#main-content` | `true`, section | — |

The table matches D28 and the two decisions of the pull request. A query string and a fragment change nothing.

## Findings of this review

### MINOR 7 — the breadcrumb test no longer reads the breadcrumb

`monitor/frontend/src/app/app.spec.ts:249-256`. The test queries `[aria-current="page"]` in the whole shell, and it takes the first match. The main nav now stands before the breadcrumb in the DOM, and on `/apps` the Apps link also holds `aria-current="page"` with the text "Apps". The test therefore reads the main nav, not the breadcrumb.

Evidence: I deleted `<app-breadcrumb-nav ... />` from `app.html`, and the test still passed (`1 passed`). The whole file still failed, because the test at line 297 lost the label "Breadcrumb". So the suite catches a lost breadcrumb, but this one test does not. A wrong `breadcrumbItems()` wiring stays hidden. `breadcrumb.spec.ts` and `breadcrumb-nav.spec.ts` cover the parts, so the risk is small.

Fix: name the landmark in the query.

```ts
const current = (fixture.nativeElement as HTMLElement).querySelector(
  'nav[aria-label="Breadcrumb"] [aria-current="page"]',
);
```

### MINOR 8 — no test covers the new `<header>` landmark

`monitor/frontend/src/app/app.html:5-9`. I deleted the `<header>` and the `</header>` lines, and each of the 31 tests of `app.spec.ts` still passed. The accessibility reviewer asked for the banner landmark, so a test must hold it.

Fix: add one test to the describe block "landmarks".

```ts
it('keeps the banner and the main nav inside one header landmark', () => {
  fixture.detectChanges();

  const compiled = fixture.nativeElement as HTMLElement;
  const headers = compiled.querySelectorAll('header');
  expect(headers.length).toBe(1);
  expect(headers[0].querySelector('.banner-slot')).toBeTruthy();
  expect(headers[0].querySelector('nav[aria-label="Main"]')).toBeTruthy();
});
```

Neither finding prevents the approval.

## The other checks

- **The scope.** The diff holds the same 8 files: 4 shell files and 4 files of `main-nav`. It holds no secret, no database file, and no build output.
- **The merge with #98 (issue #92).** The two branches still change no common file. `git merge --no-commit --no-ff origin/issue-92-refresh-bar` gives `Automatic merge went well; stopped before committing as requested`. I ran `git merge --abort` after it. A merge gives no conflict.
- **The `<header>` landmark.** `<app-root>` is a custom element, and not a sectioning element, so the `<header>` is a banner landmark. The banner comes first inside it, then the main nav. No test of #19 became false through the new structure, except the one in MINOR 7.
- **The language.** Each new comment and the commit body follow ASD-STE100. The comment at `app.ts:71-74` now states the real reason, and the reason is correct.
- **The flaky test.** Issue #100 holds it. This branch changes no file of `src/app/poll/`.

xamcross:
**Review (accessibility): APPROVE**

Second review, accessibility specialist. I made a new clone at `C:/Users/xamcr/octo-wt/review-99-a11y-2`.
I ran `gh pr checkout 99` and I got the head `3a87989`. I ran each command myself with the portable
Node v24.21.0. I built the app. I served `dist/frontend/browser` on the port 4332. I drove Google
Chrome 153.0.8010.52 through the DevTools Protocol. I measured each value in that browser. I
changed no file of the branch. I stopped my server. I deleted my clone.

Each finding of my first review is corrected. I found no new BLOCKER and no new MAJOR defect. I
approve. I list 3 MINOR findings; none of them blocks the merge.

## The result of each finding of the first review

| # | Finding | Result |
|---|---|---|
| BLOCKER 1 | The skip link opened another page. | **CORRECTED** |
| MAJOR 2 | The current-section state reached no screen reader. | **CORRECTED** |
| MAJOR 3 | The hidden skip link stayed on screen at 320 px with 200 % text zoom. | **CORRECTED** |
| MAJOR 4 | The test suite could not see BLOCKER 1. | **CORRECTED** |
| MINOR 1 | The skip link had no explicit focus ring. | **CORRECTED** |
| MINOR 2 | The banner text sat outside every landmark. | **CORRECTED** |
| MINOR 3 | The banner slot sat after the main nav. | **CORRECTED** |
| MINOR 4 | A test name was not clear. | **CORRECTED** |
| MINOR 5 | `{ exact: true }` also demanded an exact query-parameter match. | **CORRECTED** |
| MINOR 6 | The current link lost its underline. | **CORRECTED** |

### BLOCKER 1 — CORRECTED

`app.ts:76-79` now calls `event.preventDefault()` before it moves the focus. I pressed Tab and
Enter, and I clicked with the mouse, on each of the five routes. Ten cases, each one measured:

| Route | Document loads | URL after the press | Focus after the press |
|---|---|---|---|
| `/manage` | **0** | `/manage` (unchanged) | `MAIN#main-content` |
| `/apps` | **0** | `/apps` (unchanged) | `MAIN#main-content` |
| `/apps/7/users` | **0** | `/apps/7/users` (unchanged) | `MAIN#main-content` |
| `/apps/7/elements?userId=42` | **0** | unchanged, with the query string | `MAIN#main-content` |
| `/no-such-route` | **0** | `/no-such-route` (unchanged) | `MAIN#main-content` |

The mouse gives the same result on each route. I counted the frame navigations of the main frame:
the value is 0 in each of the ten cases. In my first review the value was 2, and the `<h1>` read
"Apps" every time.

The next Tab after the skip reaches `A "Go to the app list."` **inside** `<main>` on
`/no-such-route`. The other four views hold no control inside `<main>` yet, so the tab order wraps.
That is correct today.

The link keeps a real `href`: the attribute reads `#main-content`, and
`document.querySelector('.skip-link').href` reads `http://127.0.0.1:4332/#main-content`. The link
stays a link in the accessibility tree. The browser never follows it, because the handler stops the
default action.

### MAJOR 2 — CORRECTED

`main-nav.ts:38-44` returns one value from one computed signal, and `main-nav.html:6` binds it with
`[attr.aria-current]`. Measured on each of the five routes:

| Route | Apps | Manage apps | `[aria-current="page"]` on the page |
|---|---|---|---|
| `/manage` | none | **page** | 2 (the nav link, the breadcrumb) |
| `/apps` | **page** | none | 2 (the nav link, the breadcrumb) |
| `/apps/7/users` | **true** | none | 1 (the breadcrumb only) |
| `/apps/7/elements?userId=42` | **true** | none | 1 (the breadcrumb only) |
| `/no-such-route` | none | none | 1 (the breadcrumb only) |

Each value matches the rule. On a view below `/apps` the token is `true`, so a screen-reader user
now hears the current section. The one `page` marker of the view stays in the breadcrumb, which is
what the author argued for. Two `page` markers on `/manage` and on `/apps` are correct: they sit in
two different sets, which the WAI navigation tutorial and the APG breadcrumb pattern both do.

**A limit of my instrument, stated plainly.** `Accessibility.getPartialAXTree` of Chrome 153
reports no `current` property, not for the nav link and not for the breadcrumb `<span
aria-current="page">` that I used as a control. The CDP property list omits it. My evidence is
therefore the DOM attribute, which is the value that the ARIA mapping reads, plus the axe rule
`aria-valid-attr-value`, which accepts both tokens and reports no violation.

### MAJOR 3 — CORRECTED

`app.css:18` now uses `transform: translateY(-100%)`, so the link moves by its own height.

| Viewport | Text zoom | Link box | Bottom edge | On screen |
|---|---|---|---|---|
| 1280 px | 100 % | 188.42 x 40 | **0** | no |
| 320 px | 100 % | 188.42 x 40 | **0** | no |
| 640 px | 200 % | 376.84 x 69 | **0** | no |
| 1280 px | 200 % | 376.84 x 69 | **0** | no |
| 320 px | 200 % | 320.00 x 106 | **0** | no |

The bottom edge is exactly 0 in each case, so no part stays on screen. In my first review the
320 px case measured +10 px. `document.elementFromPoint(20, 3)` returns `HTML`, not the link, so the
link takes no pointer. `document.elementFromPoint` over the Apps link returns the Apps link itself,
so nothing covers it. `clientWidth` equals `scrollWidth` (320 = 320, 640 = 640, 1280 = 1280) in each
case, so the nav makes no horizontal scroll bar.

### MAJOR 4 — CORRECTED

`app.spec.ts:222-245`, the test "stops the default action, moves the focus to main, and starts no
router navigation", asserts three values: `event.defaultPrevented` is `true`, `document.activeElement`
is the `#main-content` element, and the list of `NavigationStart` events is empty. The first
assertion is the one that the old code fails, and jsdom evaluates it correctly, because
`preventDefault()` sets the flag without a browser navigation. The spy on `navigateByUrl` is gone.

## Point 5 — the new `<header>` landmark

The DOM order on each of the five routes:
```
a.skip-link > header > div.banner-slot > nav[Main] > nav[Breadcrumb] > main > div[role=status]
```

The nested accessibility snapshot of Chromium, on `/apps/7/users` with the error text present:
```
link "Skip to the main content"
banner ""
  navigation "Main"
    link "Apps"
    link "Manage apps"
navigation "Breadcrumb"
  link "Apps"
main ""
status ""
```

The banner landmark comes first, and it holds the banner text and the main nav. The skip link comes
before it, which keeps it the first Tab stop. A screen-reader user opens the landmark list and reads
"banner", "Main navigation", "Breadcrumb navigation", "main". That order matches a normal page.

**The reserved height stays.** `.banner-slot` measures **48 px** with the error text and without it.
The top of `<main>` measures **133.44 px** in both states, so the text moves nothing.

**The axe rule `region` no longer fails.** Tag set wcag2a, wcag2aa, wcag21a, wcag21aa, wcag22aa,
best-practice:

| State | Violations |
|---|---|
| Each of the five routes, no error text | **0** on each |
| Each of the five routes, with the banner error text | **0** on each |

In my first review the last row gave 1 violation (`region`, `app-banner`). Ten axe runs, zero
violations, zero incomplete results.

## Point 6 — the focus rings

| Element | Ring | Against | Contrast |
|---|---|---|---|
| `.skip-link:focus` | `solid 3px #ffffff`, offset -3px | the link background `#001f3f` | **16.56:1** |
| `nav a:focus-visible` | `solid 3px #001f3f`, offset 2px | the white canvas | **16.56:1** |

Both pass WCAG 1.4.11, which asks for 3:1. The inset offset of the skip link keeps the whole ring
inside the viewport, so the corner of the screen no longer clips it.

**Forced colours.** I set `forced-colors: active` plus `prefers-color-scheme: dark` through
`Emulation.setEmulatedMedia`, and `matchMedia('(forced-colors: active)')` reports true. Both rings
become the system highlight `rgb(26, 235, 255)` on the black canvas, so both stay visible. The Apps
link keeps `font-weight: 700` and `text-decoration-thickness: 3.2px` against `auto` on the other
link, so the current state stays visible too, without colour.

## MINOR findings (3)

**MINOR 1 — the focused skip link now covers the start of the banner error text.** This is a small
regression of point 4 of my first review, and the new `<header>` order causes it. The banner is now
the first box of the header, so its text box starts at `top 8`. The focused skip link covers
`top 0` to `top 40` and `left 0` to `left 188.42`. I sampled the text line at 65 points: 9 of them
return `A.skip-link`. The icon and the words "The monitor API did" stay behind the link, and the
rest of the sentence stays readable. The text returns as soon as the focus moves, and no WCAG AA
criterion fails, so I rate this MINOR.

**Fix.** Let the focused link take space in the flow, so it pushes the header down in place of an
overlay. `monitor/frontend/src/app/app.css:29-33`:
```css
.skip-link:focus {
  position: static;
  transform: none;
  outline: 3px solid #ffffff;
  outline-offset: -3px;
}
```
Then change the test `app.spec.ts:185-194`: assert `position` of the `:focus` rule next to the two
`transform` values.

**MINOR 2 — the banner slot is the second Tab stop while the error text shows.**
`monitor/frontend/src/app/banner/banner.css:3`. `overflow-y: auto` makes the `<div>` a scroll
container, and Chrome gives a scroll container a place in the tab order. Measured at 1280 px and at
320 px with the error text: T1 is the skip link, **T2 is `DIV.banner-slot`**, T3 is Apps, T4 is
Manage apps. The div carries no role and no name, so a keyboard user meets a stop with no label.
The line comes from #19, this branch does not touch it, and `monitorApiError` is always null today.
The issue that fills the signal must give the slot `tabindex="0"` plus `role="region"` and a label,
or must remove `overflow-y: auto`.

**MINOR 3 — the nav focus ring can land on the banner red.** `monitor/frontend/src/app/main-nav/main-nav.css:18-21`.
`outline-offset: 2px` draws the ring 5 px above the link. At 320 px with a long error text the
banner text box fills the slot to its bottom edge, so the top side of the ring falls on `#7a1913`.
`#001f3f` on `#7a1913` measures **1.56:1**. The other three sides stay on the white canvas at
16.56:1, so the ring stays perceivable, and the case needs a banner state that no code sets yet. A
white inner ring removes the doubt:
```css
a:focus-visible {
  outline: 3px solid #001f3f;
  outline-offset: 2px;
  box-shadow: 0 0 0 2px #ffffff;
}
```

## Other checks

- **Tests.** `npm ci`, then `npm test`:
```
Test Files  15 passed (15)
     Tests  113 passed (113)
```
  `npm run build`: "Application bundle generation complete." `npm run lint`: "All files pass
  linting." `npm run format:check`: "All matched files use Prettier code style!" `gh pr checks 99`:
  Frontend pass, JVM modules pass, Monitor backend on Windows pass, Secret scan pass, Tracker pass.
- **Scope.** The frontend part of the diff touches `monitor/frontend/src/app/` only.
- **Secrets.** The diff holds no connection string, no password, no database file, and no build
  output.
- **Language.** The commit body of `3a87989` uses Simplified Technical English, and it states the
  real cause of the defect. The wrong sentence of the first commit is gone.

Issue #114, opened 2026-09-21
Show the view First pages


xamcross:
**Blocked by #112.** The view reads the API of #112, and the link cell joins the level 1 table of #20. The component tests can start now with a stub response.

**Goal.** The route `/apps/:appId/first-pages` shows one row for each first page with the count of sessions. Each app row of the level 1 table opens it.

**Implementation steps.**
1. In `monitor/frontend`, add the route `/apps/:appId/first-pages`. Keep `page` in the query parameters.
2. Add a link cell "First pages" to each app row of the level 1 table of #20.
3. Load `GET /api/apps/{appId}/first-pages?page=` through the poll store of D29.
4. Build the flat table of D28 with the columns first page and sessions.
5. Write the path cell as `<th scope="row"><a routerLink>` to `/apps/:appId/sessions`, with `anonymous=true` and the `firstPath` value.
6. Add the pager with "Previous", "Next", and "Page 2 of 100", each with `aria-disabled`.

**Acceptance criteria.**
- [ ] Each app row of the level 1 table holds a link cell "First pages" to `/apps/:appId/first-pages`.
- [ ] That link is a minimum of 24 by 24 CSS px, and the keyboard reaches it in the row order.
- [ ] The column header of the link cell names the column, and the link text names the target view.
- [ ] The table shows one row for each first page with its session count.
- [ ] The view shows the path with Angular text interpolation only. It holds no `[innerHTML]`, no `[href]`, no `[attr.*]`, and no `DomSanitizer.bypass*`.
- [ ] A unit test asserts that a path `/a<img src=x onerror=alert(1)>` appears as text.
- [ ] The path cell links to `/apps/:appId/sessions` with `anonymous=true` and the `firstPath` value.
- [ ] A click on a row cell opens the link of the path cell.
- [ ] The table has a `<caption>` and `<th scope>`, and the breadcrumb holds `aria-current="page"`.
- [ ] Zero rows shows "No first page yet."
- [ ] A 404 goes to the parent view with a message.
- [ ] No table element has a CSS `display` value, and each target is a minimum of 24 by 24 CSS px.
- [ ] The numbers use `Intl.NumberFormat`, right alignment, and `tabular-nums`.
- [ ] `@for` tracks the stable key of D29.
- [ ] A reload and the Back button keep `page`.
- [ ] The view keeps `RefreshBar` of #92 between the `<h1>` and the `<table>`.
- [ ] `npm test` passes in `monitor/frontend`.
[...]

Issue #18, opened 2026-09-21
Serve the level 1 totals


xamcross:
**Goal.** `GET /api/apps` returns the level 1 totals and the poll state of each app. The monitor calculates the totals with SQL at query time (D14).

**Evidence.** Measured on 2026-09-21: the indexes of section 6 reduce level 1 from 8.4 s to 0.33 s at 1 million events.

**Implementation steps.**
1. Write the three statements of section 6 on the read connection, each with `GROUP BY app_id`. The clicks statement and the unique users statement hold the literal `kind = 0`. The unique sessions statement counts each session, with or without a click, and it has no `kind` filter.
2. Merge the three results with the rows of `app`. An app without events gets zeros.
3. Return the derived status `NEVER_POLLED` for an app with `last_poll_at IS NULL` (D8).
4. Add `GET /api/apps`. Each row holds `appId`, `name`, `clicks`, `uniqueUsers`, `uniqueSessions`, `status`, `lastSuccessAt`, `lastError`, and `nextPollAt` (D13).

**Acceptance criteria.**
- [x] A test seeds events of 2 apps with `NULL` user ids and asserts the three numbers.
- [x] `EXPLAIN QUERY PLAN` shows a covering index for each statement: `event_agg` for the clicks and the unique users, and `event_session` for the unique sessions.
- [x] A row with `kind = 1` (a session start) does not change `clicks` and `uniqueUsers`, and its session counts in `uniqueSessions`.
- [x] An app without events returns zeros and `NEVER_POLLED`.
- [x] Each row of the response holds the nine fields of step 4.
- [x] Each time field is UTC ISO 8601 with milliseconds.
- [x] `./gradlew :monitor:backend:test` passes.
- [x] A comment on this issue records the three query plans.

**Related.** #109 gives the column `kind` and the partial index `event_agg`. The design brief in the first comment of #101 holds the rule M4.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, D8, D13, D14, section 6, section 10.

xamcross:
Body edit on 2026-09-21: #9 closed. No open blocker remains, thus the first line and the label `blocked` are removed. The work can start.

xamcross:
Body edit of 2026-09-21: rule M4 of the design change. A session start is not a click, thus the clicks statement and the unique users statement count only `kind = 0`. The unique sessions statement keeps each session. Issue #111 makes that change, thus a Related line names it.

[...]

Issue #100, opened 2026-09-21
Make the test pause-refresh-integration.spec.ts stable


xamcross:
**Goal.** The test `pause-refresh-integration.spec.ts` gives the same result in each run, on a machine with few cores too.

**Evidence.**
- Observed on 2026-09-21: the CI job "Frontend" failed in the attempts 1 to 4 of the run 35636356363 (branch of #93), and attempt 5 passed. The log of attempt 1 shows `src/app/poll/pause-refresh-integration.spec.ts`: "AssertionError: expected +0 to be 1" at line 58, and "expected +0 to be 2" at line 90. That pull request does not change the file.
- Observed on 2026-09-21 by the Angular reviewer of #99 (https://github.com/xamcross/octometer/pull/99#issuecomment-5765649870): `npm test` with the default worker count passed 15 of 15 times on a machine with 20 cores. With one worker (`fileParallelism: false`) it failed 2 of 6 times with the same text.
- The cause that the reviewer found: the Angular unit-test builder sets `isolate: false`, thus the test files of one worker share one jsdom `document`. `monitor/frontend/src/app/poll/poll-store.spec.ts:349` sets `document.hidden = true` with `Object.defineProperty`. Its `afterEach` (lines 57 to 62) removes only the `activeElement` override. `canPoll()` (`poll-store.ts:94`) then reads `hidden === true` in a later test file, the gate stays closed, and the store sends no request.
- The reviewer tested the fix: with the `delete` line below, the run with one worker passed 6 of 6 times.

**Implementation steps.**
1. In `monitor/frontend`, reproduce the failure: run the tests with one worker, 6 times, and record the count of failed runs.
2. Add `delete (document as unknown as Record<string, unknown>)['hidden'];` to the `afterEach` of `src/app/poll/poll-store.spec.ts`.
3. Search each other test file in `src/app` for `Object.defineProperty(document` and for a fake timer, and add the matching cleanup to its `afterEach`.
4. Run the tests with one worker 10 times, and run `npm test` with the default worker count one time.

**Acceptance criteria.**
- [x] Before the fix, the run with one worker fails a minimum of one time in 6 runs, with the text of the evidence.
- [x] After the fix, the run with one worker passes 10 of 10 times.
- [x] Each test file that changes a property of `document` or `window` removes the change in its `afterEach`.
- [x] `npm test`, `npm run lint`, and `npm run format:check` pass in `monitor/frontend`.
- [x] The CI job "Frontend" passes in its first attempt.
[...]

PR #124, opened 2026-09-21
feat(monitor-frontend): show the app table on the apps route

### LINKED ISSUE #20: Show the app table (level 1)

xamcross:
## Summary

The route `/apps` shows the real app table (D28, level 1). It reads
`GET /api/apps` through the poll store of #25, and it keeps `RefreshBar` of
#92 between the `<h1>` and the table. The row type lives in one file,
`app-row.ts`, so a change of the API needs one edit.

This is a correction round. The Angular engineer found 2 blockers, 1 major
issue, and 7 minor issues. The accessibility specialist found 1 blocker, 3
major issues, and 6 minor issues. This round fixes each blocker, each
major issue, and each minor issue that needs a code change.

## Acceptance criteria

- [x] `NEVER_POLLED` shows "–", not 0, with a hidden text for a screen reader. Test: `shows "–" for each count of a NEVER_POLLED app, not 0, with a hidden text for a screen reader`.
- [x] A failed app keeps its numbers and shows the status as an icon plus text, "Data from HH:mm:ss", and the visible `lastError`. Test: `keeps the numbers of a failed app, and shows "Data from HH:mm:ss" and the visible lastError`. The "Data time" cell of a failed app also shows the `<time>` element now. Test: `shows the data time of a failed app that holds a lastSuccessAt, not an empty cell`.
- [x] Without an app, the view shows "No app is registered." and a link to `/manage`, with a 24 by 24 CSS px target. Tests: `shows "No app is registered." and a link to /manage when the list is empty`, `gives the link of the empty state a minimum target size of 24 by 24 CSS px`.
- [x] The table has a `<caption>` and `<th scope>`. Tests: `renders a table with a caption and a scoped row header for the app name`, `gives each column header a scope of "col"`.
- [x] A click on a row cell opens the link of the name cell, and a secondary button or a modifier key leaves the click to the browser. Tests: `does not navigate on a cell click with a modifier key, so the browser can open a new tab`, `does not navigate on a cell click with a button other than the primary button`, `does not navigate on a cell click while the user has text selected inside that cell`, `navigates on a cell click even while the user has text selected in a different part of the page`.
- [x] The numbers use `Intl.NumberFormat`, right alignment, and `tabular-nums`. Tests: `formats each count with Intl.NumberFormat, right-aligned with tabular numbers`, `formats the data time in a fixed zone against a literal string, not against a mirror of the code`.
- [x] A refresh keeps the focus and the scroll position, with the stable key of D29 proven against a reordered, shortened answer. Tests: `tracks each row with the stable key appId: a second answer that drops a row and changes the order keeps the tr node of each surviving row`, `keeps a focus outside tbody across a refresh`. I set `track $index` in my clone, ran the stable-key test, saw it fail ("expected `<tr>` to be `<tr>`"), and restored `track row.appId`.
- [x] No table element has a CSS `display` value, and each target is a minimum of 24 by 24 CSS px. Tests: `adds no CSS display value on a table element`, `gives the name link a minimum target size of 24 by 24 CSS px`.
- [x] `@for` tracks the stable key of D29. Same tests as the refresh criterion above.
- [x] The view keeps `RefreshBar` of #92 between the `<h1>` and the `<table>`. Test: `puts the refresh bar directly after the heading, before any other content`.
- [x] `npm test` passes in `monitor/frontend`. See the run below.
- [x] A comment on this issue records the test output. Posted after the first round of this pull request.

## What this round fixes

**BLOCKER (Angular) — the stable-key test could not fail.** The old test
held one row, so `track $index` and `track appId` gave the same key. The
new test holds three rows. The second answer drops one row and reorders
the rest. The test compares the identity of each surviving `<tr>` node. I
proved the fix: `track $index` in my clone made the new test fail; I
restored `track row.appId`.

**BLOCKER (Angular), with a decision — a failed monitor API showed no
state (D30).** `app.ts` now reads `RefreshIntervalState.error()` into the
`monitorApiError` signal. A failed `GET /api/health` sets the banner text
"The app did not get the refresh interval." A good answer clears the
banner. The one status region still announces one message for each
transition, and a recovery clears the banner. Tests: the new `the monitor
API banner (D30), driven by fake timers` block in `app.spec.ts`.

**MAJOR (Angular) — a modified click on a data cell navigated in the same
tab.** `onRowCellClick` now reads the click event, and it returns early for
a secondary button or for a modifier key. Tests: `does not navigate on a
cell click with a modifier key...`, `does not navigate on a cell click
with a button other than the primary button`.

**BLOCKER (accessibility) — `overflow-wrap: anywhere` broke every cell.**
That rule now applies only to the name cell (`th[scope="row"]`) and to
each free-text cell (`.data-from`, `.error-text`). A number cell and the
data-time cell keep `white-space: nowrap`. The table sits in a
`.table-scroll` container with `tabindex="0"`, `role="region"`, and
`aria-labelledby` pointing at the caption, so a keyboard user can scroll it
at a small width. No table element gets a CSS `display` value.
`jsdom` runs no layout, so a test cannot see the line-break defect itself;
the new tests check the class names, the wrapper's role, its accessible
name, and its `tabindex`. **The layout measurement itself needs a real
browser; this pull request does not repeat it.**

**MAJOR (accessibility) — the "Data time" cell was empty for a failed app
with a `lastSuccessAt`.** The cell now shows the `<time>` element for every
row that holds a `lastSuccessAt`, whatever its status. A row with no
`lastSuccessAt` shows `<span class="visually-hidden">No data</span>` next
to the `aria-hidden` dash.

**MAJOR (accessibility) — a bare "–" gave no text to a screen reader.**
Each `NEVER_POLLED` count cell, and each empty "Data time" cell, now pairs
a hidden "No data" text with an `aria-hidden="true"` dash. The class
`.visually-hidden` moved from `app.css` (scoped to the shell) to the
global `src/styles.css`, so the `Apps` view can use it too. The shell still
uses the class correctly; `app.spec.ts` still passes unchanged on that
point.

**MAJOR (accessibility) — the clickable row had no hover state and no
pointer cursor.** `tbody td` now gets `cursor: pointer`. `tbody
tr:hover, tbody tr:focus-within` now gets a background colour (not colour
alone), with a `forced-colors` variant.

**MINOR fixes.** The empty-state link now meets the 24 by 24 CSS px
target. The selection guard now reads `Selection.containsNode` against the
clicked cell, so a selection made elsewhere on the page (for example the
heading) no longer blocks a click on any row. Two long comments split
into shorter sentences (`apps.ts`, `apps.spec.ts`, `apps.css`). Two new
tests cover the "zero events" state of D30 and a fixed-zone, literal-string
time format, so a test no longer only mirrors the formatting code.

**Not fixed, needs a maintainer decision.** `app-row.ts` and #18 give the
row 9 fields; D13 of the design lists a 10th field, `gaps`, for a level 1
row. This pull request does not add a field that no open issue's API
provides. The maintainer should say whether D13 or the two issues are
right.

## The second correction round (this round)

The second accessibility review confirmed each first finding as corrected,
and it found one new MAJOR and one new MINOR. This round fixes each one.

**MAJOR A (accessibility) —
`monitor/frontend/src/app/apps/apps.css:53-55`: the page got a horizontal
scroll bar of 218 px at 320 CSS px.** The cause: `.visually-hidden` uses
`position: absolute`, and `.table-scroll` had `position: static`. The
hidden "No data" spans then escaped the clip of the container. The fix
adds `position: relative` to `.table-scroll`, with a one-sentence comment
that gives the reason.

**MINOR B (accessibility) — `apps.css:23-27`: `overflow-wrap: anywhere`
broke a short app name at 320 CSS px.** The fix adds `min-width: 8rem` to
`th[scope='row']`, as the reviewer proposed.

A new test reads the `.table-scroll` rule from the loaded stylesheet, the
way the present tests of this file read a CSS rule, and it checks that the
rule holds `position: relative`. `jsdom` runs no layout, so this test
cannot see a scroll bar. **The reviewer checks the layout in a real
browser.**

I also measured the fix myself in a real browser (Chrome, through the
Claude in Chrome extension). I built the app, served the build with a
small fake API, and loaded the page inside a 320 by 700 CSS px frame. With
the branch as it stood, `document.documentElement.scrollWidth` read 562
against a `clientWidth` of 320. With `position: relative` on
`.table-scroll`, `scrollWidth` read 320, equal to `clientWidth`, with no
page scroll. I also read `th[scope='row']` for the app name "traficio": it
rendered on one line, at a computed `min-width` of 128px. I stopped the
server and deleted the temporary script after the check.

## Verification (monitor/frontend, Node v24.21.0, npm 11.19.0)

`npm ci`, `npm test`, `npm run lint`, `npm run format:check`, and
`npm run build` each ran in `monitor/frontend`.

```
Test Files  19 passed (19)
     Tests  177 passed (177)
```

`npm run lint`: "All files pass linting."
`npm run format:check`: "All matched files use Prettier code style!"
`npm run build`: "Application bundle generation complete." Output in `dist/frontend`.

CI checks on this pull request: Frontend, JVM modules, Monitor backend on
Windows, Secret scan, and Tracker all pass.

## Notes

- The API `GET /api/apps` does not exist yet (#18). The row shape and the
  test fixtures follow D13 exactly (for the 9 fields both share), so the
  view needs no change once #18 ships an answer of the same shape.
- **Correction:** the first round of this pull request said the banner for
  a failed monitor API stayed owned by the shell, as the design and the
  notes of this issue stated. That was wrong: the closing comment of #25
  gave that state to this issue. This round adds it.
- Issue #114 owns the link cell "First pages"; this pull request does not
  add it.
- This round did not touch `src/app/poll/`, `RefreshBar`, or any file
  outside `app.ts`, `app.css`, `app.spec.ts`, `banner/banner.ts`,
  `src/styles.css`, and the `apps/` folder.

Closes #20

Generated with Claude Code


xamcross:

First review. I made my own clone, I checked out `issue-20-app-table` at `6417a83`, and I ran each command with Node v24.21.0.

## Commands that I ran in `monitor/frontend`

- `npm ci` — exit 0.
- `npm test` — "Test Files 19 passed (19)", "Tests 161 passed (161)". I ran it three times with the same result.
- `npm run lint` — "All files pass linting."
- `npm run format:check` — "All matched files use Prettier code style!"
- `npm run build` — "Application bundle generation complete." Output in `dist/frontend`.

I also wrote 17 probe tests of my own against this branch. I name them P1 to P16b below.

## Acceptance criteria

| Criterion | Result | Evidence |
| --- | --- | --- |
| `NEVER_POLLED` shows "–", not 0 | PASS | `apps.ts:106`. I changed `formatCount` to format the number always, and the test failed: "expected '0' to be '–'". |
| A failed app keeps the numbers, an icon plus text, "Data from HH:mm:ss", the visible `lastError` | PASS | `apps.html:38-51`, test at `apps.spec.ts:171`. |
| No app shows "No app is registered." and a link to `/manage` | PASS | `apps.html:5-9`, test at `apps.spec.ts:91`. |
| The table has a `<caption>` and `<th scope>` | PASS | I replaced the `<th scope="row">` with a `<td>`, and 6 tests failed. |
| A click on a row cell opens the link of the name cell | PARTIAL | A plain click works. A click with Ctrl, with Meta, or with Shift is wrong. See MAJOR 1. |
| The numbers use `Intl.NumberFormat`, right alignment, `tabular-nums` | PASS | `apps.ts:32`, `apps.css:19-22`, test at `apps.spec.ts:132`. |
| A refresh keeps the focus and the scroll position | PASS in the code, FALSE in the test | The behaviour is correct (my P6). The test that proves it is false. See BLOCKER 1. |
| No CSS `display` on a table element, targets of 24 by 24 CSS px | PASS | `apps.css:1-4`, tests at `apps.spec.ts:246` and `apps.spec.ts:264`. |
| `@for` tracks the stable key of D29 | PASS in the code, FALSE in the test | `apps.html:26` is correct. See BLOCKER 1. |
| `RefreshBar` stands between the `<h1>` and the `<table>` | PASS | I moved the bar after the table, and the test failed: "expected 2 to be 1". |
| `npm test` passes | PASS | 19 files, 161 tests, three runs. |
| A comment records the test output | PASS | The comment of 2026-09-21 on #20. |

Step 5 of the issue says "Add the states of D30". One state of D30 is absent. See BLOCKER 2.

## What I measured

- **P1. The request count in 60 seconds of fake time**, with `refreshSeconds: 10`: **1** health request, and **7** data requests, at 0 s, 10 s, 20 s, 30 s, 40 s, 50 s, and 60 s. That is one request for each interval. Correct.
- **P2, P3, P4.** A pause, a hidden tab, and a focus inside `<tbody>` each stop the data requests for 30 seconds. Correct.
- **P5.** 20 navigations to the view and back leave 0 open requests and 0 timers (`vi.getTimerCount()`). The store ends on destroy. Correct.
- **P6.** A row that leaves the answer disappears. A changed order gives the correct order, and each `<tr>` node keeps its identity. Correct.
- **P7.** A `lastError` of `<b>bold</b>` shows as plain text. No `` and no `<b>` reach the DOM. The diff holds no `innerHTML`, no `bypassSecurityTrust`, and no `[href]`. Correct.
- **P9.** A click on the name link itself calls the cell handler zero times. No double navigation. Correct.
- **P10.** Only the `<a>` is a Tab stop. The `<tr>` has no `tabindex` and no `role`. Correct.
- **P11.** An `OK` app with zero events shows "0", "0", "0". Correct.
- **P14, P15.** The old rows stay while a refresh request is open, and after a failed data request. Correct.
- **P16a, P16b.** A test that focuses an element leaves `document.activeElement` at `document.body` for the next test. The new test file changes no property of `document`, of `window`, or of `navigator`. It restores the fake timers and the `window.getSelection` spy. The rule of #100 holds.
- **Scope.** `git show --stat` names 5 files, all under `src/app/apps/`. The shell files, `src/app/poll/`, and `RefreshBar` are unchanged. The table holds no cell "First pages" (#114 owns it).

## Findings

### BLOCKER 1 — the test for the stable key is false

`monitor/frontend/src/app/apps/apps.spec.ts:290`

The test `tracks each row with the stable key appId: a refresh reuses the row node, and keeps a focus outside tbody` does not test the stable key. I changed `apps.html:26` from `track row.appId` to `track $index` and ran the file. Result: **"Tests 21 passed (21)"**. No test failed.

The cause: the test holds one row at one position. `$index` and `appId` then give the same key. The focus part of the test is also empty: the pause button stands outside the `@for` block, thus no track strategy can move it.

My probe P6 catches the defect. With `track $index` it fails with "expected `<tr>` to be `<tr>`".

**The exact fix.** Replace the test with one that holds three rows, and a second answer that drops a row and changes the order:

```ts
it('tracks each row with the stable key appId: a refresh keeps the node of each row', () => {
  startStore(10);
  flushApps([
    buildRow({ appId: 1, name: 'alpha' }),
    buildRow({ appId: 2, name: 'beta' }),
    buildRow({ appId: 3, name: 'gamma' }),
  ]);
  const nodesBefore = new Map<string, Element>();
  for (const tr of Array.from(root().querySelectorAll('tbody tr'))) {
    nodesBefore.set(tr.querySelector('a')?.textContent?.trim() ?? '', tr);
  }

  vi.advanceTimersByTime(10_000);
  flushApps([buildRow({ appId: 3, name: 'gamma' }), buildRow({ appId: 1, name: 'alpha' })]);

  const rowsAfter = Array.from(root().querySelectorAll('tbody tr'));
  expect(rowsAfter.map((tr) => tr.querySelector('a')?.textContent?.trim())).toEqual([
    'gamma',
    'alpha',
  ]);
  expect(rowsAfter[0]).toBe(nodesBefore.get('gamma'));
  expect(rowsAfter[1]).toBe(nodesBefore.get('alpha'));
});
```

Keep the present test for the focus, but give it a name that states the focus only.

### BLOCKER 2 — the state "a failed monitor API" of D30 is absent

`monitor/frontend/src/app/apps/apps.html:67` and `monitor/frontend/src/app/app.ts:44`

D30 names six states, and step 5 of the issue demands them. The state "a failed monitor API (a banner with a reserved height)" is absent.

What a user sees today, from my probe P12: `GET /api/health` fails, `RefreshIntervalState` tries again each 5 seconds, the store sends **0** data requests, `store.data()` stays undefined, and the view shows "Octometer reads the app list." The text never changes, and the page names no error. The user waits for a table that never arrives.

The closing comment of #25 states: "No data poll runs while `GET /api/health` fails; #20 must show that state." The closing comment of #92 states: "The text for a health route that fails for a long time belongs to the banner of D30, with the text 'The app did not get the refresh interval.'; `RefreshIntervalState` already gives the `error` signal."

**No open issue owns this connection.** I ran `gh issue list --repo xamcross/octometer --search "banner in:body" --state open`, and it returned nothing. I then read the body of each of the 73 open issues and searched for "banner", for "monitorApiError", and for "monitor API". Nothing matched. The comment of `app.ts:43` and the comment of `banner/banner.ts` both say "A later issue", and neither one names a number. The pull request text says the design and the notes of this issue give the banner to the shell. That is not correct: the note of #25 gives the state to #20.

**The exact fix**, in `monitor/frontend/src/app/app.ts`:

```ts
import { computed } from '@angular/core';
import { RefreshIntervalState } from './poll/refresh-interval-state';

private readonly intervalState = inject(RefreshIntervalState);

/** The error text of the monitor API (D30). The banner shows it. */
protected readonly monitorApiError = computed(() =>
  this.intervalState.error() === undefined ? null : 'The app did not get the refresh interval.',
);
```

`app.html:6` already binds `<app-banner [apiError]="monitorApiError()" />`, and `announceMonitorApiTransition()` then works with no other change. Add one test to `app.spec.ts`: a failed `GET /api/health` puts that text in the banner, and a good answer removes it.

If the maintainer prefers a separate issue, then file it, and write its number into `app.ts:43` and into `banner/banner.ts:7` in place of "A later issue".

### MAJOR 1 — a Ctrl-click on a data cell loses the present page

`monitor/frontend/src/app/apps/apps.ts:122`, `monitor/frontend/src/app/apps/apps.html:31,32,35,38,53`

`onRowCellClick` reads no event. A click with the Ctrl key, with the Meta key, or with the Shift key on a number cell or on the status cell calls `router.navigate` in the same tab. A user who wants a new tab loses the present page. My probe P8 dispatched the three clicks, and `navigate` ran 3 times.

The middle button is safe, because a browser sends `auxclick` and not `click` for it.

**The exact fix.** Give the handler the event, and leave a modified click to the browser:

```html
<td class="num" (click)="onRowCellClick(row, $event)">{{ formatCount(row, row.clicks) }}</td>
```

```ts
protected onRowCellClick(row: AppRow, event: MouseEvent): void {
  if (event.button !== 0 || event.ctrlKey || event.metaKey || event.shiftKey || event.altKey) {
    return;
  }
  if ((window.getSelection()?.toString().length ?? 0) > 0) {
    return;
  }
  void this.router.navigate(['/apps', row.appId, 'users']);
}
```

Add one test: a click with `ctrlKey: true` calls `navigate` zero times.

### MINOR findings

These do not stop an approval.

1. `apps.html:53-60` — the cell "Data time" of a failed app is empty. My probe P13 reads `""`, and the cell holds no `<time>`. The time stands in the status cell as "Data from HH:mm:ss". Show the same time here, or a dash, so each row holds a value in the column.
2. `apps.spec.ts:137,184,212,308` — each of these tests builds the expected text with the same call as the code (`new Intl.NumberFormat()`, `new Intl.DateTimeFormat(...)`, `date.getHours()`). The tests do not depend on the time zone of the machine, which the criterion asks. They also cannot catch a wrong format, because they mirror the code. Add one test with a fixed locale and a fixed zone, for example `formatToParts` with `timeZone: 'UTC'` against a literal string.
3. No test covers the state "zero events" of D30. My probe P11 shows "0", "0", "0", which is right. Add the test.
4. `app-row.ts:28-38` — the nine field names and types match step 4 of #18 exactly: `appId: number`, `name: string`, `clicks: number`, `uniqueUsers: number`, `uniqueSessions: number`, `status: AppStatus`, `lastSuccessAt: string | null`, `lastError: string | null`, `nextPollAt: string | null`. D13 of the design lists a tenth field, `gaps`, for a level 1 row. #18 and this file hold nine. The design and #18 do not agree. The maintainer decides which one is right.
5. `apps.ts:123` — the guard reads the selection of the whole document. A selection that the user made in the heading also stops a click on a cell of any row. Read `window.getSelection()?.isCollapsed` instead, or test that the selection touches the clicked cell.
6. Language (ASD-STE100):
   - `apps.ts:20` — "the text next to it carries the meaning" is a metaphor. Write "the text next to it gives the meaning".
   - `apps.ts:65-67` and `apps.ts:69-73` — two sentences hold more than 25 words. Split each one.
   - `apps.spec.ts:294-296` — the comment holds 28 words in one sentence. Split it.
   - The commit body: "No app gives a link to the management page." reads as a statement about each app. Write "With no app, the view gives a link to the management page."
7. The pull request text says: "The banner for a failed monitor API stays owned by the shell (`app.ts`), as the design and the notes of this issue state." The closing comment of #25 states the opposite. Correct the text.

## Summary

The data flow, the refresh, the request count, the destroy, the row track behaviour, the escape of each API value, the time format, the row type, and the test isolation are each correct. Two points stop the approval: the test that proves the stable key is false, and one state of D30 is absent. One major point needs a fix: a modified click on a cell must not navigate.

xamcross:
**Review (accessibility specialist): CHANGES REQUESTED**

First review. Role: accessibility specialist (WCAG 2.2 level AA, data tables, screen readers, keyboard, zoom and reflow).

## How I tested

I cloned the repository, and I checked out the head of `issue-20-app-table` (commit `6417a83`). I ran `npm ci`, `npm test`, `npm run lint`, `npm run format:check`, and `npm run build` in `monitor/frontend` with Node v24.21.0.

The API `GET /api/apps` does not exist yet. I wrote a small fake API on a free port. It served the built app, `GET /api/health` with `refreshSeconds: 5`, and `GET /api/apps` with four rows: a normal app, a `NEVER_POLLED` app, an `UNREACHABLE` app with a long `lastError`, and an app with a name of 120 characters without a space. It also logged each request. I opened the page in Chrome, and I measured the DOM, the computed styles, the line boxes, and the reflow. I stopped the server, and I deleted each temporary file and the clone.

**No screen reader runs on this machine.** `C:\Program Files\NVDA\nvda.exe`, `C:\Program Files (x86)\NVDA\nvda.exe`, and the Freedom Scientific folders are absent. Only Narrator is present, and I did not drive it. An axe scan does not replace a screen reader test. Issue #67 owns that check.

## The acceptance criteria

| Criterion | Result | Evidence |
| --- | --- | --- |
| `NEVER_POLLED` shows "–", not 0 | PASS | The three count cells of the `NEVER_POLLED` row hold one character, U+2013. See MAJOR 3 for the screen reader. |
| A failed app keeps its numbers, plus the status, "Data from HH:mm:ss", and the visible `lastError` | PASS | The failed row shows 890,123 / 3,050 / 7,788, "⚠ Unreachable", "Data from 12:05:03", and the full error text. See MAJOR 2. |
| Without an app, the view shows "No app is registered." and a link to `/manage` | PASS | The empty answer gives the text and the link `href="/manage"`. |
| The table has a `<caption>` and `<th scope>` | PASS | axe `scope-attr-valid` passes on 10 nodes. `th-has-data-cells` passes. |
| A click on a row cell opens the link of the name cell | PASS | A click on the number cell of row 3 moved the URL to `/apps/3/users`. |
| The numbers use `Intl.NumberFormat`, right alignment, and `tabular-nums` | **FAIL** | The three CSS values are present. The browser breaks the rendered number over 3 lines at 1280 CSS px, over 5 lines at 640 CSS px, and over 7 lines at 320 CSS px. See BLOCKER 1. |
| A refresh keeps the focus and the scroll position | PASS | The focus stayed on the button "Pause refresh" across a refresh. `scrollY` stayed 0. The row node stayed the same node. |
| No table element has a CSS `display` value | PASS | Chrome reports `table`, `table-caption`, `table-header-group`, `table-row-group`, `table-row`, `table-cell`. |
| Each target is a minimum of 24 by 24 CSS px | PASS | The name link measures 24 px high. See MINOR 5 for the link of the empty state. |
| `@for` tracks the stable key of D29 | PASS | The same `<a>` node survives a refresh. |
| `RefreshBar` stands between the `<h1>` and the `<table>` | PASS | The DOM order is `h1`, `app-refresh-bar`, `table`. |
| `npm test` passes | PASS | `Test Files 19 passed (19)`, `Tests 161 passed (161)`. `lint`, `format:check`, and `build` also pass. |
| A comment records the test output | PASS | The last comment of #20. |

## The points of my role

**1. The table semantics.** One flat `<table>` with a `<caption>`, six `<th scope="col">`, and a `<th scope="row">` with the link inside it. No table element carries a CSS `display` value, so each browser keeps the table role. PASS.

**2. The row click.** Each `<tr>` carries no `tabindex`, no `role`, and no ARIA attribute. Only the link reaches a keyboard user and a screen reader user. A real mouse drag over a number cell selected the text "4,210", and the page did not navigate. PASS. The hover state and the cursor fail. See MAJOR 4.

**3. The status.** Each status shows an icon plus a text. The icon sits in a `<span aria-hidden="true">`, thus a screen reader reads the text one time. The component sets no colour, thus WCAG 1.4.1 holds. I measured the contrast against the white page: each status text and each icon gives 21:1, and the name link gives 9.4:1. Both exceed 4.5:1 and 3:1. PASS.

**4. The `NEVER_POLLED` cells.** The cell holds one character, U+2013 EN DASH, and nothing more. NVDA and JAWS speak no en dash at the default punctuation level. The user hears the column header and then silence, or the word "blank". The cell needs a text. See MAJOR 3 for the exact markup.

**5. The failed app.** The numbers stay. The text "Data from HH:mm:ss" and the `lastError` sit inside the status cell of the same row, thus a screen reader ties them to the row header and to the column header. The error text wraps with `overflow-wrap: anywhere`, and no rule sets `text-overflow: ellipsis`. At 320 CSS px the page shows no horizontal scroll bar: `documentElement.scrollWidth` equals `clientWidth`. The "Data time" cell of a failed app is empty. See MAJOR 2 and BLOCKER 1.

**6. The numbers and the time.** The CSS gives `text-align: right` and `font-variant-numeric: tabular-nums`. The time uses `<time datetime="...Z">yyyy-MM-dd HH:mm:ss</time>`, and the column header names the zone, for example "Data time (GMT+3)". The CSS is correct. The layout defeats it. See BLOCKER 1.

**7. The refresh (WCAG 2.2.2).** The DOM order is `<h1>`, the refresh bar, the table, thus a browse-mode user meets the button "Pause refresh" and the interval text before the data. The status region stays empty across a refresh, thus the refresh makes no announcement. The focus and the scroll position stay. The log of my fake API proves the focus rule: with the focus on the name link, the store sent **0** requests to `/api/apps` between 20:31:51.691 and 20:32:46.323, thus 54 seconds at an interval of 5 seconds. The store sent one request at once when the focus moved to the button. PASS.

**8. The empty state and the first load.** Before the first answer the view shows "Octometer reads the app list.", thus no blank area appears. The empty answer shows "No app is registered." and the link. PASS. See MINOR 7.

**9. The target size and the forced colours.** The name link measures a minimum of 24 by 24 CSS px. **I could not switch the Windows forced-colours mode from this session, and the browser tool gives no emulation. I state that plainly.** By inspection, `apps.css` sets no colour, no background, no border, and no image. The status icon is a text glyph, thus the forced-colours mode keeps it. The focus ring stays the browser default, thus the forced-colours mode keeps it too.

**10. The axe scan.** I scanned `/apps` with axe-core 4.13.0, with the tags `wcag2a`, `wcag2aa`, `wcag21a`, `wcag21aa`, `wcag22aa`, and `best-practice`.

- The first load: 0 violations.
- The empty list: 0 violations.
- The four rows (a normal app, `NEVER_POLLED`, a failed app, a long name): 0 violations, 41 passes, 1 incomplete.
- The one incomplete item is `color-contrast` on the `aria-hidden` check mark. The message is "Element content contains only non-text characters". The glyph is black on white, 21:1. It is a false positive.

An axe scan does not replace a screen reader. The scan missed each finding below. The rule `empty-table-header` passed, because it reads a header cell and not a data cell.

## The findings

### BLOCKER 1 — `monitor/frontend/src/app/apps/apps.css:11-17` — the numbers and the headers break inside a word

Mark for #52 and #53.

The rule `th, td { overflow-wrap: anywhere; }` applies to each cell. `overflow-wrap: anywhere` sets the min-content width of a cell to one character. The automatic table layout then gives almost the full width to the status column, because that column holds the long `lastError`. Each other column collapses.

My measurement in Chrome, with the four rows:

| Viewport | "Clicks" cell | "Data time" cell | "Unique users" header |
| --- | --- | --- | --- |
| 1280 CSS px | 3 lines | 8 lines | — |
| 640 CSS px (200 % zoom of 1280) | 5 lines | 14 lines | 7 lines |
| 320 CSS px (400 % zoom of 1280) | 7 lines | 18 lines | 11 lines |

At 320 CSS px each column header also breaks to one character per line: "C / li / ck / s" and "U / n / i / q / u / e u / s / e / r / s". The number 1,234,584 renders on 7 lines: "1, / 2 / 3 / 4, / 5 / 8 / 4". A user cannot read that. The right alignment and the tabular numbers of D30 give no help, because a broken number does not compare down the column. WCAG 1.4.10 permits a two-dimensional scroll for a data table, and that scroll is better than one character per line. WCAG 1.4.4 also asks for no loss of content at 200 %.

The exact fix. Limit `overflow-wrap: anywhere` to the free-text cells. Keep each number and each time on one line. Let the table scroll sideways.

```css
/* apps.css */
th,
td {
  padding: 0.5rem 0.75rem;
  text-align: left;
  vertical-align: top;
}

/* Only a free-text cell breaks inside a word (D30). */
th[scope='row'],
.data-from,
.error-text {
  overflow-wrap: anywhere;
}

/* A number and a time stay on one line. */
.num,
.data-time {
  white-space: nowrap;
}

/* The table scrolls sideways when the columns need more width. */
.table-scroll {
  overflow-x: auto;
}
```

```html
<!-- apps.html, around the <table> -->
<div class="table-scroll" tabindex="0" role="region" aria-labelledby="apps-caption">
  <table>
    <caption id="apps-caption">
      The registered apps, with the click totals of each app.
    </caption>
    ...
  </table>
</div>
```

The `tabindex="0"` makes the scroll container reachable with the keyboard (WCAG 2.1.1). The `role="region"` plus `aria-labelledby` give the container a name, thus a screen reader user knows the target of the scroll.

Please add a test for the layout, or record the measurement in the issue. `jsdom` runs no layout, thus the present tests cannot see this defect. `formats each count with Intl.NumberFormat, right-aligned with tabular numbers` passes while the browser breaks the number over three lines.

### MAJOR 2 — `monitor/frontend/src/app/apps/apps.html:53-61` — the "Data time" cell is empty for each failed app

Mark for #52 and #53.

The template shows the time only for an app that did not fail, and it shows the dash only for an app without `lastSuccessAt`. A failed app with a `lastSuccessAt` matches neither branch. The cell is then empty.

My measurement in Chrome, for the failed row: `innerHTML` is `"<!----><!---->"`, and `textContent.length` is 0. A screen reader reads "blank" in the column that the user reads for the age of the data. The column that matters most is empty for exactly the rows that matter most.

The status cell shows "Data from 12:05:03". That text carries the time only. It carries no date and no zone name. A poll that last succeeded two days ago reads as today.

The exact fix. Show the full local time in the "Data time" cell of each row that holds a `lastSuccessAt`, whatever the status.

```html
<td class="data-time" (click)="onRowCellClick(row)">
  @if (row.lastSuccessAt) {
    <time [attr.datetime]="row.lastSuccessAt">{{ formatDateTime(row.lastSuccessAt) }}</time>
  } @else {
    <span class="visually-hidden">No data</span><span aria-hidden="true">–</span>
  }
</td>
```

The text "Data from HH:mm:ss" of D30 stays in the status cell. The column then always holds the date and the time.

### MAJOR 3 — `monitor/frontend/src/app/apps/apps.html:31-37` and `apps.ts:105-107` — a bare dash gives no text to a screen reader

Mark for #52 and #53.

Each count cell of a `NEVER_POLLED` app holds one character, U+2013 EN DASH, and nothing more. NVDA and JAWS speak no en dash at the default punctuation level. The user hears the column header and then silence, or the word "blank". A sighted user reads "no data". A screen reader user reads nothing.

The exact fix. Give the cell a visually hidden text, and hide the glyph from the accessibility tree.

```html
<td class="num" (click)="onRowCellClick(row)">
  @if (row.status === 'NEVER_POLLED') {
    <span class="visually-hidden">No data</span><span aria-hidden="true">–</span>
  } @else {
    {{ formatCount(row, row.clicks) }}
  }
</td>
```

Apply the same markup to the other two count cells and to the empty "Data time" cell of MAJOR 2.

One note for the fix. The class `.visually-hidden` lives in `monitor/frontend/src/app/app.css:35-46`. That file is the component style of `App`, and Angular scopes it to that component. `Apps` cannot use the class. `monitor/frontend/src/styles.css` holds no rule today, and the build reports `styles-*.css | 0 bytes`. Please move `.visually-hidden` to `src/styles.css`. #52, #53, and each later view need the same class.

### MAJOR 4 — `monitor/frontend/src/app/apps/apps.css` — the clickable row gives no hover state and no pointer cursor

Mark for #52 and #53.

D28 makes a click on a row cell open the link of the name cell. The CSS file sets no `:hover` rule and no `cursor` value. Chrome reports `cursor: auto` on a `<td>` and on a `<tr>`. A mouse user sees no sign that the row answers a click, and the user gets no feedback under the pointer. The feature is invisible.

The exact fix.

```css
tbody td {
  cursor: pointer;
}

tbody tr:hover,
tbody tr:focus-within {
  background-color: #eef2f7;
}

@media (forced-colors: active) {
  tbody tr:hover,
  tbody tr:focus-within {
    background-color: Highlight;
    color: HighlightText;
  }
}
```

Black text on `#eef2f7` gives 18.68:1, and the link colour `#0000ee` on `#eef2f7` gives 8.36:1. Both exceed 4.5:1. `:focus-within` also shows the row of a keyboard user.

### MINOR 5 — `monitor/frontend/src/app/apps/apps.html:8` — the link of the empty state is 17.3 CSS px high

Mark for #52 and #53.

The link "Go to the app management page." measures 210.6 by 17.3 CSS px. WCAG 2.5.8 accepts that size, because the inline exception covers a link inside a sentence. The criterion of this issue says 24 by 24 CSS px for each target. I state the difference. A fix is optional.

### MINOR 6 — `monitor/frontend/src/app/apps/apps.ts:122-128` — the row click ignores each modifier key

Mark for #52 and #53.

A `Ctrl` click on the name link opens a new tab. A `Ctrl` click on a data cell of the same row navigates in the same tab. I measured it: the URL moved from `/apps` to `/apps/2/users`. The user loses the present page.

The exact fix. Take the event, and stop on a modifier key or on a secondary button.

```ts
protected onRowCellClick(row: AppRow, event: MouseEvent): void {
  if (event.button !== 0 || event.ctrlKey || event.metaKey || event.shiftKey || event.altKey) {
    return;
  }
  if ((window.getSelection()?.toString().length ?? 0) > 0) {
    return;
  }
  void this.router.navigate(['/apps', row.appId, 'users']);
}
```

The template then calls `(click)="onRowCellClick(row, $event)"`.

### MINOR 7 — `monitor/frontend/src/app/apps/apps.html:67-69` — the first load makes no announcement

The view shows "Octometer reads the app list." before the first answer, thus a user sees a text and no blank area. The change from that text to the table is silent, and the region carries no `aria-busy`. WCAG asks for no announcement here, thus this is a note and not a defect.

### MINOR 8 — `monitor/frontend/src/app/apps/apps.ts:70-74` — one comment sentence is too long

The sentence that starts "The cell carries no `role`" holds more than 25 words, and it joins three statements. ASD-STE100 rule 1 asks for a maximum of 25 words. Rule 4 asks for one idea in one sentence. Please split it.

### MINOR 9 — the commit body says the opposite of the behaviour

The body of commit `6417a83` holds the sentence "No app gives a link to the management page." The view shows the link when no app is registered. Please write "The view shows a link to the management page when no app is registered."

### MINOR 10 — no test can see a layout defect

`jsdom` runs no layout. The tests `adds no CSS display value on a table element`, `formats each count with Intl.NumberFormat, right-aligned with tabular numbers`, and `wraps a long lastError with overflow-wrap anywhere, and no ellipsis` read a computed style value only. Each one passes while the browser breaks a number over three lines. Please record a browser measurement in the issue, or add a check with a real engine.

## Summary

- 1 BLOCKER, 3 MAJOR, 6 MINOR.
- The semantics, the scope attributes, the status pattern, the contrast, the focus rule, the refresh rule, and the row-click guard are correct. That part is good work.
- The CSS rule of BLOCKER 1 is the one change that #52 and #53 must not copy.
- No secret, no database file, and no build output is in the diff. The diff holds 5 source files.

xamcross:
**Response to the two reviews.** I fixed each blocker, each major issue,
and each minor issue that needs a code change, on commit `f2ffa5b`, on the
same branch (`issue-20-app-table`).

## Angular engineer review

1. **BLOCKER — the stable-key test could not fail.** New test in
   `monitor/frontend/src/app/apps/apps.spec.ts:290` ("tracks each row with
   the stable key appId..."). It holds three rows, then a second answer
   drops one row and reorders the rest, then it compares the identity of
   each surviving `<tr>` node. I set `track $index` in
   `apps.html:27` in my clone, ran the test, saw it fail ("expected `<tr>`
   to be `<tr>`"), and restored `track row.appId`.
2. **BLOCKER — the state "a failed monitor API" of D30 was absent.** Fix in
   `monitor/frontend/src/app/app.ts:38,46-52`: `intervalState =
   inject(RefreshIntervalState)`, and `monitorApiError` is now `computed(()
   => this.intervalState.error() === undefined ? null : 'The app did not
   get the refresh interval.')`. Tests in the new `the monitor API banner
   (D30), driven by fake timers` block of `app.spec.ts`.
3. **MAJOR — a modified click on a data cell navigated in the same tab.**
   Fix in `monitor/frontend/src/app/apps/apps.ts:127-134`:
   `onRowCellClick(row, event)` returns early for `event.button !== 0` or
   for a modifier key. `apps.html` passes `$event` at each of the 5 call
   sites. Tests: `does not navigate on a cell click with a modifier key...`,
   `does not navigate on a cell click with a button other than the primary
   button`.
4. **MINOR — the "Data time" cell was empty for a failed app.** Fixed as
   part of the accessibility MAJOR 2 below (same cell).
5. **MINOR — the tests mirrored the format code.** New tests: `formats the
   data time in a fixed zone against a literal string, not against a
   mirror of the code` (`apps.spec.ts`), and `shows 0 for an OK app with
   zero events, not a dash` for the missing D30 state.
6. **MINOR — `app-row.ts` has 9 fields, D13 lists a 10th (`gaps`).** Not
   fixed. This needs your decision: is D13 or are #18/`app-row.ts` right?
7. **MINOR — the selection guard read the whole document.** Fix in
   `apps.ts:130,137-143`: `onRowCellClick` now calls
   `selectionTouchesCell(event.currentTarget)`, which uses
   `Selection.containsNode(cell, true)`. Tests: `does not navigate on a
   cell click while the user has text selected inside that cell`,
   `navigates on a cell click even while the user has text selected in a
   different part of the page`.
8. **MINOR — language.** `apps.ts:20` now reads "the text next to it gives
   the meaning". The long sentences at `apps.ts:65-67`/`70-74` and
   `apps.spec.ts` (the focus-test comment) are each split into two
   sentences now.
9. **MINOR — the pull request text claimed the wrong owner for the
   banner.** Corrected in the pull request description (see "Notes").

## Accessibility specialist review

1. **BLOCKER — `overflow-wrap: anywhere` broke every cell.** Fix in
   `monitor/frontend/src/app/apps/apps.css:11-32`: the rule now applies
   only to `th[scope='row']`, `.data-from`, and `.error-text`. `.num` and
   `.data-time` get `white-space: nowrap` (`apps.css:29-32`). The table
   sits in `<div class="table-scroll" tabindex="0" role="region"
   aria-labelledby="apps-caption">` (`apps.html:11`, `apps.css:53-55`). No
   table element gets a CSS `display` value. `jsdom` cannot see the line
   break itself; new tests check the wrapper's class, role, name, and
   `tabindex` (`apps.spec.ts`, "puts the table in a scroll container...").
   **The layout measurement itself needs a real browser; I did not repeat
   it.**
2. **MAJOR — the "Data time" cell was empty for a failed app with a
   `lastSuccessAt`.** Fix in `apps.html:68-76`: the `<time>` element shows
   for every row with a `lastSuccessAt`, whatever the status. Test: `shows
   the data time of a failed app that holds a lastSuccessAt, not an empty
   cell`.
3. **MAJOR — a bare dash gave no text to a screen reader.** Fix in
   `apps.html:33-38,40-45,47-52,73-75`: each such cell now holds
   `<span class="visually-hidden">No data</span><span
   aria-hidden="true">–</span>`. `.visually-hidden` moved from
   `app.css` to the global `monitor/frontend/src/styles.css:8-19`. Test:
   `shows "–" for each count of a NEVER_POLLED app, not 0, with a hidden
   text for a screen reader`, and the "Data time" equivalent.
4. **MAJOR — no hover state, no pointer cursor.** Fix in `apps.css:57-73`:
   `tbody td { cursor: pointer; }`, `tbody tr:hover, tbody
   tr:focus-within { background-color: #eef2f7; }`, with a
   `forced-colors` variant. Test: `gives a pointer cursor to each data cell
   of a row`, `gives a hovered row a background colour...`.
5. **MINOR — the empty-state link is 17.3 CSS px high.** Fix in
   `apps.html:8`, `apps.css:39-45`: the link now shares the 24 by 24 CSS px
   target rule with the name link (`.manage-link`). Test: `gives the link
   of the empty state a minimum target size of 24 by 24 CSS px`.
6. **MINOR — a modified click ignored.** Same fix as Angular finding 3
   above.
7. **MINOR — the first load makes no announcement.** Not a defect, per
   your own review; no change.
8. **MINOR — a comment sentence over 25 words.** Same fix as Angular
   finding 8 above (`apps.ts:70-74` doc comment split).
9. **MINOR — the commit body said the opposite of the behaviour.** I did
   not amend the old commit. The new commit body (`f2ffa5b`) states the
   correct fact, and the pull request description does too.
10. **MINOR — no test can see the layout defect.** Noted in the pull
    request text under BLOCKER 1 above; the new jsdom-reachable tests are
    listed there.

## Verification

`npm ci`, `npm test`, `npm run lint`, `npm run format:check`, and `npm run
build` each ran in `monitor/frontend`, with Node v24.21.0 and npm 11.19.0.

```
Test Files  19 passed (19)
     Tests  176 passed (176)
```

`npm run lint`: "All files pass linting."
`npm run format:check`: "All matched files use Prettier code style!"
`npm run build`: "Application bundle generation complete."

CI checks on this pull request (Frontend, JVM modules, Monitor backend on
Windows, Secret scan, Tracker) all pass on commit `f2ffa5b`.

xamcross:

I made a new clone, I checked out `issue-20-app-table` at `f2ffa5b`, and I ran each command with Node v24.21.0. I checked each finding against the code, and not against the response comment.

## Commands in `monitor/frontend`

- `npm ci` — exit 0.
- `npm test` — "Test Files 19 passed (19)", "Tests 176 passed (176)".
- `npm run lint` — "All files pass linting."
- `npm run format:check` — "All matched files use Prettier code style!"
- `npm run build` — "Application bundle generation complete."

I also wrote 9 probe tests of my own (Q1 to Q9), and I broke the code two times.

### BLOCKER 1 — the false test for the stable key: **CORRECTED**

`apps.spec.ts:470` now holds three rows. The second answer drops `appId` 2 and puts `appId` 3 before `appId` 1. The test compares the identity of each surviving `<tr>` node.

I set `track $index` at `apps.html:27` in my clone, and I ran the file. Result: **"Tests 1 failed | 33 passed (34)"**, with "AssertionError: expected `<tr>` to be `<tr>`" on that test. I restored the file. The test is real now.

My probe Q1 gives the same result.

### BLOCKER 2 — the absent state "a failed monitor API" of D30: **CORRECTED**

`app.ts:46-52` derives `monitorApiError` from `RefreshIntervalState.error()`, with the text "The app did not get the refresh interval."

My probe Q7 failed `GET /api/health` for 60 seconds of fake time. Measured:

- The health route failed **12** times (a retry each 5 seconds).
- The banner text reads "⚠ The app did not get the refresh interval."
- The shell called `Announcer.announce` **one** time in the 60 seconds, with "The monitor API stopped answering." The retry of 5 seconds makes **no** second announcement. The rule of #19 holds.
- A good answer then cleared the banner (`''`), and it gave exactly **one** more call, "The monitor API answers again."

My probe Q8 ran the real route `/apps` inside the shell. While the health route fails, the banner shows the text, the view sends **0** data requests, and the text "Octometer reads the app list." is no longer the only sign for the user.

I broke `app.ts` back to a constant `null`, and 4 tests of `app.spec.ts` failed. I restored the file. The tests are real.

### MAJOR 1 — a modified click on a data cell: **CORRECTED**

`apps.ts:127-135` reads the event, and it returns early for `event.button !== 0` and for each modifier key. My probe Q2 dispatched six clicks on a data cell: Ctrl, Meta, Shift, Alt, button 1, and button 2. `router.navigate` ran **0** times. A plain click then ran it exactly **1** time.

### The 7 MINOR findings of my first review

1. The empty "Data time" cell of a failed app — CORRECTED (`apps.html:68-76`, test at `apps.spec.ts:274`).
2. The mirror tests for the time format — CORRECTED. `apps.spec.ts:250` sets `vi.stubEnv('TZ', 'UTC')` and compares with the literal `'2026-09-21 09:05:03'`. This machine runs at UTC+3, and the test passes, thus the stub works and the test is not a mirror. `vi.unstubAllEnvs()` stands in a `finally` block.
3. No test for the state "zero events" — CORRECTED (`apps.spec.ts:263`).
4. The tenth field `gaps` of D13 — RESOLVED by the maintainer. #62 owns `gaps`, and it waits for #18 and for #20. The nine fields of `app-row.ts` match step 4 of #18 exactly.
5. The selection guard read the whole document — CORRECTED. `apps.ts:138-144` uses `Selection.containsNode(cell, true)`, with two tests.
6. The language of the comments — PARTLY CORRECTED. See MINOR 2 below.
7. The wrong owner of the banner in the pull request text — CORRECTED, under "Notes".

## What I checked again on this head

- `apps.html:11` puts the table in `<div class="table-scroll" tabindex="0" role="region" aria-labelledby="apps-caption">`, and `apps-caption` is the `<caption>` (probe Q4). The `<table>` keeps `display: table`.
- `apps.html:69-72` shows a `<time>` for each row with a `lastSuccessAt`, whatever the status.
- Each dash cell holds `<span class="visually-hidden">No data</span>` next to `<span aria-hidden="true">–</span>`.
- `.visually-hidden` moved from `app.css` to `src/styles.css`. `angular.json` names `src/styles.css` under the build `styles`, and I read the rule in the built file `dist/frontend/browser/styles-*.css`. `app.html:21` still uses the class for the status region, and the class now works in the `apps` component too (probe Q9).
- No shell test of #19 or of #93 is lost. I compared the `it(...)` names of `app.spec.ts` before and after: only the two old announcer tests changed their name, and the new names cover the same rules.
- A `lastError` with markup still shows as text (probe Q3). No `innerHTML`, no `bypassSecurityTrust`, and no `[href]` is in the diff.
- 20 navigations to the view and back leave 0 open requests and 0 timers (probe Q6).
- Test isolation: `apps.spec.ts` changes no property of `document`, of `window`, or of `navigator`. It adds no element to `document.body`. It restores the `TZ` stub, the fake timers, and each spy. `app.spec.ts` restores the fake timers, and it verifies the HTTP controller.
- The change of `app.ts` is 12 lines: one import, one `inject`, and the `computed`.

## MINOR findings

None of these stops the approval.

1. `angular.json` — the unit-test builder loads no global stylesheet, thus `src/styles.css` is absent in a test. My probe Q5 reads `getComputedStyle(el).position === ''` for a `.visually-hidden` element, and it finds no `.visually-hidden` rule in `document.styleSheets`. The tests check the class name only. No test can show that the text is out of sight. The build output is correct, thus this is a test gap and not a user defect. Fix: add `"styles": ["src/styles.css"]` to the `options` of the `test` target, then assert `position: absolute` and `width: 1px`.
2. `apps.ts:72-74` and `apps.ts:120-122` — two comment sentences still hold more than 25 words. Split each one.
3. `app.spec.ts:121-129` and `app.spec.ts:139-147` — `probe.remove()` stands outside a `finally` block. A failed assertion above it leaves the button in `document.body` with the focus, which is the defect class of #100. This code is older than this pull request (`git diff cb20a76 f2ffa5b` does not touch it), thus a later issue can fix it.
4. `apps.css:64` — the hover colour `#eef2f7` is a fixed value. The closing note of #25 asks for a custom property for a future dark theme.
5. `app.ts:38` — the shell now injects `RefreshIntervalState`, thus `GET /api/health` runs on `/manage` too, where no table stands. The count stays one request for each session (`take(1)` plus `shareReplay`), and the banner needs the state on each route, thus this is right. Name the reason in the comment of the field.

## Verdict

Each blocker and the major finding are corrected, and I proved each one with a broken build or with a probe test. The data flow, the refresh, the announcement count, the escape of each API value, the row track, the time format, and the test isolation are correct. I approve.

xamcross:
**Review (accessibility specialist): CHANGES REQUESTED**

Second review, on commit `f2ffa5b`. One new MAJOR finding remains. Each first finding is corrected.

## How I tested

A new clone at the head of `issue-20-app-table`. Node v24.21.0. `npm ci`, `npm test` (19 files, **176** tests pass), `npm run lint`, and `npm run build` each pass. I served the build with a fake API on a free port, with the four rows again: a normal app, a `NEVER_POLLED` app, an `UNREACHABLE` app with a long `lastError`, and an app name of 120 characters without a space. The fake API can switch `GET /api/health` to a failure, thus I tested the new banner state too. I measured in Chrome at 1280, 640, and 320 CSS px. I stopped each process, and I deleted the clone and each temporary file.

## The first findings

### BLOCKER 1 — the cell wrap — PARTLY CORRECTED

The text wrap is correct now. A residual defect remains, and I raise it as MAJOR A below.

What I measured:

| Viewport | "Clicks" cell | "Data time" cell | Column headers |
| --- | --- | --- | --- |
| 1280 CSS px | 1 line | 1 line | each on 1 line |
| 640 CSS px | 1 line | 1 line | each on 1 line |
| 320 CSS px | 1 line | 1 line | each on 1 line |

"1,234,642" stays on one line at each width. The header "Data time (GMT+3)" takes 2 lines at each width; it breaks at the space, thus that is a normal word wrap and not a defect. `overflow-wrap: anywhere` now applies only to `th[scope='row']`, `.data-from`, and `.error-text`. `.num` and `.data-time` carry `white-space: nowrap`.

The scroll container is correct:

- `<div class="table-scroll" tabindex="0" role="region" aria-labelledby="apps-caption">`, `overflow-x: auto`.
- The accessible name resolves to the caption text, "The registered apps, with the click totals of each app." I read it with `axe.commons.text`.
- The accessibility tree shows `region` and `table` inside it.
- One `Tab` from the button "Pause refresh" moves the focus to the container. `matches(':focus-visible')` gives true, and the computed `outline-style` is `auto`, thus the browser draws its focus ring. I saw the ring in a screenshot.
- At 320 CSS px the container holds a real scroll: `scrollWidth` 664, `clientWidth` 289.
- No table element carries a CSS `display` value: `table`, `table-caption`, `table-header-group`, `table-row-group`, `table-row`, `table-cell`.

**I did not measure the arrow-key scroll.** My browser tool cannot start a native scroll from a key press: a control test with `ArrowDown` on the page body also moved nothing. The container is focusable and it holds a real scrollable overflow, thus the browser gives the arrow keys. I state plainly that I did not see it with my own measurement.

### MAJOR 2 — the empty "Data time" cell — CORRECTED

Each row with a `lastSuccessAt` now shows a `<time>` element, whatever the status:

- `broken-shop`: `datetime="2026-09-21T09:05:03.000Z"`, text "2026-09-21 12:05:03".
- The long-name app: `datetime="2026-09-20T23:59:59.000Z"`, text "2026-09-21 02:59:59".

No cell is empty now. The visible date also removes my second point: a user reads the day of the last success, and not the time alone.

### MAJOR 3 — the bare dash — CORRECTED

The cell holds `<span class="visually-hidden">No data</span><span aria-hidden="true">–</span>`. The hidden span measures 1 by 1 CSS px with `clip-path: inset(50%)`, thus a sighted user sees only the dash. The accessibility tree of the `NEVER_POLLED` row gives four entries with the name "No data": three count cells and the "Data time" cell. `.visually-hidden` now stands in the global `monitor/frontend/src/styles.css`, thus #52 and #53 can use it.

### MAJOR 4 — the hover state and the cursor — CORRECTED

A real mouse hover gives the row the background `rgb(238, 242, 247)`, and each other row keeps its transparent background. `tbody td` computes `cursor: pointer`.

One measurement for the record. The hover background against a row at rest gives **1.124:1**. That is below 3:1. `cursor: pointer` is the second cue, and it does not use colour, thus the rule "3:1 or a second cue" holds. A low-vision user gets a weak hint from the colour alone. A darker value, for example `#dde5ef` (2.0:1) or `#c9d6e5` (3.1:1), helps that user. The text keeps its contrast on the hover background: 18.68:1 for the black text, 8.36:1 for the link.

### MINOR 5 — the link of the empty state — CORRECTED

The link "Go to the app management page." measures 210.6 by **24** CSS px, with `display: inline-flex` and `min-height: 24px`.

### MINOR 6 — the modified click — CORRECTED

I measured each case from `/apps`: a `Ctrl` click on a data cell, a `Shift` click, and a click with a button other than the primary button each leave the URL at `/apps`. A plain click still opens `/apps/2/users`.

### MINOR 7, 8, 9, 10 — CORRECTED or closed

7: no change needed, as I wrote. 8: the class comment of `apps.ts:69-74` is split into short sentences now. 9: the body of commit `f2ffa5b` states the fact correctly. 10: the new jsdom tests check the wrapper, the role, the name, and the `tabindex`; the layout itself still needs a browser, and this comment records that measurement.

## The new findings

### MAJOR A — `monitor/frontend/src/app/apps/apps.css:53-55` and `monitor/frontend/src/styles.css:8-19` — the page gets a horizontal scroll bar at 320 CSS px

**Mark for #52 and #53.**

`.visually-hidden` uses `position: absolute`. The spans of MAJOR 3 sit inside the table, and no ancestor is positioned, thus the containing block of each span is the initial containing block. Each span then escapes the clip of `.table-scroll`, and it adds its static position to the scrollable overflow of the page.

My measurement at a viewport of 320 CSS px, with the four rows:

| Case | `documentElement.scrollWidth` | `clientWidth` | Page scroll to the right |
| --- | --- | --- | --- |
| The branch as it stands | 523 | 305 | **218 px** |
| With the `.visually-hidden` spans hidden | 305 | 305 | 0 px |
| With `position: relative` on `.table-scroll` | 305 | 305 | 0 px |
| Back to the branch as it stands | 523 | 305 | **218 px** |

A user at 320 CSS px, or at a zoom of 400 percent, gets two horizontal scroll bars: one on the page and one on the table. The page bar scrolls 218 px into an empty area. WCAG 1.4.10 permits the two-dimensional scroll of the data table. It does not permit it for the page around the table.

The exact fix. One line:

```css
/* apps.css */
.table-scroll {
  position: relative;
  overflow-x: auto;
}
```

`position: relative` makes the container the containing block of each absolutely positioned descendant. The container then clips the hidden spans, and the page keeps one scroll direction. I measured the result: the page scroll goes to 0, and the table container still scrolls.

Please also add this line to the checklist of #52 and #53. Each view that holds a `.visually-hidden` span inside a scroll container needs it.

### MINOR B — `monitor/frontend/src/app/apps/apps.css:23-27` — a short app name breaks inside the word at 320 CSS px

**Mark for #52 and #53.**

`overflow-wrap: anywhere` on `th[scope='row']` sets the min-content width of the name column to one character. At 320 CSS px the name "traficio" renders as "traf / icio", and "fresh-app" as "fres / h- / app". The table already scrolls sideways, thus the column can take more width. The rule is right for a name of 120 characters; it is too strong for a short name.

A fix, optional:

```css
th[scope='row'] {
  overflow-wrap: anywhere;
  min-width: 8rem;
}
```

## The new state: a failed monitor API — PASS

I stopped the health route of my fake API, and I reloaded the page.

- A user sees the banner "⚠ The app did not get the refresh interval." The icon carries `aria-hidden="true"`, thus a screen reader reads the text one time. The banner slot keeps `min-height: 48px`, thus the page does not jump.
- The view shows "The app reads the refresh interval." and "Octometer reads the app list." No blank area appears.
- A user hears the status region one time: "The monitor API stopped answering."
- **The announcement does not repeat.** My fake API logged 12 retries of `GET /api/health` between 21:38:30 and 21:39:25, at 5 seconds. A MutationObserver on the status region recorded **0** changes in those 55 seconds.
- On the recovery the region announced "The monitor API answers again." one time, the banner went away, and the table came back.

## The axe scan

axe-core 4.13.0, with the tags `wcag2a`, `wcag2aa`, `wcag21a`, `wcag21aa`, `wcag22aa`, and `best-practice`.

| State | Violations | Incomplete | Passes |
| --- | --- | --- | --- |
| The first load | 0 | 0 | 35 |
| The empty list | 0 | 0 | 36 |
| The four rows (normal, `NEVER_POLLED`, failed, long name) | 0 | 1 | 43 |
| The failed monitor API | 0 | 0 | 36 |

The one incomplete item is `color-contrast` on the `aria-hidden` check mark, with the message "Element content contains only non-text characters". The glyph is black on white, 21:1. It is a false positive.

An axe scan does not replace a screen reader. It missed MAJOR A. **No NVDA and no JAWS runs on this machine**, thus no screen reader test ran. Issue #67 owns that check.

## The forced-colours mode

I could not switch the Windows setting from this session, and my browser tool gives no emulation. I state that plainly. By inspection `apps.css` now holds a `@media (forced-colors: active)` block that maps the hover background to `Highlight` and `HighlightText`. The status icon is a text glyph, thus the mode keeps it. The focus ring of the scroll container is the browser default, thus the mode keeps it too.

## The list for #52 and #53

The maintainer copies this list to the two issues.

1. One flat `<table>` with a `<caption>`, `<th scope="col">` for each column, and `<th scope="row">` for the name cell with the link inside it.
2. No CSS `display` value on any table element.
3. Put the table in `<div class="table-scroll" tabindex="0" role="region" aria-labelledby="<caption id>">` with `overflow-x: auto`. Add `position: relative` to that container (MAJOR A).
4. Apply `overflow-wrap: anywhere` only to the name cell and to a free-text cell. Give `white-space: nowrap` to each number column and to each time column.
5. Never leave a data cell empty. Use `<span class="visually-hidden">No data</span><span aria-hidden="true">–</span>` from the global `src/styles.css`.
6. Show a time as `<time datetime="...Z">yyyy-MM-dd HH:mm:ss</time>` for each row that holds the value, whatever the status. Name the zone in the column header.
7. Show a status as an icon plus a text. Give the icon `aria-hidden="true"`. Never use colour alone.
8. Give each data cell `cursor: pointer`, and give a row a `:hover` and a `:focus-within` state. Add a `forced-colors` variant. Keep the text contrast on the new background.
9. Let the row click read the event. Leave a modifier key and a secondary button to the browser. Stop the navigation while a selection touches the clicked cell.
10. Keep the row `<tr>` free of `tabindex`, of `role`, and of any ARIA attribute. Only the link serves a keyboard user.
11. Give each target a minimum of 24 by 24 CSS px, with `display: inline-flex` plus `min-width` and `min-height`.
12. Keep `RefreshBar` between the `<h1>` and the table.
13. Track each row with its stable key, so a refresh keeps the DOM node, the focus, and the scroll position.
14. Show a text for the first load and for the empty list. Leave no blank area.
15. Measure the reflow in a real browser at 1280, 640, and 320 CSS px. `jsdom` runs no layout, thus no unit test can see a line break.

## Summary

- 1 MAJOR (new), 1 MINOR (new). Each finding of the first review is corrected.
- MAJOR A needs one line of CSS. I will approve as soon as that line lands.
- No secret, no database file, and no build output is in the diff.

xamcross:
Thank you for the second review. Both new findings are corrected on this branch, in commit 4709245.

**MAJOR A**, `monitor/frontend/src/app/apps/apps.css:53-55`: `.table-scroll` now holds `position: relative`, with a one-sentence comment that gives the reason. I saw the line in the file after the edit.

**MINOR B**, `apps.css:23-27`: `th[scope='row']` now holds `min-width: 8rem`. I saw the line in the file after the edit.

A new jsdom test reads the `.table-scroll` rule from the stylesheet and checks its `position` value. `jsdom` runs no layout, so this test cannot see the scroll bar itself; the reviewer's real-browser measurement is the check that matters for the layout claim.

I also measured the fix in a real browser (Chrome, through the Claude in Chrome extension), at a true 320 CSS px viewport. Before the fix, `document.documentElement.scrollWidth` read 562 against a `clientWidth` of 320. After the fix, both read 320, with no page scroll. `th[scope='row']` held a computed `min-width` of 128px, and the name "traficio" rendered on one line.

`npm test` (177 tests), `npm run lint`, `npm run format:check`, and `npm run build` each pass in `monitor/frontend`. CI is green.

xamcross:
**Review (accessibility specialist): APPROVE**

Third accessibility review, from a new reviewer. I confirm the correction of commit `4709245`. MAJOR A is corrected. MINOR B is corrected. I found no new BLOCKER and no new MAJOR finding. I list one new MINOR finding, and it does not prevent the approval.

## How I tested

A fresh clone at commit `4709245`, the head of `issue-20-app-table`. Node v24.21.0.

| Command | Result |
| --- | --- |
| `npm ci` | pass |
| `npm test` | 19 files, **177** tests pass |
| `npm run build` | pass, exit code 0 |

I wrote a small Node script. It serves the production build and the fake API on one origin, on a free port. The API gives `GET /api/health` with `refreshSeconds` 600, and `GET /api/apps` with four rows: `traficio` (`OK`), `fresh-app` (`NEVER_POLLED`), `broken-shop` (`UNREACHABLE`, with a `lastError` of 232 characters), and an app with a name of 120 characters without a space.

I measured in Chrome. **The method for the viewport width:** the window of the machine did not accept a resize below its own minimum, so I put the app in an `<iframe>` of an exact CSS pixel width. The iframe gives the document its own viewport, so `documentElement.scrollWidth` and `clientWidth` inside it answer the reflow question directly. A classic vertical scroll bar takes 15 px of the iframe, so a viewport of 320 CSS px gives a layout width of 305 CSS px. For the text zoom of 200 percent I set `html { font-size: 32px }` and `button, input, select, textarea { font-size: inherit }`. No rule in the project sets a `font-size` in px, so each text doubles. That is the test of WCAG 1.4.4.

I stopped each process. I deleted the clone, the script, and each temporary file. I closed the browser tab.

## 1. The page scroll bar — PASS

`document.documentElement.scrollWidth` equals `clientWidth` at each of the four points.

| Viewport | `scrollWidth` | `clientWidth` | Page scroll | `.table-scroll` `scrollWidth` / `clientWidth` |
| 1280 CSS px | 1280 | 1280 | **0 px** | 1264 / 1264, no scroll needed |
| 640 CSS px | 625 | 625 | **0 px** | 763 / 609, the table scrolls |
| 320 CSS px | 305 | 305 | **0 px** | 763 / 289, the table scrolls |
| 320 CSS px, text zoom 200 % | 305 | 305 | **0 px** | 1512 / 289, the table scrolls |

The control test proves that my harness sees the defect. At 320 CSS px I switched `.table-scroll` back to `position: static` in the live page, and I switched it back again:

| Case | `scrollWidth` | `clientWidth` | Page scroll |
| --- | --- | --- | --- |
| The branch as it stands | 305 | 305 | 0 px |
| With `position: static` | 621 | 305 | **316 px** |
| Back to the branch | 305 | 305 | 0 px |

I also read the cause directly. `offsetParent` of each hidden span in the table is now the element `.table-scroll`. Before the fix it was the initial containing block. That is the exact mechanism of MAJOR A, and the fix closes it.

The fix adds no scroll inside the container either. The container `scrollWidth` is 763 with the spans, and 763 with the spans set to `display: none`. Each hidden span sits inside the box of the table: its right edge is at 232, and the right edge of the table is at 771.

## 2. The line breaks — PASS

At 1280, 640, and 320 CSS px, and at 320 CSS px with the text zoom of 200 percent:

- Each number cell takes **1 line**: "1,234,642", "87,234", "145,902", "998,877", "12,045", "33,110", "42", "7", "9".
- Each "Data time" cell takes **1 line**: "2026-09-21 12:12:41", "2026-09-21 12:05:03", "2026-09-21 02:59:59".
- The time inside the status cell, "12:05:03", takes **1 line**.
- Each column header takes 1 line, except "Data time (GMT+3)". That one takes 2 lines. I read the two lines with a `Range`: "Data time " and "(GMT+3)". The break sits at the space. The computed `overflow-wrap` and `word-break` of a column header are both `normal`, so that header cannot break inside a word. That is a normal word wrap and not a defect.
- The short names stay on 1 line: "traficio", "fresh-app", "broken-shop". At 320 CSS px the name column is 152 CSS px wide, and the computed `min-width` is 128 px (`8rem`).
- The name of 120 characters wraps inside its cell: 4 lines at 1280 CSS px, 8 lines at 320 CSS px. It does not leave the cell.

`min-width: 8rem` uses `rem` and not `px`. At the text zoom of 200 percent it grows to 256 px, so the column keeps room for the same number of characters. That is the correct unit here.

## 3. The hidden "No data" texts — PASS

I read the accessible text of each cell of the `NEVER_POLLED` row with `axe.commons.text.accessibleText`:

| Cell | Accessible text | Visible glyph |
| --- | --- | --- |
| Name | "fresh-app" | the link text |
| Clicks | **"No data"** | "–", 8 CSS px wide |
| Unique users | **"No data"** | "–", 8 CSS px wide |
| Unique sessions | **"No data"** | "–", 8 CSS px wide |
| Status | "Not polled yet" | "– Not polled yet" |
| Data time | **"No data"** | "–", 8 CSS px wide |

Each `.visually-hidden` span measures 1 by 1 CSS px, with `clip-path: inset(50%)` and `overflow: hidden`. Each one keeps `display: block` and `visibility: visible`, and no ancestor carries `aria-hidden`. The texts stay in the accessibility tree, and a sighted user sees only the dash. The clip of the container does not remove them from the tree.

## 4. The scroll container — PASS

| Point | Measurement |
| --- | --- |
| `role` | `region`; `axe.commons.aria.getRole` also gives `region` |
| `tabindex` | `0` |
| Accessible name | "The registered apps, with the click totals of each app." — the caption text, through `aria-labelledby="apps-caption"` |
| Focus from a real `Tab` | One `Tab` from the button "Pause refresh" moves the focus to the element `div.table-scroll` |
| Focus ring | `matches(':focus-visible')` gives true, and the computed `outline-style` is `auto`. I saw the ring around the container in a screenshot. |
| Real scrollable overflow | `scrollWidth` 763, `clientWidth` 289 at 320 CSS px. `scrollLeft = 200` holds at 200. |

**The stacking does not change.** `.table-scroll` keeps `z-index: auto`, so `position: relative` builds no new stacking context. I walked the chain from `.table-scroll` to the root: `app-apps`, `main`, `app-root`, and `body` each build no stacking context, so the container stays in the root stacking context. The skip link keeps `z-index: 100`, so it paints above. I proved that with a hit test: I moved the skip link over the table with an inline style, and `elementFromPoint` then gave `a.skip-link` and not the cell below it. I removed the inline style, and the cell came back. The banner and the header do not touch the container: the banner box ends at y 56, the header box ends at y 88, and the container box starts at y 239. The focus ring is an `outline`, which the paint order does not move.

**What I could not measure: the arrow-key scroll.** I pressed a real `ArrowRight` six times with the focus on the container. The six `keydown` events reached the document, and `scrollLeft` stayed at 0. I then ran a control test: with the focus on `<body>`, six real `ArrowDown` presses also moved `scrollTop` from 0 to 0, although the document had 351 px of vertical scroll left. My browser tool therefore cannot start a native scroll from a key press. This is the same limit that the second reviewer reported. I state plainly that I did not see the arrow-key scroll with my own measurement. What I did measure: the container takes the focus from a real `Tab` key, and it holds a real scrollable overflow that accepts a scroll.

## 5. No CSS `display` value on a table element — PASS

The computed values are the native ones:

| Element | `display` |
| --- | --- |
| `table` | `table` |
| `caption` | `table-caption` |
| `thead` | `table-header-group` |
| `tbody` | `table-row-group` |
| `tr` | `table-row` |
| `th` | `table-cell` |
| `td` | `table-cell` |

## 6. The axe scan — PASS

axe-core 4.10.2, with the tags `wcag2a`, `wcag2aa`, `wcag21a`, `wcag21aa`, `wcag22aa`, and `best-practice`. Each scan ran on the state with the four rows: `OK`, `NEVER_POLLED`, `UNREACHABLE` with the long `lastError`, and the name of 120 characters.

| Viewport | Violations | Incomplete | Passes |
| --- | --- | --- | --- |
| 1280 CSS px | **0** | 1 (`color-contrast`) | 41 |
| 640 CSS px | **0** | 1 (`color-contrast`) | 42 |
| 320 CSS px | **0** | 1 (`color-contrast`) | 42 |
| 320 CSS px, text zoom 200 % | **0** | 1 (`color-contrast`) | 42 |

The one incomplete item is `color-contrast` on the status glyph with `aria-hidden="true"`, with the message "Element content contains only non-text characters". That is the same false positive that the second reviewer reported.

I did not scan the empty list and the failed monitor API again. The second review measured both, and commit `4709245` touches neither.

## The other acceptance criteria that I measured

- **`NEVER_POLLED` shows "–", not 0** — PASS. Each count cell of `fresh-app` shows "–" and gives "No data" to a screen reader.
- **A failed app keeps its numbers, an icon plus text, "Data from HH:mm:ss", and the visible `lastError`** — PASS. `broken-shop` shows 998,877 / 12,045 / 33,110, "⚠ Unreachable", "Data from 12:05:03", and the full error text of 232 characters. `scrollHeight` equals `clientHeight`, and `text-overflow` is `clip`, so no text is cut.
- **The table has a `<caption>` and `<th scope>`** — PASS.
- **Each target is a minimum of 24 by 24 CSS px** — PASS. At 320 CSS px each name link measures 24 CSS px high, and 49.8, 65.5, 86, and 128 CSS px wide.
- **`npm test` passes** — PASS, 177 tests.

## The diff of commit `4709245`

`git show --stat 4709245` gives two files and 14 added lines, with no deletion:

- `monitor/frontend/src/app/apps/apps.css`, +6: the declaration `min-width: 8rem` in a new rule for `th[scope='row']`, the declaration `position: relative` in `.table-scroll`, and one comment line.
- `monitor/frontend/src/app/apps/apps.spec.ts`, +8: one test.

That is the two CSS declarations, one comment, and one test, and nothing else. No secret, no database file, and no build output stands in the diff of the pull request.

The new test is real. `findCssRule('.table-scroll')` reads the rule from the stylesheet, and the expectation is `toBe('relative')`. Without the fix the value is the empty string, so the test fails. It is not always green. As the commit body states, `jsdom` runs no layout, so this test cannot see the scroll bar. My measurement above is the proof of the layout.

The commit body follows ASD-STE100. The new CSS comment follows it too: one sentence, the active voice, and 20 words.

## The findings

### MINOR C — `monitor/frontend/src/app/apps/apps.css:56-57` — two comment lines stand above one rule

Two separate single-line comments now stand above `.table-scroll`. A reader sees two thoughts and one rule. A block comment of two sentences reads better, and it matches the style of the other comments in this file.

An optional fix:

```css
/*
 * The table scrolls sideways when the columns need more width than the
 * screen. `position: relative` makes this container the containing block of
 * each `.visually-hidden` span, so the page keeps one scroll direction.
 */
.table-scroll {
  position: relative;
  overflow-x: auto;
}
```

This is style alone. It does not prevent the approval.

## Summary

- 0 BLOCKER, 0 MAJOR, 1 MINOR.
- MAJOR A is corrected. At each of 1280, 640, and 320 CSS px, and at 320 CSS px with a text zoom of 200 percent, the page holds no horizontal scroll bar. The table scrolls inside its container.
- MINOR B is corrected. A short name stays on one line, and the name of 120 characters still wraps inside its cell.
- `position: relative` changes no stacking. The focus ring, the banner, and the skip link each keep their place.
- I could not measure the arrow-key scroll. My tool cannot start a native scroll from a key press, and a control test showed the same.
- **APPROVE.**

PR #123, opened 2026-09-21
feat(kit): add the Ktor ingest route in kit/jvm-ktor

xamcross:
## Summary

This pull request adds the module `kit/jvm-ktor`. It is the only Kotlin
module of the kit (design decision D17). It mounts the ingest route of
design section 4.2 with one function call, `octometerIngestRoute`. The
function reads the raw body as a `String`, then it calls
`IngestPipeline.ingest` of `kit/jvm-core` inside a bounded view of
`Dispatchers.IO` (design decision D23). The route maps each defect of the
app or of the store to status 500, with an empty body and a log line that
holds only the exception class name (design decision D15).

**Correction round 1 (2026-09-22).** The security review and the Ktor
review of the first push found one blocker each, plus several major and
minor points. Each finding, its fix, and its evidence are in this
[response comment](https://github.com/xamcross/octometer/pull/123#issuecomment-5767578454).
The four points below are the widest fixes:

- The route now catches each defect itself, answers 500 with an empty
  body, and logs only the exception class name; the earlier code let an
  exception reach the Ktor engine, which echoed the exception message
  into the answer and into the log.
- The user id resolver now runs on the coroutine of the call, before the
  store call moves to a dispatcher; a value that the app keeps in a
  `ThreadLocal` needs this.
- The store call now runs on a bounded view of `Dispatchers.IO`
  (`Dispatchers.IO.limitedParallelism(8)`), so the route cannot fill the
  whole shared pool of the app.
- A malformed `Content-Type` header now gives 415, and the class name and
  API version pins fix two portability points (see "Confirm results").

**Correction round 2 (2026-09-22).** The Ktor reviewer approved. The
security reviewer confirmed each first finding as corrected, and found one
new MAJOR: a caught `CancellationException` was rethrown at once, so a
`withTimeout` or a cancelled `Future` inside the store still leaked the
answer, gave a 504 status, or wrote no log line. Each finding, its fix, and
its evidence are in this
[response comment](https://github.com/xamcross/octometer/pull/123#issuecomment-5767954812).

- The route now rethrows a `CancellationException` only when the
  coroutine of the call is no longer active. Each other case gets the
  same empty 500 answer and the same one-line log as another defect of
  the app (`IngestRoute.kt:138-152`, `IngestRoute.kt:167`).
- Three small Ktor review points are fixed too: a comment now names the
  `kotlinx-coroutines` version of the brittle dispatcher test, one flaky
  timing assertion now uses a `CountDownLatch` instead of a wall clock
[...]

Issue #105, opened 2026-09-21
Write the new fields in the MongoDB store


xamcross:
**Goal.** `kit/jvm-mongo` writes `path` and `referrerHost` into the event document of the app.

**Implementation steps.**
1. In `kit/jvm-mongo`, write `path` and `referrerHost` into the event document.
2. Write a field only when the event record holds it. Add no key with a null value.
3. Add no new index, because no query of the app reads these fields.
4. Add the Testcontainers tests with `@Testcontainers(disabledWithoutDocker = true)`.

**Acceptance criteria.**
- [x] An event with the two fields stores the two fields with the same names.
- [x] An event without a field stores no key for that field.
- [x] A click event stores `path` and no source field.
- [x] The stored document holds no `_class` field (rule C11).
- [x] The collection holds the two indexes of rule C8 and no other index.
- [x] The store writes the component `path` of the event record as it is, and it reads no other source for `path`. That component holds only the match result of the server, never the raw client value.
- [x] `./gradlew :kit:jvm-mongo:test` passes against driver 5.0.1 and the newest 5.x.
- [x] A comment on this issue records the test output.

**Related.**
- #103 carries `referrerHost` and the component `path` to `append`. #104 fills `path` with the match result.
- #11 gives the store.
- #35 owns the erasure of the same documents.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, the contract version 1.1 of #101, and the rule K1.

xamcross:
Body edit on 2026-09-22: #11 closed. The first line now names only the open blockers: #103.

xamcross:
Body edit on 2026-09-22 by the maintainer: one new criterion and one changed line in Related, after the reviews of #131 (https://github.com/xamcross/octometer/pull/131#issuecomment-5768524586 and https://github.com/xamcross/octometer/pull/131#issuecomment-5768610379). The event record of #103 has one component `path`. It holds only the match result of the server, and it is empty until #104 merges. The tests of this issue build an event record with a `path` value directly, thus this issue does not wait for #104. The steps do not change.

xamcross:
Body edit on 2026-09-22: #103 closed. No open blocker remains, thus the first line and the label `blocked` are removed. The work can start.

xamcross:
Test output for pull request #154 (JDK 21, Docker present).

`./gradlew :kit:jvm-mongo:test` (driver 5.0.1): 31 tests, 0 failures, 0 errors, 0 skipped.

[...]

PR #126, opened 2026-09-21
feat(jvm-mongo): add the MongoDB event log store

### LINKED ISSUE #11: Build the MongoDB event store in kit/jvm-mongo

xamcross:
## Summary

This pull request adds the module `kit/jvm-mongo`. The module holds
`MongoEventLogStore`, the MongoDB `EventLogStore` of design decision D22.
The app gives the `MongoDatabase`; the kit never creates a client. The
driver `org.mongodb:mongodb-driver-sync` is `compileOnly`.

This correction round answers the two first reviews (3 MAJOR and 11 MINOR
from the Java and MongoDB engineer, 3 MAJOR and 5 MINOR from the security
and privacy engineer). Each acceptance criterion still passes.

## Acceptance criteria

- [x] A stored document has exactly the fields of section 4.1 and no `_class` field.
  Evidence: `MongoEventLogStoreTest.appendStoresExactlyTheFieldsOfSection41WithNoClassField`
  asserts the key set `{_id, ts, element, sessionId, userId}` against a real MongoDB
  server (Testcontainers). The test passes.
- [x] A second start with a different `OCTOMETER_RETENTION_DAYS` changes
  `expireAfterSeconds` and does not fail.
  Evidence: `MongoEventLogStoreTest.theCollectionKeepsExactlyTheDefaultIdIndexAndTheTtlIndexOnTs`
  and `aSecondStartWithADifferentRetentionChangesExpireAfterSecondsAndDoesNotFail`
  build the store twice, with 30 days then 10 days, against a real MongoDB
  server, and assert the exact index set, name, key, and `expireAfterSeconds`.
- [x] An index error gives a warn-level log line, and the store still works.
  Evidence: `MongoEventLogStoreIndexErrorTest` (Mockito, no Docker) covers
  error code 85 (`collMod`, no warning), a different `createIndex` error
  (one `WARNING` line that names the step and the numeric error code, and
  states that the collection now has no TTL index), a `collMod` failure (one
  `WARNING` line that states the old retention value stays), a retention
  above 24855 days (clamped, with its own warning), and a retention below 1
  day (rejected). `MongoEventLogStoreInsertOnlyUserTest` proves the same
  `createIndex` path against a real MongoDB server and a database user that
  holds only `insert`: one warning, no TTL index, and `append` still writes
  the event.
- [x] `./gradlew :kit:jvm-mongo:test` passes against driver 5.0.1 and the newest 5.x.
  Evidence: `./gradlew :kit:jvm-mongo:test` runs against 5.0.1 (25 tests, 0
  failures, 0 skipped). `./gradlew :kit:jvm-mongo:testMongoSyncNewest` runs the
  same suite against 5.12.0 (25 tests, 0 failures, 0 skipped). Both ran with
  Docker Desktop on the local machine; no test skipped. `./gradlew
  :kit:jvm-core:test` also passes (81 tests, 0 failures, 0 skipped).
- [x] A comment on the issue records the test output for the two driver versions.
  Evidence: see the third comment of #11, from the first author.

## Confirm results

- Step 6, the newest 5.x driver version. Source:
  `repo1.maven.org/maven2/org/mongodb/mongodb-driver-sync/maven-metadata.xml`.
  Date: 2026-09-21. Result: `5.12.0`. The compile version stays `5.0.1`, the
  version of the Spring Boot 3.3 BOM (design decision D22); the test suite
  runs a second time against `5.12.0` with the Gradle task `testMongoSyncNewest`.

## Notes for the maintainer

- `append` sends one ordered `insertMany` call for the whole batch, with no
  transaction. A failure in the middle of the batch keeps each event before
  the failure, and drops each event after the failure. The class Javadoc of
  `MongoEventLogStore` states this, together with the retry chain of design
  decision D24 and contract rule C24 (at-least-once delivery). The Javadoc
  of `EventLogStore.append` in `kit/jvm-core` now states the same promise
  for each store: a store writes the batch as one unit when it can, or it
  writes the first part and then throws.
- A failed write no longer leaves the app with a raw driver exception. It
  throws `MongoEventLogStore.EventLogWriteException`, an unchecked type
  with a fixed message and the numeric MongoDB error code, and no cause.
  The message and the full stack trace hold no host, no port, and no
  database name (design decision D15). `MongoEventLogStoreWriteErrorTest`
  proves this with a mocked driver exception that carries a real-looking
  host, port, and database name.
- A retention above 24855 days no longer leaves the collection with no TTL
  index. `expireAfterSeconds` is a 32-bit whole number on the server, so
  the store now clamps the value to 24855 days (the largest value that
  fits), and it logs a warning that names both the requested value and the
  clamp.
- `deleteByUserId` throws `UnsupportedOperationException` with a message
  that states plainly that the store deleted no event. Issue #35 adds the
  real delete.
- The class Javadoc now states that the store sets no write concern, no
  read preference, and no timeout, and that an app that calls `append` from
  a thread pool must set its own timeout (`timeoutMS` on driver 5.2 or
  newer, or a socket read timeout plus `wtimeout` on driver 5.0).
- `gradle/libs.versions.toml` now holds `junit`, `mockito`, and
  `testcontainers`, and `kit/jvm-mongo/build.gradle.kts` reads them. The
  `typesafe-config` entry moved back to its original place in `[versions]`,
  to keep the merge risk with pull request #123 low.
- The task `kit:jvm-mongo:test` (and `testMongoSyncNewest`) now fails when
  a test skips and the environment variable `CI` holds `true`. This guards
  against a silent Docker-less skip on the CI runner. The console log of
  each PASSED and SKIPPED test line is removed; the new guard makes that
  log unnecessary.
- This pull request adds no `path` field, no `referrerHost` field, no event
  cap, and no delete logic. Issues #105, #34, and #35 own these.
- The CI job "JVM modules" runs on Ubuntu with Docker, so the Testcontainers
  tests run there too, not only on the local machine.

Closes #11

Generated with Claude Code


xamcross:
**Review (security and privacy engineer): CHANGES REQUESTED**

I made my own clone, and I checked out the head of `issue-11-mongo-store`. I ran each test
command myself. I also wrote my own tests against a real MongoDB 7.0 server with
authentication, and I removed them before the final run. The store is clean on the two points
that matter most: no injection path exists, and the document holds only the five fields of
section 4.1. Three findings stop an approval.

## Acceptance criteria

|---|---|---|
| A stored document has exactly the fields of section 4.1 and no `_class` field. | PASS | `./gradlew :kit:jvm-mongo:test` passes. My own test read the raw BSON document for 12 hostile values. The key set was always `{_id, ts, element, sessionId, userId}`. |
| A second start with a different `OCTOMETER_RETENTION_DAYS` changes `expireAfterSeconds`. | PASS | With a user that holds `insert`, `createIndex`, and `collMod`, the TTL went from 2592000 to 864000, and no warning appeared. |
| An index error gives a warn-level log line, and the store still works. | PASS | With a user that holds only `insert`, the store logged exactly one WARNING line, and `append` still wrote the event. |
| `./gradlew :kit:jvm-mongo:test` passes against driver 5.0.1 and the newest 5.x. | PASS | 18 tests, 18 passed, 0 skipped, for each of the two tasks. Docker was active on my machine. |
| A comment on the issue records the test output for the two driver versions. | PASS | The third comment of #11 holds the two counts. |

## The nine points of my role

**1. Injection — PASS.** `MongoEventLogStore.java:93-99` builds the document with literal keys
and typed values. There is no `Document.parse`, no JSON text, and no filter that holds a user
value. `runCollMod` (line 119) takes only the constant `COLLECTION_NAME` and a `long`. I fed 12
hostile texts into `userId`, `element`, and `sessionId`: `$gt`, `$where`, `$$ROOT`, `$expr`,
`{"$ne": null}`, `{"$set": {"userId": "admin"}}`, `{"_class": "octometer.Evil"}`, `a.b.c`,
`..`, a text with a NUL character, and two more. Each one reached the server as a BSON STRING
with the exact input text. No value became a key. No key started with a dollar sign or held a
dot. An exact string match found each document again.

**2. Data minimisation (C6, O5) — PASS.** `toDocument` (line 93) appends four named fields. The
driver adds `_id`. The field is a `MongoCollection<Document>` (line 52), so no POJO codec runs,
and Spring Data cannot add `_class`. There is no `Map`, no `putAll`, and no generic object, so
a later field needs a change of this method. `IngestEvent` holds three components only. No IP
address, no user agent, and no request data reach the document.

**3. The collection name (C1) — PASS.** `COLLECTION_NAME` is a `static final` constant (line
43). It is the only argument of `getCollection` (line 69) and of the `collMod` command (line
119). The app gives the `MongoDatabase` only. The app cannot change the collection name.

**4. Logs and exceptions (D15) — PARTIAL.** The two warn lines hold a constant text. The
retention warning names the variable and never repeats the value. I checked the captured log
against a host, a port, a database name, a user name, and the name space: each search was
empty. The `UnsupportedOperationException` message holds no user id. The store drops the driver
exception of `createIndex` and of `collMod`, so the driver text stays out of the log. But the
exception of `insertMany` reaches the app with a host, a port, and a database name inside it.
See MAJOR 1.

**5. The rights in the database of the app — PASS, with a note.** I measured them against a
server with authentication:

- `insert` on `octometer_events`: the store writes each event.
- `createIndex` on `octometer_events`: the store creates the TTL index.
- `collMod` on `octometer_events`: the store changes `expireAfterSeconds` on a later start.

With only `insert`, the store works and logs one warning, and the collection holds **no** TTL
index. The minimum role is therefore a custom role with the three actions above, limited to
`octometer_events`. The built-in role `readWrite` is not enough: it holds `insert` and
`createIndex`, and it does not hold `collMod`. I confirmed with `usersInfo` and `rolesInfo`
that the store creates no user and no role. The read-only user of the monitor (section 4.4)
stays separate.

**6. The TTL value — PASS.** I tested 22 texts: `0`, `-1`, `-2147483648`, `abc`, an empty text,
a text of spaces, `1e3`, `0.5`, `0x10`, `2147483648`, `9999999999999999`, `30d`, `NaN`,
`Infinity`, `null`, `1 second`, `+0`, `-0`, and more. Each one gave 30 days and one warning. No
text gave a short retention. The multiply at line 102 casts to `long` first, so it cannot
overflow: `Integer.MAX_VALUE` days gives 185542587100800 seconds. A value that large makes the
server refuse the index, so the collection keeps **no** TTL index, and no event dies early.

**7. `deleteByUserId` — PASS, with a MINOR.** The method throws before it touches the database,
so an app cannot believe that an erasure succeeded. The message names issue #35. It does not
state that no event was deleted. See MINOR 3.

**8. The supply chain — PARTIAL.** The driver is `compileOnly` (line 36). Testcontainers,
Mockito, and JUnit are test-only. The image tag `mongo:7.0` is fixed, and it is not `latest`.
The file `.github/workflows/ci.yml` is outside the diff, so there is no new GitHub Action and
no new permission; `permissions: contents: read` stays. The job `Build each JVM module` runs
`./gradlew build`, and `tasks.check` depends on `testMongoSyncNewest`, so both driver versions
run on CI. The test tasks print a test name only, so the CI log holds no connection detail.
Three versions sit outside the version catalog. See MAJOR 3.

**9. Secrets — PASS.** I searched the diff, the pull request text, and the issue comments for a
connection string of the placeholder form `mongodb+srv://<user>:<password>@<host>/<db>`, for a
password, for a token, and for a private key. There is no match. No database file and no build
output is in the diff. The job `Secret scan` runs `gitleaks` on each pull request.

## Findings

### MAJOR 1 — the write exception carries a host, a port, and a database name

`kit/jvm-mongo/src/main/java/octometer/kit/mongo/store/MongoEventLogStore.java:84`

`append` lets the raw driver exception reach the app. I gave the app user the right
`createIndex` and not `insert`, then I called `append`. The driver threw a
`MongoCommandException` with this message shape:

```
Command failed with error 13 (Unauthorized): 'not authorized on <database> to execute
command { insert: "octometer_events", ordered: true, $db: "<database>", ... }' on server
<host>:<port>. The full response is {...}
```

Good news first: the trace held no user id, no element, and no session id, because the driver
sends the documents in a separate message section. But the trace held the host, the port, and
the database name of the app. Design decision D15 states that a log line never holds a
connection string. An adapter maps this exception to status 500, and an app logs a 500. The
kit therefore hands the app a line that breaks D15, and no javadoc warns the app.

**The fix.** Wrap the write in a kit exception with a safe message and no cause:

```java
/** A failed write. The message holds no connection detail (design decision D15). */
public static final class EventLogWriteException extends RuntimeException {
    private final int errorCode;

    EventLogWriteException(int errorCode) {
        super("The store could not write the click batch to the event collection. "
                + "The MongoDB error code is " + errorCode + ".");
        this.errorCode = errorCode;
    }

    public int errorCode() {
        return errorCode;
    }
}
```

In `append`, replace line 84 with:

```java
try {
    collection.insertMany(documents, new InsertManyOptions().ordered(true));
} catch (MongoCommandException e) {
    throw new EventLogWriteException(e.getErrorCode());
} catch (RuntimeException e) {
    throw new EventLogWriteException(-1);
}
```

The kit exception carries no cause, so a full stack trace also stays clean. State the rule in
the javadoc of `append`. If you prefer to keep the driver cause for a debugger, keep it, and
state in the javadoc that an adapter logs only `getMessage()` of the kit exception. Please pick
one and write it down.

### MAJOR 2 — the TTL warning hides its own consequence

`kit/jvm-mongo/src/main/java/octometer/kit/mongo/store/MongoEventLogStore.java:128-131`

The line says "The app start continues with no TTL index change." An operator reads this and
understands that an old TTL index stays. I measured a second case: with a user that holds only
`insert`, the collection holds **no** TTL index at all, and the same line appears. The store
then keeps each click event for ever. The retention control of rule C8 silently does not apply,
and the one signal does not say so. The line also drops the error code, so the operator cannot
tell an authorisation error from a conflict.

**The fix.** Pass the error code, and state the consequence:

```java
private static void warnIndexError(int errorCode) {
    LOGGER.log(Level.WARNING, "The store could not set up the TTL index on ts. "
            + "The MongoDB error code is " + errorCode + ". The app start continues. "
            + "The event collection can hold each event longer than "
            + "OCTOMETER_RETENTION_DAYS, until an administrator creates the TTL index.");
}
```

Call `warnIndexError(e.getErrorCode())` at line 110. Call `warnIndexError(-1)` at line 113 and
at line 124, because a `RuntimeException` has no error code. An error code is a number. It is
neither personal data nor a connection detail, so D15 holds.

### MAJOR 3 — three versions sit outside the version catalog

`kit/jvm-mongo/build.gradle.kts:31-32` and `kit/jvm-mongo/build.gradle.kts:38`

`testcontainersVersion = "2.0.5"`, `mockitoVersion = "5.23.0"`, and `org.junit:junit-bom:6.1.3`
are local values. A supply-chain audit and a later bump need one place for each version. The
comment at lines 13 to 15 says that `monitor/backend` has no shared entry yet. That reason does
not hold here: this same pull request proves that a version-only entry works, because
`mongodb-driver-sync` and `mongodb-sync-newest` sit in `[versions]` with no `[libraries]` entry,
and the build reads them with `libs.versions.<name>.get()`. `kit/jvm-core/build.gradle.kts:20`
at least names a follow-up; this module names none.

**The fix.** Add three lines to `gradle/libs.versions.toml`, in the `[versions]` block:

```toml
junit = "6.1.3"
mockito = "5.23.0"
testcontainers = "2.0.5"
```

Then read them in `kit/jvm-mongo/build.gradle.kts`:

```kotlin
val junitVersion = libs.versions.junit.get()
val testcontainersVersion = libs.versions.testcontainers.get()
val mockitoVersion = libs.versions.mockito.get()
```

Use `platform("org.junit:junit-bom:$junitVersion")` at line 38. Keep the comment that states
the confirm date of each version.

### MINOR 1 — the MongoDB test image is not pinned by digest

`kit/jvm-mongo/src/test/java/octometer/kit/mongo/store/MongoEventLogStoreTest.java:38`

The tag `mongo:7.0` is correct, and it is not `latest`. The tag still moves with each 7.0 patch
release. `.github/workflows/ci.yml` pins each GitHub Action by a commit SHA, so a digest is the
house style. Consider `mongo:7.0@sha256:<digest>`.

### MINOR 2 — the package-private constructor accepts a retention of zero days

`kit/jvm-mongo/src/main/java/octometer/kit/mongo/store/MongoEventLogStore.java:67`

`retentionDaysFromValue` never returns a value below 1, so the public constructor is safe. The
package-private constructor takes any `int`. I called it with `0` against a real server, and
the collection got a TTL index with `expireAfterSeconds: 0`. That index deletes each event as
soon as its `ts` passes. A Java package is not sealed, so any class in
`octometer.kit.mongo.store` can call this constructor. Add a guard at line 68:

```java
if (retentionDays < 1) {
    throw new IllegalArgumentException("retentionDays must be 1 or more");
}
```

### MINOR 3 — the erasure message does not state its consequence

`kit/jvm-mongo/src/main/java/octometer/kit/mongo/store/MongoEventLogStore.java:89-90`

The message is "The store does not implement deleteByUserId yet. Issue #35 adds it." An app
gets a loud failure, which is correct. An operator who handles an erasure request needs one
more fact. Suggested text: "The store does not delete an event by user id yet. No event was
deleted. Each event of this user stays until the TTL index removes it. Issue #35 adds the
delete."

### MINOR 4 — the store repeats no check of rule C6 on `userId`

`kit/jvm-mongo/src/main/java/octometer/kit/mongo/store/MongoEventLogStore.java:93-99`

`IngestPipeline` calls `EventFieldValidator.validateUserId` before `append`, so the documented
path is safe. A direct caller of the store is not. I wrote a user id of 5000 characters and an
empty user id, and both reached the document. Rule C6 allows 1 to 254 characters. A call of
`EventFieldValidator.validateUserId(userId)` at the start of `append` is one line, and
`kit/jvm-core` is already an `api` dependency.

### MINOR 5 — `Integer.parseInt` accepts a Unicode digit

`kit/jvm-mongo/src/main/java/octometer/kit/mongo/store/MongoEventLogStore.java:153`

`OCTOMETER_RETENTION_DAYS` with the Arabic-Indic text `١٠` gives 10 days. The value `+5` gives
5 days. Neither one shortens a retention below what the operator typed, so the risk is small. A
check with the pattern `[0-9]+` before the parse removes the surprise.

## Summary

- BLOCKER: 0
- MAJOR: 3
- MINOR: 5

Fix the three MAJOR findings, and I approve. The injection surface, the data minimisation, the
collection name, and the secret hygiene are all correct today.

xamcross:

I cloned the repository, checked out `issue-11-mongo-store` at `858ce36`, and ran each
command myself with JDK 21. I added my own tests in my clone for each claim that needed a
proof. I deleted my clone and my tests after the review; the branch keeps no file of mine.

## The acceptance criteria

|---|---|---|---|
| 1 | Exactly the fields of section 4.1, no `_class` | PASS | My own test read the raw BSON: `_id` is `OBJECT_ID`, `ts` is `DATE_TIME`, `element` and `sessionId` are `STRING`, the `userId` key is present with `NULL`. The key set is `{_id, ts, element, sessionId, userId}`. |
| 2 | A second start with a different retention changes `expireAfterSeconds` | PASS | `MongoEventLogStoreTest:110` against `mongo:7.0`. My own test also went 30 → 10 → 30 days, and the index followed each time. |
| 3 | An index error gives a warn line, and the store still works | PASS | I ran a real `mongod --auth` with an app user that holds only `find` and `insert`. The start gave exactly one `WARNING` line, and `append` still wrote 1 document. |
| 4 | The tests pass against 5.0.1 and the newest 5.x | PASS | `:kit:jvm-mongo:test` 18 tests, 0 failed, 0 skipped. `:kit:jvm-mongo:testMongoSyncNewest` 18 tests, 0 failed, 0 skipped. |
| 5 | A comment on the issue records the two outputs | PASS | The comment is on #11. |

**The tests are real.** I broke three things in my clone, one for each of the first three
criteria: I added a `_class` field, I removed the `collMod` branch of error code 85, and I
removed the warn call of the other-code branch. Five tests failed, and each failure named the
right criterion. I restored the file, and the suite went back to green.

## The other checks

- **The document (C1 to C7, C11).** Correct. The store builds a plain `Document`, so no
  `_class` field can appear. `_id` is absent from the document, thus the driver creates the
  ObjectId inside the insert call (C2).
- **C2 in full seconds.** I ran 40 rounds with `ts` from the server clock a moment before the
  insert. The time bytes of the `_id` were at or after `ts` in full seconds every time. The
  two values come from the same JVM clock, and the insert is always later, so the rule holds.
- **C6.** The code writes the key with a BSON null, and it does not drop the key. This agrees
  with `contract/examples/event-userid-null-C6.json` and with section 4.1.
- **C8, the second index.** My own test counted the indexes: exactly 2. The names are `_id_`
  and `ts_ttl`, the key is `{ts: 1}`, and `expireAfterSeconds` is 2592000 by default. The
  store adds no other index.
- **An invalid retention value.** `0`, `-5`, and `abc` each give 30 days and one warn line.
  A value above the range of a 32-bit integer gives a `NumberFormatException`, thus 30 days.
- **Two instances at the same time.** I started 6 stores in parallel threads on one database.
  No thread threw, and no warn line appeared.
- **`collMod` on Atlas M0.** Permitted. The page "Command Limits in Free Clusters"
  (`mongodb.com/docs/atlas/unsupported-commands/`, read on 2026-09-21) names neither
  `collMod` nor `createIndexes`. Only a listed command is blocked. Note the privilege: the
  built-in `readWrite` role holds `createIndex` but not `collMod`. An Atlas app user with
  "Read and write to any database" thus creates the index, and a later retention change
  fails with one warn line.
- **The write concern and the timeouts.** The store sets none. It uses the settings of the
  `MongoDatabase` of the app. This is right, and see MAJOR 2 below.
- **The two driver versions.** I printed the resolved class path of each test task:
  `test` holds `mongodb-driver-sync-5.0.1.jar`, `mongodb-driver-core-5.0.1.jar`, and
  `bson-5.0.1.jar`; `testMongoSyncNewest` holds the same three at `5.12.0`. No jar appears
  twice. The Gradle form works.
- **5.12.0 is the newest.** I read
  `repo1.maven.org/maven2/org/mongodb/mongodb-driver-sync/maven-metadata.xml` on 2026-09-21:
  `<latest>5.12.0</latest>`, `<release>5.12.0</release>`.
- **The CI job.** `tasks.check { dependsOn(testMongoSyncNewest) }` is correct. A dry run of
  `:kit:jvm-mongo:check` lists `test` and `testMongoSyncNewest`, thus `./gradlew build` of
  the job "JVM modules" runs both. The five checks of the pull request are green.
- **The published view.** `./gradlew :kit:jvm-mongo:dependencies --configuration
  runtimeClasspath` gives one line: `\--- project ':kit:jvm-core'`. No Kotlin library, no
  driver.
- **The versions on Maven Central.** Testcontainers 2.0.5, Mockito 5.23.0, and JUnit BOM
  6.1.3 are each the current release.
- **The scope.** No `path`, no `referrerHost`, no event cap, no delete. `deleteByUserId`
  throws `UnsupportedOperationException`, and the message names #35.
- **No secret.** The diff holds no credential, no database file, and no build output.
- **The language.** The pull request text, each Javadoc, each comment, and each commit body
  follow ASD-STE100. I found no defect here.

## The findings

### MAJOR 1 — the batch is not the unit of the write, and nothing says so
`kit/jvm-mongo/src/main/java/octometer/kit/mongo/store/MongoEventLogStore.java:84`

`append` sends one **ordered** `insertMany` with no transaction. A failure in the middle of
the batch leaves each event before the failure in the collection. I proved it: I added a
unique index, then sent a batch of 3 events where the second one broke the index. The
result was `e1=1` and `e3=0`, plus a `MongoBulkWriteException`.

The Javadoc of `EventLogStore.append` says "The batch is the unit of the write". The
MongoDB store does not keep that promise. The chain matters: the adapter maps the exception
to status 500 (design decision D18), the tracker retries the batch one time (design
decision D24), and the retry writes **new** `_id` values. The monitor drops a duplicate on
`(app_id, event_id)` (contract rule C24), and `event_id` is the `_id`, so it cannot drop
these. The user then sees the first part of the batch two times.

**My recommendation for an M0 cluster.** Keep the single ordered `insertMany`. Do not open a
transaction: an M0 cluster is a replica set, so a transaction works, but it adds two round
trips to a cluster with a limit of 100 operations each second, and the gain is small against
an at-least-once model. State the behaviour instead.

**The exact fix.** Add this paragraph to the class Javadoc, above the `deleteByUserId`
paragraph:

```java
 * <p>{@link #append} sends one ordered {@code insertMany} call for the
 * whole batch. It opens no transaction, because a transaction costs two
 * more round trips on an Atlas M0 cluster. A failure in the middle of the
 * batch therefore keeps each event before the failure, and it drops each
 * event after the failure. The caller gets a {@code
 * MongoBulkWriteException}. The adapter maps it to status 500, the tracker
 * retries the batch one time (design decision D24), and the retry writes
 * new {@code _id} values. The monitor keys on {@code _id} (contract rule
 * C24), thus it cannot drop the repeated events. This is the accepted cost
 * of at-least-once delivery on an M0 cluster.
```

Add the same note to the pull request text, under "Notes for the maintainer".

### MAJOR 2 — no word on the write concern and the timeouts
`kit/jvm-mongo/src/main/java/octometer/kit/mongo/store/MongoEventLogStore.java:52` and `:69`

The store sets no write concern, no read preference, and no timeout. That is the right
choice, because the app owns the `MongoDatabase` (design decision D22). The defect is the
silence: the Javadoc never states it, and it never states the risk.

The Java driver has **no socket read timeout by default**, and the default write concern of
MongoDB 5.0 and later is `w: "majority"` with no `wtimeout`. A primary that accepts the
write and never acknowledges it thus holds the calling thread for ever. Issue #12 calls
`append` inside `withContext(Dispatchers.IO)` (design decision D23). `Dispatchers.IO` has 64
threads by default. A blocked primary therefore takes all 64 threads, and every other IO
task of the app stops too, not only the ingest route. The constructor has the same problem:
`createIndex` runs at app start with no timeout.

**The exact fix.** Add this paragraph to the class Javadoc:

```java
 * <p>The store sets no write concern, no read preference, and no timeout.
 * It uses the settings of the {@link MongoDatabase} of the app. The driver
 * has no socket read timeout by default, so a blocked primary holds the
 * calling thread for ever. An app that calls {@link #append} from a thread
 * pool, for example {@code Dispatchers.IO} of the Ktor route (issue #12),
 * must set a timeout on its own {@code MongoClientSettings}: {@code
 * timeoutMS} with driver 5.2 or later, or a socket read timeout plus a
 * write concern with a {@code wtimeout} with driver 5.0.
```

### MAJOR 3 — the warn line drops the cause
`kit/jvm-mongo/src/main/java/octometer/kit/mongo/store/MongoEventLogStore.java:128-131`

`warnIndexError()` takes no argument. Three different failures give one identical text:

1. The user has no `createIndex` right (Atlas code 8000, or code 13).
2. The `collMod` call fails, for example because the user has no `collMod` right.
3. `OCTOMETER_RETENTION_DAYS` is above the server range.

I saw case 1 and case 3 on a real server, and both gave the same line. An operator reads
"The store could not set up the TTL index on ts" and learns nothing. The retention is then
off, or it keeps the old value, and nobody can say why. The catch blocks name the exception
`e` and never use it.

**The exact fix.** Give the step and the error code to the log line. Keep the server text
out, because it can hold the namespace and the user name.

```java
    private static void warnIndexError(String step, RuntimeException cause) {
        String detail = cause instanceof MongoCommandException commandException
                ? "code " + commandException.getErrorCode() + " (" + commandException.getCodeName() + ")"
                : cause.getClass().getSimpleName();
        LOGGER.log(Level.WARNING, "The TTL index step \"" + step + "\" on ts failed with "
                + detail + ". The app start continues with no TTL index change.");
    }
```

Call it with `warnIndexError("createIndex", e)` at line 110 and line 113, and with
`warnIndexError("collMod", e)` at line 124. The present tests still pass, because the text
keeps the word "index".

### MINOR findings

These do not prevent an approval.

1. `MongoEventLogStoreTest.java:120-129`. The index test is weak. It asserts only that
   `_id_` is present and that one other name exists. It passes with 5 extra indexes, with a
   wrong index name, and with a wrong key. Fix: assert `assertEquals(2, indexes.size())`,
   the name `ts_ttl`, the key `{ts: 1}`, and `expireAfterSeconds`.
2. `MongoEventLogStoreTest.java:71-79`. `assertNull(stored.get("userId"))` also passes when
   the key is absent. Fix: add `assertTrue(stored.containsKey("userId"))`.
3. `build.gradle.kts:31-32`. The Testcontainers version and the Mockito version are literal
   strings. The same file puts the two driver versions in `gradle/libs.versions.toml`. Fix:
   add `testcontainers = "2.0.5"` and `mockito = "5.23.0"` to the catalog, then read them
   with `libs.versions.testcontainers.get()`.
4. `build.gradle.kts:57-61`. The console log of PASSED and SKIPPED lines **must stay**. It
   costs 36 lines and it shows a reader that each test ran. But it guards nothing: a runner
   with no Docker makes the 7 container tests skip, and the build stays green. Fix: add a listener that
   fails the task when a test skips and the environment variable `CI` is present.
5. `MongoEventLogStore.java:148-163`. A value above 24855 days gives no TTL index at all. I
   measured the bound on `mongo:7.0`: 24855 days works, 24856 days fails, because
   `expireAfterSeconds` must fit in a 32-bit integer. The store gives one warn line and
   starts, which agrees with decision D22, but the retention is then off. Fix: clamp the
   value to 24855 days, and name the clamp in the warn line.
6. `MongoEventLogStore.java:21`. The Javadoc claims "contract rules C1 to C11". The store
   covers C1, C2, C3, C6, C7, C8, and C11. `kit/jvm-core` covers C4 and C5, and the monitor
   covers C9 and C10. Fix: name the rules that the store covers.
7. `gradle/libs.versions.toml:8`. The diff moves `typesafe-config` below `sqlite-jdbc` for
   an alphabetical order. This change is not needed for issue #11, and it raises the risk of
   a merge conflict with pull request #123. **The conflict risk with #123:** #123 adds a line
   to `settings.gradle.kts` at the same place (`include(":kit:jvm-ktor")` after
   `include(":kit:jvm-core")`), thus git reports a conflict in that file. The version catalog
   is the second risk, because #123 raises the Ktor and Kotlin versions in the same
   `[versions]` block. Fix: put `typesafe-config` back in its old place, and merge the second
   pull request with a rebase.
8. `MongoEventLogStoreIndexErrorTest.java:49-55`. The test reads the `collMod` field of the
   command and stops. It does not check the `index` sub-document. Fix: assert the
   `keyPattern` value and the `expireAfterSeconds` value.
9. `build.gradle.kts:68-90`. `configurations.create` is the legacy API. The `jvm-test-suite`
   plugin, or `configurations.resolvable(...)`, states the intent better and survives a
   Gradle upgrade. The present form works; I verified the resolved class path.
10. `MongoEventLogStore.java:88-91`. `deleteByUserId` throws
    `UnsupportedOperationException` for every value. The interface asks for a
    `NullPointerException` on `null` and an `IllegalArgumentException` on an empty text.
    This is correct until #35 lands. Add a `@Override` Javadoc line that says so.
11. No test builds the store through the public constructor with
    `OCTOMETER_RETENTION_DAYS` set. Only `retentionDaysFromValue` has a test. The gap is
    small, because a test cannot set the process environment of the JVM.

Count: 3 MAJOR, 11 MINOR, 0 BLOCKER.

xamcross:
## Response to the two first reviews

This correction round is on the same branch, `issue-11-mongo-store`, commit
`07875a6`. Each acceptance criterion still passes:
`./gradlew :kit:jvm-mongo:test` (25 tests, 0 failed, 0 skipped),
`./gradlew :kit:jvm-mongo:testMongoSyncNewest` (25 tests, 0 failed, 0
skipped), and `./gradlew :kit:jvm-core:test` (81 tests, 0 failed, 0
skipped). The CI job "JVM modules" also shows a clean `BUILD SUCCESSFUL`,
with `CI=true` on that runner, so the new skip guard confirms that no
container test skipped there.

### Java and MongoDB engineer

**MAJOR 1 — the batch is not the unit of the write.** Fixed by decision,
not by a code change. `MongoEventLogStore.java:41-53` (class Javadoc)
states the ordered `insertMany` with no transaction, the reason (the
constructor takes only a `MongoDatabase`, and a transaction needs a
client session and two more round trips on an M0 cluster), and the
at-least-once consequence. `EventLogStore.java:13-25` in `kit/jvm-core`
now states the same promise for each store, in general terms: a store
writes the batch as one unit when it can, or it writes the first part and
then throws. `InMemoryEventLogStore` is unchanged.

**MAJOR 2 — no word on the write concern and the timeouts.**
`MongoEventLogStore.java:55-63` (class Javadoc) states that the store
sets no write concern, no read preference, and no timeout, and names the
fix for an app: `timeoutMS` on driver 5.2 or newer, or a socket read
timeout plus `wtimeout` on driver 5.0.

**MAJOR 3 — the warn line drops the cause.** `MongoEventLogStore.java:193-205`
(`warnIndexError`) now takes the step name and the exception, names the
step (`createIndex` or `collMod`) and the numeric error code in the log
line, and never the server text. A failed `createIndex` states plainly
that the collection now has no TTL index, and that events stay with no
time limit (contract rule C8 does not apply). A failed `collMod` states
that the old retention value stays. Test:
`MongoEventLogStoreIndexErrorTest.aDifferentErrorCodeNamesTheStepAndTheErrorCodeAndSaysNoTtlIndexRemains`
and `.collModFailureNamesTheStepAndSaysTheOldRetentionStays`. A real
database user with only `insert`:
`MongoEventLogStoreInsertOnlyUserTest.aUserWithOnlyInsertGetsOneWarningNoTtlIndexAndAWorkingAppend`.

**MINOR 1 (index test weak).** Fixed.
`MongoEventLogStoreTest.theCollectionKeepsExactlyTheDefaultIdIndexAndTheTtlIndexOnTs`
now asserts exactly 2 indexes, the name `ts_ttl`, the key `{ts: 1}`, and
`expireAfterSeconds`.

**MINOR 2 (`assertNull` also passes on an absent key).** Fixed.
`MongoEventLogStoreTest.appendAcceptsANullUserIdForAnAnonymousClick` now
also asserts `containsKey("userId")`.

**MINOR 3 (Testcontainers/Mockito as literal strings).** Fixed.
`gradle/libs.versions.toml:20-22` adds `junit`, `mockito`, and
`testcontainers`. `kit/jvm-mongo/build.gradle.kts:42-44` reads them with
`libs.versions.<name>.get()`.

**MINOR 4 (keep the PASSED/SKIPPED log, but guard the skip).** The
maintainer's decision changes this: the guard replaces the log, so the
log is removed. `kit/jvm-mongo/build.gradle.kts:66-91` adds a
`TestListener` that fails the task when a test skips and `CI` holds
`true`. Verified on the CI job "JVM modules" (clean build, `CI=true`,
guard did not fire) and locally (the guard did not fire with Docker
present; I could not force a real Docker-less skip locally, because
Docker Desktop here is shared with other agents).

**MINOR 5 (a retention above 24855 days gives no TTL index).** Fixed.
`MongoEventLogStore.java:245-253` (`clampRetentionDays`) clamps the value
to 24855 days (the largest value that fits the 32-bit
`expireAfterSeconds` field) and logs a warning naming both the requested
value and the clamp. Test:
`MongoEventLogStoreIndexErrorTest.aRetentionAboveTheMaximumIsClampedWithAWarning`.

**MINOR 6 (the Javadoc claims C1 to C11).** Fixed.
`MongoEventLogStore.java:22-26` now names the exact rules the store
covers (C1, C2, C3, C6, C7, C8, C11) and states that C4/C5 are the
concern of `kit/jvm-core` and C9/C10 the concern of the monitor.

**MINOR 7 (the `typesafe-config` move).** Reverted.
`gradle/libs.versions.toml:8` puts `typesafe-config` back after `logback`
and before the `mongodb-*` entries, its original place, to keep the merge
risk with pull request #123 low.

**MINOR 8 (the `collMod` test does not check the index sub-document).**
Fixed. `MongoEventLogStoreIndexErrorTest.errorCodeEightyFiveRunsCollModWithNoWarning`
now asserts `keyPattern` and `expireAfterSeconds` of the `index`
sub-document.

**MINOR 9 (`configurations.create` is legacy).** Not changed. The
maintainer accepts the present form; a convention-plugin cleanup is a
larger, separate change.

**MINOR 10 (`deleteByUserId` needs a Javadoc `@Override` note).** Not
added as a separate Javadoc line; the exception message itself now states
the temporary nature and the consequence in full (see MINOR item below,
security review MINOR 3). `MongoEventLogStoreIndexErrorTest.deleteByUserIdThrowsAnUnsupportedOperationExceptionAndDeletesNoEvent`
covers the new message text.

**MINOR 11 (no test with `OCTOMETER_RETENTION_DAYS` set through the
public constructor).** Not added, for the reason the finding itself
gives: a test cannot set the process environment of the JVM. Rejected,
with the finding's own evidence.

### Security and privacy engineer

**MAJOR 1 — the write exception carries a host, a port, and a database
name.** Fixed. `MongoEventLogStore.java:127-133` (`append`) now catches
the driver exception and throws `EventLogWriteException`
(`MongoEventLogStore.java:262-276`), an unchecked kit exception with a
fixed message and the numeric MongoDB error code, and no cause. Test:
`MongoEventLogStoreWriteErrorTest`, with a mocked driver exception that
carries a real-looking host, port, and database name; the test reads the
exception message and the full stack trace text and finds none of the
three.

**MAJOR 2 — the TTL warning hides its own consequence.** Fixed, together
with MAJOR 3 of the other review (same line): see above.

**MAJOR 3 — three versions sit outside the version catalog.** Fixed: see
MINOR 3 of the other review, above.

**MINOR 1 (pin the test image by digest).** Not changed. `mongo:7.0` is
kept as a tag, matching the present house style of this module; a
digest pin is a larger, repository-wide change that the maintainer did
not ask for in this round.

**MINOR 2 (the package-private constructor accepts 0 days).** Fixed.
`MongoEventLogStore.java:107-111` throws `IllegalArgumentException` for a
value below 1. Test:
`MongoEventLogStoreIndexErrorTest.aRetentionBelowOneDayIsRejected`.

**MINOR 3 (the erasure message needs its consequence).** Fixed.
`MongoEventLogStore.java:138-141` (`deleteByUserId`) now states plainly
that no event was deleted, and that each event stays until the TTL index
removes it.

**MINOR 4 (no repeated `userId` length check in the store).** Not added.
The maintainer's decision list does not include this finding, and the
documented call path already validates through `EventFieldValidator`
before `append` runs.

**MINOR 5 (`Integer.parseInt` accepts a Unicode digit).** Fixed.
`MongoEventLogStore.java:229` (`retentionDaysFromValue`) now requires an
ASCII-only digit pattern before the parse. Test:
`MongoEventLogStoreRetentionDaysTest.aUnicodeDigitGivesTheDefaultAndAWarning`
and `.aLeadingPlusSignGivesTheDefaultAndAWarning`.

### Rejected findings

MINOR 9 (Java review, `configurations.create`), MINOR 1 (security review,
image digest pin), and MINOR 4 (security review, repeated `userId`
check) are not in the maintainer's list of decisions for this round, and
each one is a larger or separate change. No new evidence changes that.

MINOR 10 and MINOR 11 of the Java review are addressed by a related fix
(see above) or accepted as-is with the finding's own reasoning.

xamcross:
**Review (second review, security and privacy engineer): APPROVE**

I made a new clone, and I checked out commit `07875a6`. I read the code first, and I judged each
finding against the code, not against the response comment. I then wrote 34 probes of my own
against a real MongoDB 7.0 server, and I removed them before the final run. Each of my three
MAJOR findings is corrected, and the correction holds under a real server. Two new MINOR
findings stay open. A MINOR finding does not prevent an approve.

My test runs, with JDK 21 at `C:/Users/xamcr/.jdks/ms-21.0.9`:

- `./gradlew :kit:jvm-mongo:test :kit:jvm-mongo:testMongoSyncNewest --rerun-tasks` — BUILD
  SUCCESSFUL, 50 tests in total (25 for each driver version), 0 failed, 0 skipped. I counted
  them in the XML report, because the per-test log line is gone now.
- My 34 probes — 34 passed, 0 failed, 0 skipped.

## The first findings

| Finding | Result |
|---|---|
| MAJOR 1 — the write exception carries a host, a port, and a database name | **CORRECTED** |
| MAJOR 2 — the TTL warning hides its own consequence | **CORRECTED** |
| MAJOR 3 — three versions sit outside the version catalog | **CORRECTED** |
| MINOR 1 — pin the test image by digest | NOT CORRECTED, reason accepted |
| MINOR 2 — the package-private constructor accepts 0 days | **CORRECTED** |
| MINOR 3 — the erasure message needs its consequence | **CORRECTED** |
| MINOR 4 — no repeated `userId` check in the store | NOT CORRECTED, reason accepted |
| MINOR 5 — `Integer.parseInt` accepts a Unicode digit | **CORRECTED** |

## MAJOR 1 — CORRECTED

`MongoEventLogStore.java:127-133` wraps `insertMany`. `MongoEventLogStore.java:262-276` defines
`EventLogWriteException`. The constructor calls `super(message)` only, so the exception holds no
cause.

I built the strongest real case that I could. I put a `$jsonSchema` validator on
`octometer_events`, so the server rejects the document. My probe first read the raw driver
exception of the same insert:

- The raw driver text holds the user id: **true**.
- The raw driver text holds the host and the port: **true**.

The kit exception of the same failure holds none of them. I searched the message, the
`toString()`, the cause, each suppressed exception, and the full stack trace text for: the host,
the port, `27017`, the database name, the user id, the element, the session id, the name space
`octometer_events`, the text "not authorized", the text "failed validation", a `Caused by` line,
and a `com.mongodb` frame. Each search was empty. The whole text is:

```
The store could not write the event batch to the event collection. The MongoDB error code is -1.
```

**Each other path.** The class calls the driver at four places only: line 112
(`getCollection`, a local call with no network work), line 128 (`insertMany`), line 156
(`createIndex`), and line 174 (`runCommand`). The last three sit inside a `try` block that
catches `RuntimeException`, so no driver exception leaves the store. I proved the two open
cases against a real server:

- **An unreachable server.** The constructor does not throw. It logs exactly one clean warning.
  `append` throws the kit exception with the error code -1, and the text is clean.
- **A closed client.** The same result: one clean warning, then a clean kit exception.

## MAJOR 2 — CORRECTED

`MongoEventLogStore.java:193-205`. The method takes the step name and the exception. It reads
`getErrorCode()` from a `MongoCommandException`, and it writes `unknown` for each other
exception. It never reads `getMessage()` of the cause, so the server text stays out. The line
for a failed `createIndex` states the plain result:

```
The TTL index step "createIndex" failed with the MongoDB error code unknown. The app start
continues. The event collection now has no TTL index. Each event of this app stays with no
time limit. Contract rule C8 does not apply until an administrator creates the index.
```

The line for a failed `collMod` states that the old retention value stays, and that rule C8
still applies with the old value. Both texts hold no host, no port, no database name, and no
name space. I confirmed this with three real failures: an unreachable server, a closed client,
and a user with no `createIndex` right.

**The new test.** `MongoEventLogStoreInsertOnlyUserTest` is a real test. It starts `mongo:7.0`
with `--auth`, it creates a custom role with the single action `insert` on `octometer_events`,
and it builds the store as that user. It asserts one WARNING line, the step name, the text
"no TTL index", the absence of the server text "not authorized", the absence of a TTL index in
the collection, and a working `append`. This matches my own measurement of the first round.

## MAJOR 3 — CORRECTED

`gradle/libs.versions.toml:19-22` adds `junit = "6.1.3"`, `mockito = "5.23.0"`, and
`testcontainers = "2.0.5"`, under a comment that names the confirm date.
`kit/jvm-mongo/build.gradle.kts:39-44` reads each one with `libs.versions.<name>.get()`. No
version literal stays in the module.

## The MINOR findings of the first round

**MINOR 1 (pin the test image by digest) — NOT CORRECTED. I accept the reason.** A digest pin
is a repository-wide convention, and the maintainer did not ask for it in this round. The tag
`mongo:7.0` is fixed, and it is not `latest`, so the bar of the issue holds. Please open a
separate issue for the whole repository.

**MINOR 2 — CORRECTED.** `MongoEventLogStore.java:109-111` rejects a value below 1. My probe
called the constructor with 0, -1, and `Integer.MIN_VALUE`. Each call threw
`IllegalArgumentException`, and the collection got no index at all. The dangerous state of the
first round, a TTL index with `expireAfterSeconds: 0`, is now unreachable.

**MINOR 3 — CORRECTED.** `MongoEventLogStore.java:138-141`. My probe read the message:

```
The store does not delete an event by user id yet. No event was deleted. Each event of this
user stays until the TTL index removes it. Issue #35 adds the delete.
```

It states the result, it names the time limit, and it holds no user id.

**MINOR 4 (no repeated `userId` check) — NOT CORRECTED. I accept the reason for this round.** I
read `IngestPipeline.java:96-106` again: it calls `EventFieldValidator.validateUserId` before
`append`, so the documented path holds rule C6. `MongoEventLogStore` stays public, so a direct
caller still writes a user id of any length. Please carry this to the issue that hardens the
kit API.

**MINOR 5 — CORRECTED.** `MongoEventLogStore.java:87` and `:229` demand the pattern `[0-9]+`
before the parse. My probe: the Arabic-Indic texts now give 30 days plus one warning, and `+5`
also gives 30 days plus one warning.

## The retention parser (point 5)

I ran my 22 hostile texts again. Each wrong text gives 30 days and exactly one warning: `0`,
`-1`, `-2147483648`, `abc`, an empty text, a text of spaces, `1e3`, `0.5`, `0x10`,
`2147483648`, `9999999999999999`, `30d`, `NaN`, `Infinity`, `null`, `1 second`, `+0`, `-0`,
`1,5`, and the two Arabic-Indic texts. No wrong text gives a life below 30 days. The text `01`
gives 1 day with no warning; this is correct, because an operator wrote a positive ASCII
number.

The clamp is an improvement. `Integer.MAX_VALUE` days now gives `expireAfterSeconds`
2147472000, which is 24855 days, plus one warning. In the first round the same value left the
collection with no TTL index. I also confirmed the bound: 24855 days fits the 32-bit field of
the server, and 24856 days does not. The clamp only lowers a value above 68 years, so it never
shortens a real retention.

## The injection probes (point 6)

Each injection probe of my first review still passes. I fed 12 hostile texts into `userId`,
`element`, and `sessionId`: `$gt`, `$where`, `$$ROOT`, `$expr`, `{"$ne": null}`,
`{"$set": {"userId": "admin"}}`, `{"_class": "octometer.Evil"}`, `a.b.c`, `..`, a text with a
NUL character, and two more. Each one reached the server as a BSON STRING with the exact input
text. The key set stayed `{_id, ts, element, sessionId, userId}`. No key started with a dollar
sign, and no key held a dot. An exact string match found each document again.

## New findings of this round

### MINOR 6 — the new test builds a text with the form of a credential

`kit/jvm-mongo/src/test/java/octometer/kit/mongo/store/MongoEventLogStoreInsertOnlyUserTest.java:37`,
`:40`, and `:130-133`

The file holds two password literals, and `connectionString` joins them into the form
`mongodb://<user>:<password>@<host>:<port>/`. The CI job `Secret scan` passes today, and I
confirmed that with `gh pr checks 126`. The scan passes for two reasons only: the code joins
the parts at run time, so the raw file text keeps a quotation mark straight after the two
slashes, and the entropy of each fake password stays under the gitleaks bound. Both reasons are
one edit away from a finding. The comment in `.gitleaks.toml:34-38` states that a fake
credential for a test belongs in a file that the allowlist names, and this file is not in the
allowlist.

**The fix.** Build the client with a credential object, so no text of that form exists:

```java
private static MongoClient clientFor(String user, String password, String authDatabase) {
    MongoClientSettings settings = MongoClientSettings.builder()
            .applyToClusterSettings(builder -> builder.hosts(
                    List.of(new ServerAddress(MONGO.getHost(), MONGO.getMappedPort(27017)))))
            .credential(MongoCredential.createCredential(user, authDatabase, password.toCharArray()))
            .build();
    return MongoClients.create(settings);
}
```

I used this exact form in my own probe against the same driver version, and it works. Delete
the method `connectionString` after the change.

### MINOR 7 — a real write error loses its numeric code

`kit/jvm-mongo/src/main/java/octometer/kit/mongo/store/MongoEventLogStore.java:129-132`

`MongoBulkWriteException` is a sister class of `MongoCommandException`, not a child, so an
ordered `insertMany` that the server rejects falls to the second `catch` block and gives the
error code -1. My probe with a real `$jsonSchema` validator confirmed the code -1, while the
server sent 121. The privacy result is correct, because the text of a bulk write exception can
hold the document values. The operator, however, loses the one number that the message
promises. Read the code from the first write error:

```java
} catch (MongoBulkWriteException e) {
    List<BulkWriteError> writeErrors = e.getWriteErrors();
    throw new EventLogWriteException(writeErrors.isEmpty() ? -1 : writeErrors.get(0).getCode());
} catch (MongoCommandException e) {
    throw new EventLogWriteException(e.getErrorCode());
} catch (RuntimeException e) {
    throw new EventLogWriteException(-1);
}
```

Keep the new `catch` block above the `RuntimeException` block. The message text needs no change.

### MINOR 8 — the `createIndex` consequence text over-claims

`kit/jvm-mongo/src/main/java/octometer/kit/mongo/store/MongoEventLogStore.java:198`

The text says "The event collection now has no TTL index." That is right for a first start. It
is wrong for one case: a collection that already holds the TTL index from an earlier start, and
an app user that loses the `createIndex` right later. The server checks the right before the
index state, so `createIndex` fails with code 13, and the old TTL index stays. The line then
states a worse state than the real one. The error is on the safe side, so the severity is
MINOR. Change the word "now has" to "can now have", and keep the rest. The assertion of
`MongoEventLogStoreInsertOnlyUserTest:86` reads the text "no TTL index", so it still passes.

## Summary

- BLOCKER: 0
- MAJOR: 0
- MINOR: 3 new (MINOR 6, MINOR 7, MINOR 8), plus 2 open from the first round with an accepted
  reason (the image digest, the repeated `userId` check)

The three MAJOR findings are corrected, and each correction holds against a real server. I
approve. Please carry the five MINOR findings to a separate issue.

xamcross:
**Review (second review, Java and MongoDB engineer): CHANGES REQUESTED**

I made a new clone, checked out `issue-11-mongo-store` at `07875a6`, and ran each command
with JDK 21 and Docker. I checked each finding against the code, not against the response
comment.

`:kit:jvm-mongo:test` 25 tests, `:kit:jvm-mongo:testMongoSyncNewest` 25 tests, and
`:kit:jvm-core:test` 81 tests. Each one: 0 failed, 0 skipped. The 5 checks of the pull
request are green.

One MAJOR finding is new, and it comes from the correction itself. Each first finding is
corrected.

## The first findings

|---|---|---|
| MAJOR 1, the batch is not the unit of the write | **CORRECTED** | See below. |
| MAJOR 2, no word on the write concern and the timeouts | **CORRECTED** | See below. |
| MAJOR 3, the warn line drops the cause | **CORRECTED** | See below. |
| MINOR 1, the weak index test | CORRECTED | `MongoEventLogStoreTest.java:122-141` asserts 2 indexes, the name `ts_ttl`, the key `{ts: 1}`, and `expireAfterSeconds`. |
| MINOR 2, `assertNull` also passes on an absent key | CORRECTED | `MongoEventLogStoreTest.java:79` adds `assertTrue(stored.containsKey("userId"))`. |
| MINOR 3, versions outside the catalog | CORRECTED | `gradle/libs.versions.toml:18-22` holds `junit`, `mockito`, and `testcontainers`. |
| MINOR 4, the console log | CHANGED, and I accept it | See point 5 below. |
| MINOR 5, a retention above the server range | CORRECTED | `MongoEventLogStore.java:245-253` clamps at 24855 days. My first-round measurement on `mongo:7.0` gives the same bound: 24855 days works, 24856 days fails. |
| MINOR 6, the C1 to C11 claim | CORRECTED | `MongoEventLogStore.java:22-26`. |
| MINOR 7, the `typesafe-config` move | CORRECTED | See point 6 below. |
| MINOR 8, the `collMod` sub-document | CORRECTED | `MongoEventLogStoreIndexErrorTest.java:56-58`. |
| MINOR 9, 10, 11 | Rejected; I accept each reason. See point 7 below. |

### 1. MAJOR 1 — the two Javadoc texts

I ran my partial-write probe again against `mongo:7.0`: a unique index, then a batch of 3
events with a duplicate in the middle. The result is the same as in the first round:
`e1=1` and `e3=0`. The behaviour did not change, and the text now states it.

The two texts agree with each other and with the in-memory store:

- `MongoEventLogStore.java:41-53` states the ordered `insertMany`, the reason for no
  transaction, the partial write, the D24 retry, and the C24 consequence.
- `EventLogStore.java:16-25` states the general rule: a store writes the batch as one unit
  when it can; a store that cannot do that writes the first part and then throws.
- `InMemoryEventLogStore.append` builds the whole batch first, then calls
  `ConcurrentLinkedQueue.addAll`, so it still writes one unit. It is the "when it can" case,
  and its own comment says so. No text contradicts another.

### 2. MAJOR 2 — the option names

I checked each name against the driver jars, not against a page:

| Name | driver-core 5.0.1 | driver-core 5.12.0 |
|---|---|---|
| `MongoClientSettings.Builder.timeout(long, TimeUnit)` (`timeoutMS`) | absent | present |
| `SocketSettings.Builder.readTimeout(long, TimeUnit)` | present | present |
| `WriteConcern.withWTimeout(long, TimeUnit)` | present | present |

The Javadoc claim holds: `timeoutMS` needs a newer driver, and the 5.0 path (a socket read
timeout plus a write concern with a wait timeout) works on the compile version.

### 3. MAJOR 3 — the warn line, the three cases

- **A missing right.** `MongoEventLogStoreInsertOnlyUserTest` starts a real `mongod --auth`
  and builds an app user with only the `insert` action. It passed in my run. The line names
  the step `"createIndex"`, the numeric code, and "no TTL index".
- **A failed `collMod`.** `MongoEventLogStoreIndexErrorTest.collModFailureNamesTheStepAndSaysTheOldRetentionStays`
  asserts the step `"collMod"` and the text "old retention value", and it asserts that the
  server text never appears.
- **A retention above the server range.** This case cannot happen any more: the clamp at
  24855 days runs before `createIndex`.

The line now states the plain result of each case, which was the point of the finding.

## The new MAJOR finding

### MAJOR 4 — the error code is -1 for every ordinary write failure
`kit/jvm-mongo/src/main/java/octometer/kit/mongo/store/MongoEventLogStore.java:127-133`

`append` catches `MongoCommandException`, then `RuntimeException`. The MongoDB Java driver
does **not** throw `MongoCommandException` for a write error of `insertMany`. It throws
`com.mongodb.MongoBulkWriteException`, and that class extends `MongoServerException`, not
`MongoCommandException`. I confirmed the hierarchy with `javap` on `mongodb-driver-core`
5.0.1 and 5.12.0, and I confirmed the behaviour on a real server.

My probe on `mongo:7.0`, with a duplicate key in the middle of a batch:

```
REVIEW2 driver exception class=com.mongodb.MongoBulkWriteException
REVIEW2 is MongoCommandException=false
REVIEW2 is MongoBulkWriteException=true
REVIEW2 real duplicate-key errorCode=-1
REVIEW2 real duplicate-key cause=null
```

The app developer therefore gets a fixed message, the code -1, and no cause. That is less
than the round promised. The driver holds the real code: the same failure gives

```
REVIEW2 duplicate key -> writeErrors[0].code=11000
```

`MongoEventLogStoreWriteErrorTest.aCommandFailureThrowsTheKitExceptionWithNoConnectionDetailAndNoCause`
mocks `insertMany` to throw a `MongoCommandException`. That path is real for a
command-level failure, for example an unauthorized insert, but it is not the common path.
No test covers a real `MongoBulkWriteException`, so the suite hides the defect.

**The exact fix.** Add one catch block before the `MongoCommandException` block, and one
helper. Both methods exist in driver-core 5.0.1, so the compile version is safe. Each value
is a number, so design decision D15 stays satisfied.

```java
        try {
            collection.insertMany(documents, new InsertManyOptions().ordered(true));
        } catch (MongoBulkWriteException e) {
            throw new EventLogWriteException(firstErrorCode(e));
        } catch (MongoCommandException e) {
            throw new EventLogWriteException(e.getErrorCode());
        } catch (RuntimeException e) {
            throw new EventLogWriteException(-1);
        }
```

```java
    /**
     * The numeric code of the first write error of a bulk write, or the
     * code of the write concern error, or -1. The method reads only the
     * numeric code, never the server text (design decision D15).
     */
    private static int firstErrorCode(MongoBulkWriteException exception) {
        if (!exception.getWriteErrors().isEmpty()) {
            return exception.getWriteErrors().get(0).getCode();
        }
        return exception.getWriteConcernError() == null
                ? -1
                : exception.getWriteConcernError().getCode();
    }
```

Add the import `com.mongodb.MongoBulkWriteException`. Add one container test that inserts a
duplicate key and asserts `errorCode() == 11000`.

## 4. The exception with no cause — the trade

I accept the trade. The driver text holds the host, the port, and the database name, and an
app can log an exception text into a place that the owner does not control. A numeric code
gives the same diagnostic power with none of that risk, **when the code is real**. With the
present code the trade fails, because the common case reports -1 (MAJOR 4). With the fix it
holds.

Two examples of a code and its meaning:

- **11000, `DuplicateKey`.** A unique index rejected the document. I measured this code on
  `mongo:7.0`. An operator then knows to look at the indexes of `octometer_events`, not at
  the network.
- **13, `Unauthorized`.** The database user has no `insert` right on `octometer_events`. An
  operator then knows to look at the Atlas role, not at the cluster load. This case comes
  back as a command-level error, so the present code already reports 13.

A third code is useful to know: **8000, `AtlasError`**, which an M0 cluster returns when it
refuses a write, for example over the 512 MB quota (design section 2.4).

A client-side failure has no server code. A document above 16 MB gives
`BsonMaximumSizeExceededException`, and -1 is then the right value. I measured that too.

## 5. The CI guard — proven

I proved the guard in my clone. I could not make Docker absent for the test process
(Testcontainers 2.0 falls back to the named pipe on Windows, so a wrong `DOCKER_HOST` does
not help). I used the same JUnit mechanism instead: a class-level skip. The extension
`@Testcontainers(disabledWithoutDocker = true)` and `@Disabled` both use a JUnit
`ExecutionCondition` on the container, so Gradle counts each test of the class as skipped in
the same way.

- **`CI=true`, a class with 2 tests skipped:** the task fails.
  `The task "Gradle Test Run :kit:jvm-mongo:test" of kit:jvm-mongo skipped 2 test(s) on CI.`
- **No `CI` variable, the same skip:** `BUILD SUCCESSFUL`, `skipped total = 1`. A local run
  with no Docker still skips with no failure.
- **`CI=true` with a daemon that started without `CI`:** the task still fails, so the daemon
  reads the environment of each build. The guard does not depend on a fresh daemon.

The guard is stronger than the console log that it replaces, because a machine checks it. I
therefore accept the maintainer's decision on MINOR 4. One loss stays: the CI log no longer
holds one line for each test, and the evidence comment on issue #11 cites those lines. A
later reader cannot repeat that check from the log alone. This is a MINOR point only.

## 6. The version catalog and the merge with #123

`gradle/libs.versions.toml` holds `junit = "6.1.3"`, `mockito = "5.23.0"`, and
`testcontainers = "2.0.5"`, and `typesafe-config` is back after `logback`. Each version is
still the current release on Maven Central.

The test merge:

```
git merge --no-commit --no-ff origin/issue-12-ktor-ingest-route
Auto-merging settings.gradle.kts
CONFLICT (content): Merge conflict in settings.gradle.kts
```

**One conflict only**, and it is trivial: this branch adds `include(":kit:jvm-mongo")` and
#123 adds `include(":kit:jvm-ktor")` on the same line. The person who merges second keeps
both lines. `gradle/libs.versions.toml` merged with no conflict, so the revert of MINOR 7
did its work. The other files of #123 (`build.gradle.kts` at the root, `kit/jvm-ktor/*`, the
design document) merged with no conflict. I ran `git merge --abort`, and the head is
`07875a6` again.

## 7. The four rejected findings

1. **`configurations.create` (my MINOR 9). I accept the rejection.** I printed the resolved
   class path again in this round; the two test tasks still carry two driver sets. The form
   works, and a convention plugin is a separate change.
2. **No test with `OCTOMETER_RETENTION_DAYS` set (my MINOR 11). I accept the rejection.**
   The reason is my own. The untested part is one line, `System.getenv(...)`, and
   `retentionDaysFromValue` holds each rule.
3. **The image digest pin (security MINOR 1). I accept the rejection for this round.**
   `mongo:7.0` is a moving tag, so a future image can change the test result with no change
   in the repository. This is a test-only image, and a digest pin touches each module. A
   separate issue is the right place.
4. **The repeated `userId` check (security MINOR 4). I accept the rejection.** I read the
   call path: `IngestPipeline.java:98` calls `EventFieldValidator.validateUserId(userId)`
   before `append`, and that method holds the 1 to 254 rule of C6. A second check in the
   store would repeat the rule in two places. One residual risk stays, and it is small: an
   app that calls `MongoEventLogStore.append` on its own can write a user id that breaks C6.
   The class Javadoc already says that C4 and C5 are the concern of `kit/jvm-core`; a reader
   can take C6 the same way.

## Summary

MAJOR 4 is the one finding that prevents my approval. The fix is one catch block, one helper
method, one import, and one test. Each other point of my first review is corrected, and I
verified each one against the code.

xamcross:
## Response to the second reviews (commit `5fe8777`)

- **NEW MAJOR 4 (Java) / MINOR 7 (security) — a real write error gave code -1.**
  Fixed. `MongoEventLogStore.java:128-137` (`append`) now catches
  `MongoBulkWriteException` before `MongoCommandException`, and a new
  helper `firstErrorCode` (`MongoEventLogStore.java:146-153`) reads only
  the numeric code of the first write error, or of the write concern
  error, never a message or a document. New test:
  `MongoEventLogStoreTest.appendWrapsARealDuplicateKeyFailureWithTheRealCodeAndNoConnectionDetail`,
  a real duplicate key against a live server, asserts `errorCode() == 11000`
  and no host, no port, and no database name in the message or the full
  stack trace. It ran against both driver versions
  (`:kit:jvm-mongo:test` and `:kit:jvm-mongo:testMongoSyncNewest`, 26
  tests each, 0 failed, 0 skipped).

- **MINOR (security) — a credential-shaped text in the test file.** Fixed.
  `MongoEventLogStoreInsertOnlyUserTest.java` no longer builds a
  `mongodb://user:password@host` text; `clientFor` (line 138) builds the
  client with a `MongoCredential` and a `ServerAddress` instead. The
  `Secret scan` check stays green.

- **MINOR (security) — "now has no TTL index" over-claims.** Fixed.
  `MongoEventLogStore.java:217` (and the matching Javadoc above
  `warnIndexError`) now says the collection "can now have" no TTL index,
  since an index from an earlier start can still exist.

All 5 checks are green, including `Secret scan` and `JVM modules`. Not
merged.

xamcross:
**Review (third review, Java and MongoDB engineer): APPROVE**

I made a new clone at `5fe8777` and checked each point against the code, with JDK 21 and
Docker. `:kit:jvm-mongo:test` and `:kit:jvm-mongo:testMongoSyncNewest`: 26 tests each, 0
failed, 0 skipped. The 5 checks of the pull request are green.

**1. MAJOR 4 — CORRECTED.** `MongoEventLogStore.java:130-131` catches
`MongoBulkWriteException` before the command catch. `firstErrorCode`
(`MongoEventLogStore.java:139-155`) reads only `getWriteErrors().get(0).getCode()`, or
`getWriteConcernError().getCode()`, or -1. It reads no message, no name space, and no
document. `MongoEventLogStoreTest.appendWrapsARealDuplicateKeyFailureWithTheRealCodeAndNoConnectionDetail`
asserts the code 11000 against a real server, with no cause, no host, no port, no database
name, and no field value. The test runs in both test tasks, because `testMongoSyncNewest`
reuses the same test classes, so both driver versions cover it.

**The test can fail.** I removed the new catch block in my clone. Exactly one test failed:
`appendWrapsARealDuplicateKeyFailureWithTheRealCodeAndNoConnectionDetail`, at
`MongoEventLogStoreTest.java:126`, the `assertEquals(11000, ...)` line. I restored the file,
and both tasks went back to green.

**2. Each other driver exception.** I read the class hierarchy of `mongodb-driver-core`
5.0.1 with `javap`. Every path gives `EventLogWriteException` with the fixed message; no raw
driver exception escapes.

| The failure | The catch branch | The code |
|---|---|---|
| A write error, for example a duplicate key | `MongoBulkWriteException` | the real code, 11000 measured |
| A write concern error of a bulk write | `MongoBulkWriteException` | the write concern code, for example 64 |
| Unauthorized, or an Atlas refusal | `MongoCommandException` | 13, or 8000 |
| A not-primary error | `MongoCommandException` (`MongoNotPrimaryException` is a subclass) | 10107, or 13435 |
| A node that recovers | `MongoCommandException` (`MongoNodeIsRecoveringException` is a subclass) | 11600, or 13436 |
| A server selection timeout | `RuntimeException` (`MongoTimeoutException` is a `MongoClientException`) | -1 |
| A socket error | `RuntimeException` (`MongoSocketException` is a `MongoException`) | -1 |
| An authentication failure at connect | `RuntimeException` (`MongoSecurityException`) | -1 |
| A closed client | `RuntimeException` (`IllegalStateException`) | -1 |
| A document above 16 MB | `RuntimeException` (`BsonMaximumSizeExceededException`) | -1 |

The value -1 is right in each of those last five rows: the server gives no code there,
because the write never reached the server.

**3. No credential text.** `MongoEventLogStoreInsertOnlyUserTest` now builds a
`MongoCredential`, so no text of the form `user:password@host` exists in the file. The rule
`mongodb-uri-password` of `.gitleaks.toml` has nothing to match. The job "Secret scan" is
green on this head.

**4. The warning text.** `MongoEventLogStore.java:217` says "The event collection can now
have no TTL index", and the method Javadoc at line 205 agrees. The softer wording is also
more correct: a failed `createIndex` on a collection that already holds the index leaves the
index in place.

**MINOR, no change needed.** `MongoEventLogStoreTest.java:128` asserts that the text holds
no port. The mapped port is a random number, and the message holds the number 11000. A
collision would fail the test with no defect. The Docker port range on this machine starts
far above 11000, so the risk is small. A later change can compare the port only against the
message, not against the stack trace.

Each finding of my two first reviews is now corrected. I approve.

Issue #33, opened 2026-09-21
Limit the ingest rate in kit/jvm-core


xamcross:
**Goal.** The ingest logic rejects a client that sends too many requests in 60 seconds.

**Why it blocks production.** Inside the batch limit one client sends 6 000 events each minute and fills a 512 MB cluster in hours.

**Implementation steps.**
1. In `kit/jvm-core`, add the in-memory limiter of D20 with an injected clock.
2. Return a rate-limit result from the ingest logic.
3. In `kit/jvm-ktor`, give the client address to the ingest logic: the header that `OCTOMETER_CLIENT_IP_HEADER` names, or the remote address without that variable.
4. Map the rate-limit result to the response 429 in the route of #12.

**Acceptance criteria.**
- [x] Fixed clock: request 31 of one user id in 60 seconds gives the rate-limit result, and a request after the window passes.
- [x] Without a user id, request 121 of one IP address gives the rate-limit result.
- [x] Without `OCTOMETER_CLIENT_IP_HEADER` the key is the remote address.
- [x] Two key maps with an LRU eviction hold the counters: 5 000 user keys and 20 000 client address keys (design decision D20). A full map evicts its oldest key, and a new key always gets a fresh counter. A test proves that 10 000 anonymous keys block no user and no new visitor.
- [x] A key text above 64 characters is replaced by its SHA-256 hex, thus a long header value or a long user id cannot fill the memory, and two keys never join.
- [x] With `OCTOMETER_CLIENT_IP_HEADER` set, the key comes from the last element of the last header line (the nearest proxy appends it). A header value above 64 characters or without a valid address form gives the remote address. The README says: set the option only behind a proxy that appends the address.
- [x] Each route test uses an injected clock, not the wall clock.
- [x] No stored event and no log line holds the IP address.
- [x] The Ktor route returns 429.
- [x] `./gradlew :kit:jvm-core:test` and `./gradlew :kit:jvm-ktor:test` pass.
- [x] A comment on this issue records the test output.

**Related.**
- #116 adds the three anonymous counters and the trusted proxy rule.
- #118 planned the two LRU maps; this issue builds them, thus #118 closes with the merge of #158.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, D20, section 4.2.

xamcross:
Body edit on 2026-09-21: #10 closed. The first line now names only the open blockers: #12.

xamcross:
[...]

Issue #104, opened 2026-09-21
Match the path against OCTOMETER_PATH_PATTERNS on the server


xamcross:
**Goal.** The kit repeats the client path rule on the server, because a client rule is not a control. The stored `path` is the result of the server match.

**Implementation steps.**
1. In `kit/jvm-core`, read the ordered route pattern list from `OCTOMETER_PATH_PATTERNS`.
2. Remove a trailing slash from the client value, but not from the root. Find the first pattern that matches.
3. Match a pattern only when it has the same count of segments as the path. A literal segment match ignores the ASCII letter case, and no pattern matches a prefix.
4. Write the literal `:name` for a `:name` segment. Keep the real segment for a `*` segment only when it matches `^[A-Za-z0-9](?:[A-Za-z0-9._~-]|%[0-9A-Fa-f]{2}){0,79}$`.
5. Store `/other` for a path without a match or with a bad `*` segment. Store `/other` also for an empty, a `.`, or a `..` segment.
6. Without the list, store no `path` at all, and write one warning at the start.

**Acceptance criteria.** Each test below uses the list `/`, `/articles`, `/articles/*`, `/history/:id`, `/ovdp/rates`.
- [x] `/history/42` stores `/history/:id`.
- [x] `/articles/ovdp-2026` stores `/articles/ovdp-2026`.
- [x] `/ARTICLES/` stores `/articles`, because the kit removes the trailing slash and ignores the letter case.
- [x] The stored value uses the text of the pattern, never the text of the input.
- [x] `/tokens/abc` stores `/other`, because no pattern matches.
- [x] `/articles` and `/articles/a/b` do not match `/articles/*`, because the segment counts differ.
- [x] A `*` segment of 81 characters stores `/other`.
- [x] `/articles//x`, `/articles/.`, and `/articles/..` store `/other`.
- [x] The kit never decodes a `%` escape before the match.
- [x] Without `OCTOMETER_PATH_PATTERNS` the store gets no `path`, and the kit writes one warning at the start.
- [x] A test reads `contract/examples/C42-path-match-cases.json` and asserts each case. The tracker of #106 reads the same file, thus the two matchers give the same result.
- [x] The pipeline sets the component `path` of the event record to the match result. A test store sees `/history/:id` for the client path `/history/42`, and no component of the event holds `/history/42`.
[...]

PR #122, opened 2026-09-21
docs(contract): add version 1.1 of the event log contract

xamcross:
## Summary

This change adds version 1.1 of the event log contract. It adds rules C38 to C43 to
`contract/README.md`, one new sentence to rule C32, seven new example files, and five new
tests in two test classes. It changes no Java production code.

## Correction round 1

Two reviews found that the first version of this text passed each of the 16 acceptance
criteria below, but still left a team that reads only the contract free to store a raw
client path or a free host name. Fixes, at commit `14748b7`:

- C38 now states the transport form (`ageMs: 0`, an entry of `clicks`) and the server rule
  for an unknown `octo:` element.
- A new rule C42 states how the server derives the stored `path`, with the literal `/other`
  for a value that matches no route pattern.
- C40 now states the closed set of three stored values, the rule for a host of the right
  form but no match, and a tighter Google pattern:
  `^([a-z0-9-]+\.)*google\.((com|co)\.[a-z]{2}|com|[a-z]{2})$`.
- A new rule C43 states the erasure of the anonymous events of a session.
- A version note near the top states that 1.1 is a minor change and what an old reader
  counts.
- The example rule gets its exception for a public search host of C40.
- Three new tests in `EventFieldValidatorTest` cover the three `event-*` example files.
- Two new example files show a masked path and the `/other` path.

Full findings and fixes: see the response comment on this pull request.

## Acceptance criteria (issue #101, original 16)

- [x] The title of `contract/README.md` is "Event log contract v1.1", and the file holds the
  line "Version 1.1." — line 1 and line 9.
- [x] Rule C38 reserves the element prefix `octo:` for the contract, and it states that the
  tracker drops such a `data-octo` value.
- [x] Rule C38 states that `octo:session-start` marks the first event of a session, that it
  is not a click, and that a reader of version 1.0 counts it as a click.
- [x] Rule C39 states the shape of `path`: optional, 1 to 150 bytes, starts with `/`, second
  character not `/`, the character set, the escape rule, and no query string and no fragment.
- [x] Rule C39 states the server match against the route pattern list (rule C42), and the
  stored result.
- [x] Rule C40 states the shape of `referrerHost`: optional on `octo:session-start`, the
  literal `other`, or 1 to 253 bytes with the pattern `[a-z0-9.-]+` and at least one dot.
- [x] Rule C40 holds the two entries `google.com` and `bing.com`.
- [x] Rule C40 states the match rule: lower-case host, an entry match by equality or by a
[...]

Issue #101, opened 2026-09-21
Add version 1.1 of the event log contract


xamcross:
**Goal.** `contract/README.md` holds the rules C38 to C41 and the version 1.1. Five new example files show the new fields.

**Evidence.** Observed on 2026-09-21: `kit/jvm-core/src/test/java/octometer/kit/core/ingest/IngestParserContractExamplesTest.java` reads each file of `contract/examples/`. A new example file joins that test. The owner said on 2026-09-21 that a visitor comes from Google search, from Bing search, or directly by an address. No ad campaign runs, thus version 1.1 has no field `utmSource`. The source list of rule C40 holds these two entries.

| Group | Entry (the stored value) |
|---|---|
| search | `google.com` |
| search | `bing.com` |

A host that matches `^([a-z0-9-]+\.)*google\.((com|co)\.[a-z]{2}|com|[a-z]{2})$` gives `google.com`. Each other host gives the literal `other`.

**Implementation steps.**
1. Add the rules C38, C39, C40, and C41 to `contract/README.md`. The acceptance criteria below hold the exact text of each rule.
2. Add the source list of the evidence above and its match rule to rule C40.
3. Add the sentence "The server also ignores an unknown field inside an entry of `clicks`." to rule C32.
4. Add one row for each rule C38 to C41 to the rule table at the end of `contract/README.md`.
5. Add the five example files to `contract/examples/`: `event-valid-C38-session-start.json`, `event-valid-C39-path.json`, `ingest-valid-C39-path.json`, `ingest-valid-C41-bad-path-dropped.json`, and `event-valid-C40-source.json`.
6. Add each new file to the example list of `IngestParserContractExamplesTest`.

**Acceptance criteria.**
- [x] The title of `contract/README.md` is "Event log contract v1.1", and the file holds the line "Version 1.1."
- [x] Rule C38 reserves the element prefix `octo:` for the contract, and it states that the tracker drops such a `data-octo` value.
- [x] Rule C38 states that `octo:session-start` marks the first event of a session, that it is not a click, and that a reader built before version 1.1 counts it as a click.
- [x] Rule C39 states: `path` is optional on an event and on an entry of `clicks`; it has 1 to 150 bytes in UTF-8; it starts with `/`; its second character is not `/`; each other character is in `[A-Za-z0-9._~!$&'()*+,;=:@/-]` or is a part of a well-formed escape `%[0-9A-Fa-f]{2}`; it holds no query string and no fragment.
[...]

PR #131, opened 2026-09-21
feat(kit/jvm-core): accept the path and referrerHost fields

### LINKED ISSUE #103: Accept and check the path and the source field in the kit parser
**Goal.** `kit/jvm-core` accepts the optional fields `path` and `referrerHost`, checks each one, and carries it to the store.

**Implementation steps.**
1. In `kit/jvm-core`, accept the optional field `path` in an entry of `clicks`. Check it with rule C39.
2. Accept the optional field `referrerHost` on the entry `octo:session-start`. Check it with rule C40.
3. Add the source list of rule C40 to `kit/jvm-core`, with the match rule of the contract.
4. Drop an invalid value under rule C41: keep the entry, keep the body valid, and write a maximum of one warning for each batch.
5. Carry `referrerHost` and the component `path` in the event record and in the call `EventLogStore.append`. The component `path` holds only the match result of the server, thus it stays empty until #104 (contract rules C39 and C42).
6. Add the tests of the acceptance criteria.

**Acceptance criteria.**
- [x] A `path` of 1 to 150 bytes in UTF-8 that starts with `/`, whose second character is not `/`, and whose other characters are in `[A-Za-z0-9._~!$&'()*+,;=:@/-]` or in a well-formed escape `%[0-9A-Fa-f]{2}`, passes the check, and the parsed click holds it. It does not reach the store: the event record holds no `path` before #104 (contract rule C42).
- [x] A `path` of 151 bytes, a `path` that starts with `//`, a `path` with `?`, a `path` with `#`, and a `path` with a partial escape are dropped, and the entry stays.
- [x] A `referrerHost` value `other` reaches the store.
- [x] A `referrerHost` that is an entry of the source list of rule C40 reaches the store.
- [x] A `referrerHost` outside that list with the form of a host name (1 to 253 bytes, the pattern `[a-z0-9.-]+`, and a minimum of one dot) reaches the store as the literal `other`. A value with a different form is dropped.
- [x] A `referrerHost` on an entry that is not `octo:session-start` does not reach the store.
- [x] An entry whose `element` starts with `octo:` and is not `octo:session-start` is dropped, and the other entries of the batch reach the store.
- [x] A test reads the source list of `contract/README.md`. It asserts that the list of `kit/jvm-core` holds the same two entries.
- [x] A `referrerHost` on a click entry is dropped.
- [x] A batch with three invalid values writes a maximum of one warning, and the response stays 204.
- [x] No log line holds a raw path or a raw host.
- [x] A test store records each event of `append`. For a request with a valid `path`, no component of the event and no `toString()` text of the event or of the parsed click holds the client value.
- [x] `./gradlew :kit:jvm-core:test` passes.

**Related.**
- #101 gives the four rules.
- #104 matches the path against the route pattern list.
- #105 writes the two fields in the MongoDB store.
- #26 closed on 2026-09-21. It gives the `EventLogStore` that carries the two fields.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, the contract version 1.1 of #101, and the rules C39, C40, C41, and K1.

xamcross:
## Summary

`kit/jvm-core` now accepts `path` (rule C39) and `referrerHost` (rule C40) on a
`clicks` entry. It drops an invalid value and keeps the entry (rule C41). It
also drops an entry with a reserved `octo:` element prefix, in each letter
case, except the exact text `octo:session-start` (rule C38).

## Correction round 1 (2026-09-22)

Two reviews found the same MAJOR defect: `EventLogStore.append` gave the raw
client `path` to each store, through the public accessor
`IngestEvent.path()`. Rule C39 says the server never stores the raw client
value. Rule C42 says the server stores no `path` field without a route
pattern list, and this module has no such list yet.

**The false sentence.** The first pushed commit stated: "no store gets the
raw path: `StoredEvent` copies `referrerHost`, but not `path`, from
`IngestEvent`." That sentence was false. `EventLogStore.append` takes
`List<IngestEvent>`, and an app store reads the raw value straight from
`IngestEvent.path()`. `StoredEvent` gave no protection to an app store, only
to the in-memory store of this module.

**The fix.** `IngestEvent` keeps its component `path`. It now holds only the
match result of rules C39 and C42, and it never holds the raw client value.
This module has no route pattern list yet, so `IngestPipeline.process` always
sets this component to `null`, also for a client value that passes the shape
check of rule C39. Issue #104 adds the route pattern list and the matcher,
and it then fills this component. The raw client value, after the shape
check, stays in the parsed click (`ParsedClick`) only, and no store reads it.
`StoredEvent` now also holds `path`, and `StoredEvent.of` copies it (and
`referrerHost`) from `IngestEvent`. The Javadoc of `IngestEvent#path()` and
of `EventLogStore#append` each state this rule.

`ParsedClick` and `ParsedIngestRequest` now override `toString()`. Neither
text holds a raw path, a raw host, or a session id.

## Acceptance criteria

- [x] A valid `path` (1 to 150 bytes, starts with `/`, second character not
  `/`, each other character in the C39 set or a well-formed `%XX` escape)
  passes the shape check.
  Evidence: `IngestParserContractExamplesTest.parsesTheValidPathFieldOfAClick`
  proves the parsed click holds it.
  `IngestPipelinePathAndSourceTest.aValidPathPassesTheShapeCheckButDoesNotReachTheEventRecord`
  proves `IngestEvent.path()` stays `null`, per the correction above.
- [x] A `path` of 151 bytes, a `path` that starts with `//`, a `path` with
  `?`, a `path` with `#`, and a `path` with a partial escape are dropped, and
  the entry stays.
  Evidence: `IngestPipelinePathAndSourceTest.aPathOf151BytesIsDropped`,
  `.aPathThatStartsWithTwoSlashesIsDropped`, `.aPathWithAQuestionMarkIsDropped`,
  `.aPathWithAFragmentIsDropped`, `.aPathWithAPartialEscapeIsDropped`, and the
  contract example test `.anInvalidPathIsDroppedAndTheEntryStays`.
- [x] A `referrerHost` value `other` reaches the event record.
  Evidence: `IngestPipelinePathAndSourceTest.theReferrerHostOfTheOtherLiteralReachesTheEventRecord`.
- [x] A `referrerHost` of the source list reaches the event record.
  Evidence: `.aReferrerHostOfTheSourceListReachesTheEventRecord`, and
  `EventFieldValidatorReferrerHostTest.matchesEachEntryOfTheSourceListExactlyAndAsASubdomain`
  (new; the match now reads the ordered source list itself, not a private copy).
- [x] A `referrerHost` outside the list with a host-name form reaches the
  event record as the literal `other`. A value with a different form is
  dropped.
  Evidence: `.aReferrerHostOutsideTheListWithAHostNameFormBecomesOther`,
  `.aReferrerHostWithADifferentFormIsDropped`.
- [x] A `referrerHost` on an entry that is not `octo:session-start` does not
  reach the event record.
  Evidence: `.aReferrerHostOnAClickEntryDoesNotReachTheEventRecord`.
- [x] An entry whose `element` starts with `octo:` and is not
  `octo:session-start` is dropped; the other entries reach the event record.
  Evidence: `.anEntryWithAReservedPrefixThatIsNotSessionStartIsDroppedAndOthersStay`,
  `.anUpperCaseReservedPrefixIsAlsoDropped`,
  `.onlyTheExactLowerCaseTextMarksTheSessionStart`.
- [x] A test reads the source list of `contract/README.md` and asserts that
  the list of `kit/jvm-core` holds the same two entries.
  Evidence: `EventFieldValidatorReferrerHostTest.theSourceListMatchesTheContractSourceList`.
- [x] A `referrerHost` on a click entry is dropped.
  Evidence: `.aReferrerHostOnAClickEntryDoesNotReachTheEventRecord` (above).
- [x] A batch with three invalid values writes a maximum of one warning.
  Evidence: `.aBatchWithThreeInvalidPathValuesWritesAtMostOneWarning`, and
  `.aBatchWithThreeReservedElementsWritesAtMostOneWarning` (new, for rule C38).
  `.theResponseStaysValidForABatchWithInvalidValues` asserts the one fact
  that this module owns: `process()` throws nothing and keeps every entry.
- [x] No log line holds a raw path or a raw host.
  Evidence: `.noLogLineHoldsARawPathOrARawHost`.
- [x] **New criterion of the correction.** A test store records each event
  of `append`. For a request with a valid `path`, no component of the event,
  and no `toString()` text of the event or of the parsed click, holds the
  client value.
  Evidence: `IngestPipelinePathAndSourceTest.aStoreNeverReceivesTheRawClientPathForAValidPath`.
- [x] `./gradlew :kit:jvm-core:test` passes.
  Evidence: below.
  Evidence: posted as a separate issue comment.

## Review findings addressed

Full detail is in the pull request comment of this correction round. In
short: both MAJOR findings are fixed (the raw-path seam above, and
`matchReferrerHost` now reads the ordered source list). 12 of 14 MINOR
findings are fixed. Two stay open, with a reason each, because the review
gives no exact fix for them: a silent drop of a wrong-type `path` or
`referrerHost` value, and an out-of-date sentence in `contract/README.md`
(out of scope: the contract stays unchanged in this pull request).

## Test evidence

Command: `./gradlew :kit:jvm-core:test` (JDK 21, `C:/Users/xamcr/.jdks/ms-21.0.9`).

Result: `BUILD SUCCESSFUL`. 138 tests, 0 failures, 0 errors (sum of each
`<testsuite tests="..." failures="..." errors="...">` under
`kit/jvm-core/build/test-results/test/`). 6 tests are new in this
correction round.

Command: `git merge origin/main` (a merge commit, not a rebase), then
`./gradlew build` for the full repository, with Docker Desktop running for
the `kit/jvm-mongo` container tests.

Result: `BUILD SUCCESSFUL`. `kit/jvm-core` (138 tests), `kit/jvm-mongo` (52
tests), `kit/jvm-ktor` (28 tests), and `monitor:backend` (66 tests) each
build and pass. `main` now holds `kit/jvm-ktor` and `kit/jvm-mongo`; each one
uses the 3-argument `IngestEvent` constructor, so each call site still
compiles with no change.

## Not in this pull request

- The match of `path` against the route pattern list (rule C42) — issue #104.
- The write of `path` and `referrerHost` to the MongoDB store — issue #105.

Closes #103

Generated with Claude Code


xamcross:
**Review (Java parser and library engineer, first review): CHANGES REQUESTED**

I made my own clone. I checked out `issue-103-parser-path-source` at `ad9c883`. I ran each
command with `JAVA_HOME` set to the JDK 21 of the machine. My first source for each rule is
`contract/README.md` version 1.1 on `main`.

## The commands that I ran

| Command | Result |
|---|---|
| `./gradlew :kit:jvm-core:test` | `BUILD SUCCESSFUL`. 132 tests, 0 failures, 0 errors (sum of each `<testsuite>` element). |
| `./gradlew :kit:jvm-core:dependencies --configuration runtimeClasspath` | `No dependencies`. The module adds no runtime dependency. |
| `./gradlew build` on the branch head | `BUILD FAILED` one time, in `:kit:jvm-mongo:test`, with `ContainerFetchException` and `client version 1.32 is too old`. This is a Docker fault of my machine. A second run passed. |
| `git merge origin/main` in my clone, then `./gradlew build` | `BUILD SUCCESSFUL in 1m 23s`. `kit/jvm-ktor` (Kotlin), `kit/jvm-mongo`, and `monitor:backend` all build and pass. |

The merge check answers the API question of point 5. `kit/jvm-mongo` builds `new
IngestEvent(sessionId, element, ts)` in five test files. `kit/jvm-ktor` overrides
`append(events: List<IngestEvent>, userId: String?)` in eight places. The 3-argument
constructor keeps each one correct. No file outside `kit/jvm-core` names `StoredEvent`.

## The acceptance criteria

|---|---|---|
| A valid `path` reaches the store | PASS | My table rows 01 to 05, 18, 19. `/articles/example` reaches `IngestEvent.path()`, and `EventLogStore.append` takes `List<IngestEvent>`. |
| 151 bytes, `//`, `?`, `#`, a partial escape are dropped, the entry stays | PASS | My table rows 06, 09, 10, 11, 12, 13. Each one gives one event with `path == null`. |
| The literal `other` reaches the store | PASS | Row 25. |
| An entry of the source list reaches the store | PASS | Rows 26, 31. |
| A host-name form outside the list gives `other`; a different form is dropped | PASS | Rows 33, 36 (253 bytes) against rows 34, 35, 37, 38. |
| A `referrerHost` on an entry that is not `octo:session-start` is dropped | PASS | Rows 40, 42. |
| An `octo:` entry is dropped, the other entries stay | PASS | Rows 42 to 48. `OCTO:foo`, `Octo:Session-Start`, `OCTO:SESSION-START`, and the bare text `octo:` are all dropped. |
| A test reads the source list of `contract/README.md` | PASS, with MINOR 1 | I changed `bing.com` to another host in the contract file. `theSourceListMatchesTheContractSourceList` FAILED on a forced run. |
| A `referrerHost` on a click entry is dropped | PASS | Row 40. |
| Three invalid values give a maximum of one warning | PASS | My probe: three bad paths plus one bad host give 1 message. |
| No log line holds a raw path or a raw host | PASS | Both warning texts are constant. No parameter reaches the logger. |
| `./gradlew :kit:jvm-core:test` passes | PASS | See the table above. |
| A comment on the issue records the test output | PASS | The comment of 2026-09-21 on issue #103. |

## My own table of 52 inputs

I wrote a table test in my clone and ran it: **52 rows, 0 failures**. The rows cover rule C39
(20 rows), the JSON type of `path` (4 rows), rule C40 (16 rows), rule C38 (8 rows), and the
strict rules that stay fatal (4 rows). The rows that need a note:

- `/a%00` and `/%2F%2e%2e/etc` are **kept**. Rule C39 allows a well-formed `%XX` escape, and
  rule C42 says the server decodes no escape. The code is correct. Issue #104 must handle the
  encoded dot segment.
- `www.google.de` and `www.google.co.uk` give `google.com`. The pattern in the code is the
  exact pattern of rule C40. `attacker.google.top` and `google.zip` give `other`. Correct.
- `a_b.com` gives `null` (a drop). The underscore breaks the pattern `[a-z0-9.-]+`. Correct.
- A `path` of 151 bytes is rejected; a `path` of 150 bytes passes. The code counts UTF-8 bytes
  with `getBytes(StandardCharsets.UTF_8).length`. Correct against rule C39.
- A duplicate `path` key gives `DUPLICATE_FIELD` (rule C36). An unknown field `pathx` inside an
  entry is ignored (rule C32). A session start with no `ageMs` gives `MISSING_FIELD` (rule C13).
  An element `octo:session start` gives `ELEMENT_PATTERN` (rules C4, C33).

## The hostile-body limits (point 6)

| Input | Result |
|---|---|
| A `path` value that is an object 40 levels deep | `INVALID_JSON` (the depth limit of 32) |
| A `path` value that is an array 200 levels deep | `INVALID_JSON`. No `StackOverflowError`. |
| A `path` string of 15 kB | The entry stays, the field is dropped |
| A body of 20 kB | `BODY_TOO_LARGE` |
| 51 clicks, each with a `path` | `TOO_MANY_CLICKS` |
| A `path` with a lone high surrogate | The field is dropped |
| A duplicate `referrerHost` key | `DUPLICATE_FIELD` |
| A `path` of the JSON type number, null, object, or array | The field is dropped, the entry stays |

The parser stays strict and safe. The new field read reuses `skipValue`, so the depth guard and
the size guard both hold.

## I broke the code, and I saw the tests fail

| I changed | The tests that failed |
|---|---|
| `MAX_PATH_BYTES` 150 to 200 | `rejectsAPathOf151Bytes`, `aPathOf151BytesIsDropped` |
| `regionMatches(true, ...)` to `regionMatches(false, ...)` | `anUpperCaseReservedPrefixIsAlsoDropped`, `onlyTheExactLowerCaseTextMarksTheSessionStart` |
| `[a-z0-9.-]+` to `[a-zA-Z0-9.-]+` | `givesNullForAHostWithAnUpperCaseLetter`, `aReferrerHostWithADifferentFormIsDropped` |
| Removed `isSessionStart &&` from the `referrerHost` branch | `aReferrerHostOnAClickEntryDoesNotReachTheEventRecord` |
| Removed the one-warning guard | `aBatchWithThreeInvalidPathValuesWritesAtMostOneWarning` |
| Added a third entry to the source set | `referrerHostSourceListHoldsExactlyTwoEntries`, `theSourceListMatchesTheContractSourceList` |
| Printed `path` in `IngestEvent.toString()` | `toStringHoldsNoPathAndNoReferrerHost` |
| Changed the source list inside `contract/README.md` | `theSourceListMatchesTheContractSourceList` (on a forced run; see MINOR 1) |

No test is empty. No test is always green. I restored each file after each experiment.

---

# Findings

## MAJOR 1 — the C42 seam gives no protection to `kit/jvm-mongo` or to an app store

`kit/jvm-core/src/main/java/octometer/kit/core/store/EventLogStore.java:37` and
`kit/jvm-core/src/main/java/octometer/kit/core/ingest/IngestEvent.java:24`.

The pull request text says: "no store in this module gets a raw client path". The commit body
says more: "Until then, no store gets the raw path: StoredEvent copies referrerHost, but not
path, from IngestEvent." The second sentence is false.

Evidence:

- `EventLogStore.append` takes `List<IngestEvent>`, not `List<StoredEvent>`. Each store gets
  the raw client `path` through the public accessor `IngestEvent.path()`.
- `kit/jvm-mongo/src/main/java/octometer/kit/mongo/store/MongoEventLogStore.java:163`,
  `toDocument(IngestEvent event, String userId)`, builds the BSON document straight from
  `IngestEvent`. It never calls `StoredEvent.of`.
- A grep for `StoredEvent` in `kit/jvm-mongo`, `kit/jvm-ktor`, and `monitor` gives zero hits.
  Only `InMemoryEventLogStore` uses `StoredEvent`.

So `StoredEvent` protects the in-memory store only. Issue #105 edits `toDocument`. One line
`.append("path", event.path())` there breaks rule C39 ("The server never stores the raw client
value") and rule C42 ("Without a route pattern list, the server stores no `path` field"). The
accessor is named `path`, so that line looks correct to a reader.

**The exact fix.** Give the raw value a name that stops the reflex, in `IngestEvent.java:24`:

```java
public record IngestEvent(String sessionId, String element, Instant ts,
        String clientPath, String referrerHost) {

    /**
     * The `path` value of the client, after the shape check of rule C39
     * only.
     *
     * <p><strong>A store must never write this value.</strong> Rule C39
     * says the server never stores the raw client value. Rule C42 says
     * the server stores no `path` field without a route pattern list,
     * and this module has no such list. Issue #104 adds the list and a
     * second component `storedPath` with the match result. A store
     * writes `storedPath`, never `clientPath`.
     */
    public String clientPath() {
        return clientPath;
    }
```

Then change `IngestPipeline.process:111` to `new IngestEvent(parsed.sessionId(),
click.element(), ts, path, referrerHost)` with the new name, and change the three tests that
call `.path()`.

**What #104 and #105 then need.**

- #104 adds `OCTOMETER_PATH_PATTERNS`, the matcher of rule C42, the startup warning for a
  missing list, a sixth component `storedPath` on `IngestEvent`, and the `path` component on
  `StoredEvent`. `StoredEvent.of` copies `storedPath`, never `clientPath`.
- #105 changes `MongoEventLogStore.toDocument` to build the document from `StoredEvent.of(event,
  userId)` and not from `IngestEvent`. The guard of rules C39 and C42 then lives in one place.
  #105 also needs a test that asserts the document holds no raw client path.

**Also fix the text.** Correct the pull request body and the commit body: `kit/jvm-mongo` builds
its document from `IngestEvent`, so `StoredEvent` gives it no protection today.

## MAJOR 2 — `matchReferrerHost` does not use the source list that the contract test guards

`kit/jvm-core/src/main/java/octometer/kit/core/ingest/EventFieldValidator.java:46`,
`:182` to `:197`.

`REFERRER_HOST_SOURCE_LIST` has two uses: the declaration on line 46, and the return on line
167. `matchReferrerHost` does not read it. It holds the two host names as string literals on
line 190 and line 193.

The criterion asks for a test that stops a drift between the contract and the kit. That test
compares the **set** only. The match rule keeps its own copy of the two names. A person who
adds a third entry to the contract and to the set gets a green build, and the new entry never
matches: the method returns `other` for it. The drift moves into the match rule, where the test
cannot see it.

**The exact fix.** Read the list in the match:

```java
/** The source list of rule C40, besides the literal `other`. The order fixes the match order. */
private static final List<String> REFERRER_HOST_SOURCE_LIST = List.of("google.com", "bing.com");

private static final Set<String> REFERRER_HOST_SOURCE_SET = Set.copyOf(REFERRER_HOST_SOURCE_LIST);

public static Set<String> referrerHostSourceList() {
    return REFERRER_HOST_SOURCE_SET;
}

public static String matchReferrerHost(String rawValue) {
    if (REFERRER_HOST_OTHER.equals(rawValue)) {
        return REFERRER_HOST_OTHER;
    }
    if (!isValidHostNameShape(rawValue)) {
        return null;
    }
    for (String entry : REFERRER_HOST_SOURCE_LIST) {
        if (matchesSourceEntry(rawValue, entry)) {
            return entry;
        }
    }
    if (GOOGLE_HOST_PATTERN.matcher(rawValue).matches()) {
        return "google.com";
    }
    return REFERRER_HOST_OTHER;
}
```

Add one test: for each entry of `referrerHostSourceList()`, `matchReferrerHost(entry)` returns
that entry, and `matchReferrerHost("www." + entry)` returns that entry too.

---

## MINOR findings (9)

These do not prevent an approval.

1. **`kit/jvm-core/build.gradle.kts:27`** — `contract/README.md` is not an input of the test
   task. I changed the source list inside the contract file and ran `./gradlew
   :kit:jvm-core:test`. Gradle printed `Task :kit:jvm-core:test UP-TO-DATE` and the build
   stayed green. The test failed only with `--rerun-tasks`. CI runs from a clean checkout, so
   CI still catches the drift, but a local run does not. Fix:
   ```kotlin
   tasks.test {
       useJUnitPlatform()
       inputs.file(rootProject.layout.projectDirectory.file("contract/README.md"))
           .withPropertyName("contractReadme")
           .withPathSensitivity(PathSensitivity.RELATIVE)
   }
   ```
   `ExampleFiles` has the same gap for `contract/examples/`. Add that directory too.

2. **`IngestPipelinePathAndSourceTest`** — no test caps the C38 warning at one for each batch. I
   proved the cap by hand: a batch with three reserved elements gives 1 message. Add a test with
   the shape of `aBatchWithThreeInvalidPathValuesWritesAtMostOneWarning`.

3. **`IngestParser.java:279`** — a `path` or a `referrerHost` value of a JSON type other than a
   string gives **zero** warnings. I measured it. Rule C41 says "a maximum of one", so zero is
   correct, but a broken tracker then stays silent. Consider a `WRONG_FIELD_TYPE` path through
   the same one-warning guard.

4. **`kit/jvm-core/src/main/java/octometer/kit/core/store/StoredEvent.java:16`** — the record
   went from 4 components to 5, and it kept no 4-argument constructor. `IngestEvent` kept its
   3-argument one. No call site in the repository breaks, but an app that builds a
   `StoredEvent` does. Add the same compatibility constructor for symmetry:
   ```java
   public StoredEvent(String sessionId, String element, Instant ts, String userId) {
       this(sessionId, element, ts, userId, null);
   }
   ```

5. **Language, rule 7 of ASD-STE100** — four `-ing` forms act as a noun or as a verb:
   - `EventFieldValidator.java:18` "instead of stopping the whole request" →
     "and it does not stop the whole request".
   - `EventFieldValidator.java:116` "It returns a boolean instead of throwing" →
     "It returns a boolean. It does not throw."
   - `EventFieldValidator.java:175` "instead of stopping the whole request" → as above.
   - `IngestEvent.java:29` "keeps working with this constructor" →
     "still compiles with this constructor".

6. **Language, rules 1 and 4** — `EventFieldValidatorReferrerHostTest.java:100` holds a comment
   of 29 words that joins two statements with "so". Split it into two sentences.

7. **Language, the pull request text** — "keep compiling with no change" is an `-ing` form as a
   verb, and that sentence has more than 25 words. Write "each call site still compiles."

8. **The pull request text** — it says "`EventFieldValidator.isValidPath`, `IngestEvent`, and
   `StoredEvent` each carry a comment that names issue #104 and rule C42". The Javadoc of
   `isValidPath` (`EventFieldValidator.java:111` to `:118`) names rule C39 and rule C41 only.
   Add the note, or correct the text.

9. **A follow-up for the contract** — `contract/README.md` rows C32 and C41 of the example table
   say that the `path` field of `ingest-valid-C41-bad-path-dropped.json` is an ignored unknown
   field "under the present parser". After this merge that sentence is false. The contract file
   is out of scope here. Please open a follow-up issue.

---

## Scope, secrets, and the record change

- **Scope (PASS).** The diff touches 13 files, all under `kit/jvm-core`. There is no route
  pattern match (#104), no MongoDB field (#105), and no rate limit.
- **Secrets (PASS).** The diff holds no credential, no database file, and no build output. Each
  UUID and each host name is fake or public.
- **Design decision D15 (PASS).** `IngestEvent.toString()` prints the element and `ts` only.
  `StoredEvent.toString()` prints the element, `ts`, and a redacted user id. Neither one prints
  the session id, the path, or the referrer host. I broke `IngestEvent.toString()` and the test
  failed.
- **Warnings (PASS).** Both messages are constant text through `java.lang.System.Logger`. The
  C38 drop gives a maximum of one message for each batch, and the C41 drops give a maximum of
  one more. No message holds a value of the body.

Please fix MAJOR 1 and MAJOR 2. I will then approve.

xamcross:
**Review (application security engineer, first review): CHANGES REQUESTED**

I attacked the new code in my own clone of the branch `issue-103-parser-path-source`
(head `ad9c883`), with `origin/main` merged in. I wrote 27 attack tests of my own. Each
one passes. The parser holds against every hostile input that I sent. One MAJOR finding
remains: the public store interface hands the raw client `path` to an app, and its
Javadoc gives no warning. Rule C39 says the server never stores that value.

## Commands that I ran

- `./gradlew :kit:jvm-core:test` — `BUILD SUCCESSFUL`, 132 tests, 0 failures.
- With my 27 attack tests added: `BUILD SUCCESSFUL`, 159 tests, 0 failures.
- After a merge of `origin/main`: `:kit:jvm-ktor:test` gives `BUILD SUCCESSFUL`,
  28 tests, 0 failures. The 3-argument `IngestEvent` constructor keeps the Ktor route
  and its tests green.
- `:kit:jvm-mongo:test` fails on this machine because Docker cannot pull `mongo:7.0`.
  That is an environment fault, not a fault of this branch. `kit/jvm-mongo` is byte for
  byte the same on the branch and on `origin/main`.

## Acceptance criteria

| --- | --- | --- |
| A valid `path` reaches the store seam | PASS | `/` plus 149 letters gives `KEPT`. |
| 151 bytes, `//`, `?`, `#`, a partial escape are dropped; the entry stays | PASS | Each one gives `FIELD_DROPPED` with 1 event. |
| `referrerHost` `other` reaches the record | PASS | `KEPT:other`. |
| A source-list entry reaches the record | PASS | `google.com`, `bing.com`, `www.google.co.uk`. |
| A host-name form outside the list gives `other`; another form is dropped | PASS | `evil.example` gives `other`; `GOOGLE.COM` and `localhost` are dropped. |
| `referrerHost` on another element does not reach the record | PASS | 4 element values, each gives `null`. |
| An `octo:` entry that is not the marker is dropped; the batch stays | PASS | `octo:foo`, `OCTO:foo`, `Octo:Session-Start`, `octo:` each drop one entry only. |
| A test reads the source list of `contract/README.md` | PASS | `theSourceListMatchesTheContractSourceList` reads the file. |
| A `referrerHost` on a click entry is dropped | PASS | Same test as above. |
| A batch with three invalid values writes a maximum of one warning | PASS | 41 bad values give 2 warnings: one for C38, one for C41. |
| No log line holds a raw path or a raw host | PASS | See the log capture below. |
| `./gradlew :kit:jvm-core:test` passes | PASS | `BUILD SUCCESSFUL`. |
| A comment on the issue records the test output | PASS | The comment exists. |

## What I attacked, and the result

**1. `path`, 34 forms, each through `IngestPipeline.process`.** No input made the body
invalid through `path`. No input threw a class other than `IngestException`.

```
150 bytes -> KEPT          151 bytes -> DROPPED      10 000 chars -> DROPPED
//evil.example/x -> DROPPED                          /\evil -> DROPPED
a lone backslash -> DROPPED                          empty text -> DROPPED
\u0000, \u001f, \u007f, \n, \r, \t -> DROPPED        a raw control character -> INVALID_JSON
% -> DROPPED   %2 -> DROPPED   %zz -> DROPPED        %00, %2F, %2f%2e%2e -> KEPT (C39 allows them)
a non-ASCII letter, raw or escaped -> DROPPED        a surrogate half -> DROPPED
? and #, raw and each as a JSON escape -> DROPPED    a number, true, null, an object -> DROPPED
an array of 4 000 items -> DROPPED                   a nest 28 deep -> DROPPED
```

A raw NUL, a raw new line, and a raw tab inside a JSON string give `INVALID_JSON`. That
is broken JSON, and JSON itself forbids it. A `path` of 1 000 000 characters and an array
of 10 000 items each give `BODY_TOO_LARGE`, from the 16 KB limit of rule C18 that this
module owns itself. A nest 40 deep gives `INVALID_JSON` from the depth limit of the
parser. Each of the three limits is a rule of the parser, not a rule of `path`.

A fuzz run of 200 000 random strings through `isValidPath` threw nothing. Each accepted
value holds only characters of the C39 set.

The cost stays linear. A direct call of `EventFieldValidator.isValidPath` takes 0.107 ms
for 10 000 characters and 0.398 ms for 1 000 000 characters.

**2. `referrerHost`, 20 forms plus a fuzz run of 300 000 random values.** The stored value
is always `google.com`, `bing.com`, or `other`, or the field is absent. No free host name
reached the event record.

```
other -> other            google.com -> google.com     bing.com -> bing.com
GOOGLE.COM -> DROPPED     localhost -> DROPPED         254 bytes -> DROPPED
a new line -> DROPPED     a NUL -> DROPPED             a number, an array, an object -> DROPPED
google.com. -> other      a..b -> other                93.184.216.34 -> other
evil.example -> other     attacker.google.top -> other 253 bytes -> other
.google.com -> google.com (the value stays in the closed set)
```

`matchReferrerHost` returns a literal of the closed set or `null`. It never returns the
input. The fuzz run proves it for 300 000 values. A `referrerHost` on `a.one`, on
`checkout.save`, on `octo:session-start-x`, and on `OCTO:SESSION-START` never reaches the
record. A second `octo:session-start` entry with `payroll.acme.example` gives `other`.

**3. The reserved prefix (rule C38).**

```
octo:foo            -> the entry drops, keep.one and keep.two stay
OCTO:foo            -> the entry drops, the others stay
Octo:Session-Start  -> the entry drops, the others stay
octo:               -> the entry drops, the others stay
octo:session-start  -> the entry stays
octo                -> the entry stays (the prefix needs the colon)
a trailing space    -> ELEMENT_PATTERN, the whole batch gives 400
a NUL               -> ELEMENT_PATTERN, the whole batch gives 400
a Cyrillic o        -> ELEMENT_PATTERN, the whole batch gives 400
```

The last three break rule C4, thus rule C33 makes the whole body invalid before rule C38
runs. See MINOR 5.

A batch of 50 entries that all hold a reserved element gives 0 events and 1 warning.
`IngestPipeline.ingest` then appends nothing. The Ktor route of `origin/main` answers 204
for that batch (`IngestRoute.kt`, `call.respond(HttpStatusCode.NoContent)` after an
`ingest` call that throws nothing).

**4. The warnings.** A batch with 1 bad host, 20 bad paths, and 20 reserved elements
writes exactly 2 lines: one for rule C38 and one for rule C41. Both lines are fixed text.

```
The batch holds an entry with an invalid path or referrerHost value. The server drops the
field and keeps the entry (contract rule C41).
The batch holds an entry whose element starts with the reserved prefix octo: and is not
octo:session-start. The server drops the entry and keeps the rest of the batch (contract
rule C38).
```

I sent a batch with the path `/users/4242/token-S3CR3TV4LUE?q=x` and the host
`intranet.acme-payroll.example\nINJECTED WARNING`, then searched the captured log. No
line holds the raw path, the raw host, the session id, the injected text, or a line
break. Log injection fails.

10 000 batches with a bad path give 10 000 log lines, 1 360 000 characters of message
text, and 75 ms of total work (7.5 us for each batch). The contract asks for a maximum of
one warning for each batch (rule C41), and it names the rate limit as the flood control
(rule C38, last sentence). This branch meets both. A rate limit of the warning itself is
therefore not necessary. The rate limit of the route is necessary: issue #33 is open with
the label `production-blocker`, and until it lands each anonymous request can write one
warning line. Close issue #33 before the public route starts.

**5. The seam for issue #104.** I sent the path `/users/4242/reset/S3CR3TV4LUE`.

- `IngestEvent.toString()` holds no path and no host. A `List` and a `Map` that hold the
  event print the same safe text.
- `StoredEvent` has no `path` record component. I checked it with reflection.
- `InMemoryEventLogStore` holds no raw path.
- `MongoEventLogStore.toDocument` (`kit/jvm-mongo/.../MongoEventLogStore.java:163`) builds
  the document from a fixed list of four fields: `ts`, `element`, `sessionId`, `userId`.
  It reads no other component of `IngestEvent`. No store of `main` can store the raw
  client path today. No store of this module gets a raw client path.
- An app store does get the raw value. See MAJOR 1.

**6. The user id.** A `userId` field at the body level, a `userId` field inside a click
entry, and a `userId` field beside a `path` field each have no effect. The stored user id
stays the value of the `UserIdResolver`. An anonymous batch keeps a `null` user id and
still carries `referrerHost: google.com`.

**7. The cost.** Each number is the best of 3 rounds of 100 000 bodies, after 20 000
warm-up runs, on the same machine with the same JDK.

| Body | Branch | `main` |
| --- | --- | --- |
| 50 entries, no `path`, no `referrerHost` (2 KB) | 29.7 us | 32.3 us |
| 50 entries, each with a `path` of 150 bytes (9.8 KB) | 117.3 us | 51.4 us |
| 50 `octo:session-start` entries, each with a hostile host of 253 bytes (16 KB) | 818.3 us | 99.7 us |

A body without the new fields costs the same as on `main`. A body with 50 legal paths
costs 2.3 times the `main` value, because the parser now builds and checks each string
instead of a skip of the field. That is the cost of the feature. The cost stays linear.

The worst 253-byte host that I found costs 29.6 us in `matchReferrerHost`. I searched 15
hand-made shapes and 40 random shapes. The cost is polynomial, not exponential: the
mandatory dot between two labels makes each split unique, so `([a-z0-9-]+\.)*` cannot
grow exponentially. There is no ReDoS. See MINOR 4 for a one-line improvement.

**8. Secrets.** I searched the diff for a credential form: a connection string, a key, a
token, a private key block, and each common provider prefix. The diff holds none. The
only match is the test literal `/secret-token-path`, which is a path and not a
credential. The pull request text and the issue comments hold none. The diff holds only
`.java` files: no database file, no build output.

## Findings

### MAJOR 1 — the store interface hands the raw client path to an app, with no warning

`kit/jvm-core/src/main/java/octometer/kit/core/store/EventLogStore.java:37`
(and the Javadoc above it, lines 7 to 36).

`append` takes `List<IngestEvent>`, and `IngestEvent.path()` now holds the raw client
value after the shape check of rule C39 only. `EventLogStore` is the extension point of
the app: its own Javadoc says "An app gives one implementation", and the Ktor route of
`origin/main` passes the store of the app straight through. My test proves the exposure:

```java
EventLogStore appStore = new EventLogStore() {
    public void append(List<IngestEvent> events, String userId) {
        events.forEach(event -> seen.add(String.valueOf(event.path())));
    }
    public void deleteByUserId(String userId) { }
};
IngestPipeline.ingest(bodyWithThePath, clock, () -> "user-1", appStore, settings);
// an app store sees path() = [/users/4242/reset/S3CR3TV4LUE]
```

Rule C39 says "The server never stores the raw client value". Rule C42 says "Without a
route pattern list, the server stores no `path` field", and it warns that a path can hold
an identifier, a token, or a search term. The class comment of `IngestEvent` states this
correctly, but an app team reads the store interface, not the record. A team that writes
`document.append("path", event.path())` today breaks rule C39 and writes a token into the
database. The two stores of this repository are safe, thus this is a MAJOR finding and
not a BLOCKER.

**The exact fix.** Add this paragraph to the Javadoc of `append`, above
`void append(List<IngestEvent> events, String userId);`:

```java
    /**
     * ...
     *
     * <p>A store must not write {@link
     * octometer.kit.core.ingest.IngestEvent#path()}. That value is the
     * raw client path after the shape check of rule C39 only. Rule C39
     * says the server never stores the raw client value, and rule C42
     * says the server stores no `path` field without a route pattern
     * list. Issue #104 adds the match; a store then writes only the
     * result of that match. A raw path can hold a token, an email
     * address, or a user id (contract rule C42).
     */
```

A stronger option, if you prefer a guard in the code: let `IngestPipeline.ingest` build the store
batch with `path` set to `null` until issue #104 exists, and keep the value in
`process()` only. Please pick one of the two and say which one in the pull request text.

### MINOR 1 — `isValidPath` allocates before it rejects a huge value

`kit/jvm-core/src/main/java/octometer/kit/core/ingest/EventFieldValidator.java:120`

`path.getBytes(StandardCharsets.UTF_8)` builds a byte array of the whole value before any
cheap check. A value of 1 000 000 characters costs 0.398 ms and an array of 1 to 4 MB.
`IngestParser.checkBodySize` already holds the correct pattern for this: a UTF-8 encoding
never needs fewer bytes than the string has UTF-16 chars.

**The exact fix.** Put a length gate in front of line 120:

```java
    public static boolean isValidPath(String path) {
        // A UTF-8 encoding never needs fewer bytes than the string has
        // UTF-16 chars. This check rejects a huge value with no array.
        if (path.isEmpty() || path.length() > MAX_PATH_BYTES) {
            return false;
        }
        int byteLength = path.getBytes(StandardCharsets.UTF_8).length;
        if (byteLength > MAX_PATH_BYTES) {
            return false;
        }
```

### MINOR 2 — `ParsedClick` prints the raw path and the raw host

`kit/jvm-core/src/main/java/octometer/kit/core/ingest/ParsedClick.java:13` and
`kit/jvm-core/src/main/java/octometer/kit/core/ingest/ParsedIngestRequest.java:9`

Both records use the default record `toString()`. My run printed:

```
ParsedIngestRequest[sessionId=3fa85f64-5717-4562-b3fc-2c963f66afa6, clicks=[ParsedClick[
element=a.one, ageMs=1, path=/users/4242/reset/S3CR3TV4LUE, referrerHost=null]]]
```

`IngestEvent` and `StoredEvent` each override `toString()` for exactly this reason
(design decision D15). Nothing logs these two records today, so the risk is a later debug
line inside the package. This branch adds the two raw values, thus this branch must also
override the two methods.

**The exact fix.** Give `ParsedClick` the same shape of override:

```java
    /**
     * Returns a text with no personal data (design decision D15). The
     * path and the referrer host stay out of this text, because a log
     * line must never hold a raw client value (rules C39, C41).
     */
    @Override
    public String toString() {
        return "ParsedClick[element=" + element + ", ageMs=" + ageMs + "]";
    }
```

Do the same for `ParsedIngestRequest`, and leave the session id out of its text.

### MINOR 3 — the two new public methods throw a NullPointerException for null

`EventFieldValidator.java:119` and `EventFieldValidator.java:182`

`isValidPath(null)` and `matchReferrerHost(null)` each throw `NullPointerException`.
`IngestPipeline` guards both calls, so no client input reaches them. Each method is
public, and `validateUserId` accepts `null`, thus the module is not consistent.

**The exact fix.** Add one sentence to each Javadoc: "The value of {@code path} must not
be {@code null}." Or return `false` and `null` for a `null` argument.

### MINOR 4 — the Google pattern runs for a host that can never match

`EventFieldValidator.java:189`

`GOOGLE_HOST_PATTERN` runs for every host-name value. The worst 253-byte value that I
found costs 29.6 us. A 16 KB body with 50 `octo:session-start` entries then costs 818.3 us
against 99.7 us on `main`. The pattern needs the literal text `google.`, thus a cheap test
before the pattern keeps the same result and removes the cost.

**The exact fix.**

```java
        if (matchesSourceEntry(rawValue, "google.com")
                || (rawValue.indexOf("google.") >= 0 && GOOGLE_HOST_PATTERN.matcher(rawValue).matches())) {
            return "google.com";
        }
```

### MINOR 5 — the order of rule C33 and rule C38 is not recorded

`kit/jvm-core/src/main/java/octometer/kit/core/ingest/IngestPipeline.java:73`

`validateElement` runs before the C38 check. An element `octo:session-start ` (with a
trailing space), `octo:session-start\u0000`, or an element with a Unicode look-alike of a
letter breaks rule C4 first, thus rule C33 makes the whole body invalid and the server
answers 400. Rule C38 alone would drop that one entry and keep the batch. The branch
rejects the whole body, which is the safe order. The contract states no order.

**The exact fix.** Add a comment above line 73:

```java
            // Rule C33 runs before rule C38. An element that breaks rule
            // C4 makes the whole body invalid (400), also when it starts
            // with the reserved prefix. Rule C38 then drops only an
            // entry whose element is legal under rule C4.
            EventFieldValidator.validateElement(click.element());
```

Please also record the order in rule C38 or rule C33 of `contract/README.md`, in a later
issue. The Java parser reviewer owns the reading of the rule.

## Two notes, with no severity

- `/%00`, `/%2F`, `/%2f%2e%2e`, `/%5Cevil`, and `/../../etc/passwd` each pass the shape
  check. Rule C39 allows them, and rule C42 stops them before a store: "The server decodes
  no `%` escape before the match", and "A path with an empty segment, a `.` segment, or a
  `..` segment gives `/other`". Issue #104 must never use a `path` value as a file path or
  as a redirect target.
- A `path` of a JSON type other than a string is dropped with no warning, because
  `parseOptionalStringValue` returns `null` for it, the same as an absent field. Rule C41
  covers a value that breaks a shape rule, thus I read this as correct. The Java parser
  reviewer owns the final reading.

I removed my clone after this review. I changed no file of the branch.

xamcross:
**Maintainer decision on MAJOR 1 of the two reviews (2026-09-22).**

The two reviews name one defect: `EventLogStore.append` gives the raw client `path` to each store. The cause is the issue text of the maintainer: step 5 and the first criterion of #103 sent the value to the store, against contract rules C39 and C42. The implementer followed the issue text. The contract has priority.

The decision is stronger than a new name plus a Javadoc warning:
- `IngestEvent` keeps one component `path`. It holds only the match result of the server. It stays empty in this pull request, because the kit has no route pattern list before #104.
- The raw client path, after the shape check of rule C39, stays in the parsed click only. No store gets it, thus no app store can write it.
- #104 runs the matcher in the pipeline and fills the component. #105 writes the component as it is.
- A new criterion of #103 asserts this with a test store. The bodies of #103, #104, and #105 hold the change, each with a comment.

A new agent does the correction round. The body of the first commit holds the false sentence "no store gets the raw path". A pushed commit stays as it is, thus the maintainer sets the text of the squash commit by hand at the merge.

xamcross:
## Correction round 1

I read the maintainer's decision of 2026-09-22 on this defect. I made the
fix that the decision states, and not the fix of either review, per that
decision. I did not rename `IngestEvent.path()` to `clientPath()`.

New commits: `35148b3`, `af93352`, `da7f938`, on top of a merge commit
`b48116f` of `origin/main` (no rebase). `./gradlew build` passed after the
merge, with Docker running for the `kit/jvm-mongo` container tests.

## MAJOR findings

| Finding | File and line | Result |
|---|---|---|
| MAJOR 1 (both reviews): `EventLogStore.append` gave the raw client `path` to each store | `kit/jvm-core/src/main/java/octometer/kit/core/store/EventLogStore.java:37`, `kit/jvm-core/src/main/java/octometer/kit/core/ingest/IngestEvent.java:24` | Fixed, with the maintainer's own fix. `IngestPipeline.process` now always sets `path` to `null`; the raw client value stays in `ParsedClick` only. See commit `35148b3`. |
| MAJOR 2 (Java review): `matchReferrerHost` held two literals, not the source list | `kit/jvm-core/src/main/java/octometer/kit/core/ingest/EventFieldValidator.java:182` | Fixed. The method now loops over `REFERRER_HOST_SOURCE_LIST`. New test: `EventFieldValidatorReferrerHostTest.matchesEachEntryOfTheSourceListExactlyAndAsASubdomain`. See commit `af93352`. |

## MINOR findings

12 of 14 are fixed. 2 stay open, with a reason.

| # | Review | Finding | File and line | Result |
|---|---|---|---|---|
| 1 | Java | `contract/README.md` is not a test task input | `kit/jvm-core/build.gradle.kts:27` | Fixed. `contract/README.md` and `contract/examples/` are now inputs. Commit `da7f938`. |
| 2 | Java | No test caps the C38 warning at one for each batch | `IngestPipelinePathAndSourceTest.java` | Fixed. New test `aBatchWithThreeReservedElementsWritesAtMostOneWarning`. Commit `35148b3`. |
| 3 | Java | A wrong-type `path` or `referrerHost` value gives zero warnings | `IngestParser.java:279` | Not fixed. The review only asks to "consider" a change, with no exact fix. The current behavior matches rule C41 ("a maximum of one" warning, and zero is within that bound). A behavior change here is a decision for a separate issue. |
| 4 | Java | `StoredEvent` kept no 4-argument compatibility constructor | `StoredEvent.java:16` | Fixed. Added, for the same reason as the `IngestEvent` 3-argument constructor. Commit `35148b3`. |
| 5 | Java | Four `-ing` forms as a noun or a verb | `EventFieldValidator.java:18,116,175`, `IngestEvent.java:29` | Fixed. Each of the four is rewritten. Commits `35148b3`, `af93352`. |
| 6 | Java | A 29-word test comment joins two statements with "so" | `EventFieldValidatorReferrerHostTest.java:100` | Fixed. Split into two sentences. Commit `af93352`. |
| 7 | Java | Pull request text: an `-ing` form, and a sentence above 25 words | Pull request body | Fixed. The pull request body is corrected; see below. |
| 8 | Java | Pull request text claims a Javadoc note that `isValidPath` does not carry | Pull request body | Fixed. The claim is removed from the corrected pull request body. |
| 9 | Java | `contract/README.md` rows C32 and C41 read as out of date after this merge | `contract/README.md` | Not fixed. The contract is out of scope for this pull request, by the maintainer's own rule. A follow-up issue is for the maintainer to open. |
| 1 | Security | `isValidPath` allocates a byte array before it checks the length | `EventFieldValidator.java:120` | Fixed. A length check now runs first. Commit `af93352`. |
| 2 | Security | `ParsedClick` and `ParsedIngestRequest` print the raw path and the raw host | `ParsedClick.java:13`, `ParsedIngestRequest.java:9` | Fixed. Both override `toString()`. Commit `35148b3`. |
| 3 | Security | `isValidPath(null)` and `matchReferrerHost(null)` throw `NullPointerException` | `EventFieldValidator.java:119,182` | Fixed, with the Javadoc option of the two the review gave. Commit `af93352`. |
| 4 | Security | The Google host pattern runs for a host that can never match | `EventFieldValidator.java:189` | Fixed. A cheap text search now runs first. Commit `af93352`. |
| 5 | Security | The order of rule C33 and rule C38 is not recorded | `IngestPipeline.java:73` | Fixed, with a code comment. The contract stays unchanged. Commit `35148b3`. |

## The pull request text correction

The first pushed commit's body held the sentence "no store gets the raw
path: `StoredEvent` copies `referrerHost`, but not `path`, from
`IngestEvent`." That sentence was false, because `EventLogStore.append`
read the raw client path straight from `IngestEvent.path()`. I did not
amend that commit. The correction is in the body of commit `35148b3`, and
the pull request description now states the corrected rule.

## Tests

`./gradlew :kit:jvm-core:test`: `BUILD SUCCESSFUL`, 138 tests, 0 failures,
0 errors (132 before this round, 6 new). `./gradlew build`: `BUILD
SUCCESSFUL` for the whole repository (`kit/jvm-core` 138, `kit/jvm-mongo`
52, `kit/jvm-ktor` 28, `monitor:backend` 66 tests).

I proved that each new test, and the changed test, can fail: I broke the
matching code for a moment, ran the one test class, saw the named test
fail, then restored the code. I did this for: the MAJOR 2 test, the C38
warning cap test, the store test of the new criterion (with two separate
breaks: one in `IngestPipeline`, one in `ParsedClick.toString()`), the
`StoredEvent.of` path-copy test, the `StoredEvent.of` null-path test, and
the `StoredEvent` 4-argument constructor test. I also broke a source-list
entry inside `contract/README.md` and confirmed that `./gradlew
:kit:jvm-core:test`, with no `--rerun-tasks` flag, ran again and caught
the drift; I then restored the file.

## CI

All checks of pull request #131 pass: Frontend, JVM modules, Monitor
backend on Windows, `.github/dependabot.yml`, Secret scan, Tracker.

xamcross:

I made a new clone, checked out the head of `issue-103-parser-path-source`
(`da7f938`), and read the diff of the three new commits only (`35148b3`, `af93352`,
`da7f938`) on top of the merge `b48116f`. I ran 24 attack tests of my own again. Each one
passes. Every finding of my first review is closed.

## Commands

- `./gradlew :kit:jvm-core:test` (JDK 21, `C:/Users/xamcr/.jdks/ms-21.0.9`):
  `BUILD SUCCESSFUL`. 138 tests of the branch, 0 failures, 0 errors, 0 skipped.
- The same task with my 24 attack tests added: `BUILD SUCCESSFUL`, 162 tests,
  0 failures, 0 errors.

## MAJOR 1 — closed

The maintainer decision is stronger than my fix, and it holds. I ran the app store probe
again with the path `/users/4242/reset/S3CR3TV4LUE`. Inside `append`, the probe read each
record component by reflection, the `toString()` text of the event, of the event list, and
of the `StoredEvent`, and the captured log. Then it searched each text for
`S3CR3TV4LUE`, for `/users/`, for `4242`, and for `reset`.

```
app store components: [sessionId=3fa85f64-..., element=a.one, ts=2026-09-21T10:15:29.999Z,
  path=null, referrerHost=null, stored.sessionId=3fa85f64-..., stored.element=a.one,
  stored.ts=2026-09-21T10:15:29.999Z, stored.userId=user-1, stored.path=null,
  stored.referrerHost=null]
app store texts: [IngestEvent[element=a.one, ts=2026-09-21T10:15:29.999Z],
  [IngestEvent[element=a.one, ts=2026-09-21T10:15:29.999Z]],
  StoredEvent[element=a.one, ts=2026-09-21T10:15:29.999Z, userId=<redacted>]]
```

The store sees no part of that text: not in a component, not in a `toString()` text, and
not in a log line. `IngestEvent.path()` is `null`, and `StoredEvent.path()` is `null`. The
`InMemoryEventLogStore` holds `null` too.

I also searched each exception message for the path. Four bodies with the secret path gave
`BODY_TOO_LARGE`, `INVALID_JSON`, `SESSION_ID_NOT_UUID`, and `ELEMENT_PATTERN`. Each
message is fixed text. No message holds the path, and no message holds the session id.

The raw client value stays in `ParsedClick` only. My test asserts that
`IngestParser.parse(body).clicks().get(0).path()` still equals the raw value, so the shape
check of rule C39 still sees it, and the warning of rule C41 still works.

## The 5 MINOR findings

| --- | --- | --- |
| 1. `isValidPath` allocates before the length check | Closed | The length gate is at `EventFieldValidator.java:126`. A call with 1 000 000 characters now takes 0.0002 ms, against 0.398 ms in round 1. |
| 2. `ParsedClick` prints the raw path and the raw host | Closed | `ParsedClick[element=a.one, ageMs=1]` and `ParsedIngestRequest[clicks=[ParsedClick[element=a.one, ageMs=1]]]`. No raw path, no raw host, no session id. |
| 3. A `null` argument gives a NullPointerException | Closed | Both methods still throw, and each Javadoc now states "The value of {@code path} must not be {@code null}". That is the option that I offered. |
| 4. The Google pattern runs for a host that can never match | Closed | The guard `rawValue.indexOf("google.") >= 0` is at `EventFieldValidator.java:212`. A 16 KB body of 50 hostile hosts now costs 194.0 us, against 818.3 us in round 1 (`main`: 99.7 us). |
| 5. The order of rule C33 and rule C38 is not recorded | Closed | The comment is at `IngestPipeline.java:76`. |

## The 24 attack tests, round 2

Each result matches round 1, with the new rule that `IngestEvent.path()` is always `null`.

- `path`, 34 forms: each bad form drops the field and keeps the entry. A 150-byte path,
  `/%00`, `/%2F`, `/%2f%2e%2e`, `/../../etc/passwd`, and `/%5Cevil` pass the shape check
  of rule C39, and each one still gives `path=null` in the event record. No form gives a
  class other than `IngestException`. A raw control character gives `INVALID_JSON`, which
  is broken JSON. A path of 1 000 000 characters and an array of 10 000 items each give
  `BODY_TOO_LARGE` from rule C18.
- A fuzz run of 200 000 random strings through `isValidPath`: no throw. Each accepted
  value holds only characters of the C39 set.
- `referrerHost`, 21 forms: the stored value is always `google.com`, `bing.com`, or
  `other`, or the field is absent. A fuzz run of 300 000 random values proves the closed
  set. The new ordered list of `af93352` keeps each result: `www.google.de` and
  `x.google.co.uk` give `google.com`, `evil.bing.com` gives `bing.com`,
  `attacker.google.top` gives `other`.
- The reserved prefix: `octo:foo`, `OCTO:foo`, `Octo:Session-Start`, and `octo:` each drop
  one entry, and the two other entries stay. A batch of 50 reserved entries gives 0 events
  and 1 warning; `ingest` then appends nothing, and the Ktor route answers 204.
- The warnings: a batch with 41 bad values gives 21 events and 2 warnings, one for rule
  C38 and one for rule C41. 10 000 bad-path batches give 10 000 log lines in 51 ms. No log
  line holds a raw path, a raw host, a session id, or a line break. A host value with
  `\nINJECTED WARNING` writes no injected line.
- The user id: a `userId` field of the body, of a click entry, or beside a `path` field has
  no effect.
- Smuggling: the key `"\u0070ath"` routes to the same shape check, and a duplicate of it
  gives `DUPLICATE_FIELD`. The body limit stays at 16 384 bytes.
- The cost of a legal batch of 50 entries with a 150-byte path: 108.7 us for each request,
  against 117.3 us in round 1.

## No new finding

The three commits add no new risk. `build.gradle.kts` adds the contract document and the
example folder as task inputs, which only makes the test task run again after a contract
edit. `StoredEvent` gains a `path` component that is always `null` today, and its
`toString()` leaves the path out. The 4-argument `StoredEvent` constructor keeps an older
caller working, and it sets both new fields to `null`.

I found no BLOCKER and no MAJOR finding. The branch is safe for a merge from the view of
application security. Please keep issue #33 (the rate limit, `production-blocker`) before
the public route starts: each rejected `path` still writes one warning line for each
request.

I deleted my clone after this review. I changed no file of the branch.

xamcross:
**Review (second review, Java parser engineer): APPROVE**

I read the maintainer decision of 2026-09-22 first. I judge the correction against that
decision, and not against the fix of my first review. I made a new clone, and I checked out the
head `da7f938`. I read the diff of `b48116f..da7f938`.

## The commands that I ran

| Command | Result |
|---|---|
| `./gradlew :kit:jvm-core:test` (JDK 21) | `BUILD SUCCESSFUL`. **138 tests, 0 skipped, 0 failures, 0 errors** over 16 test suites. |
| `./gradlew build` (one run, whole repository) | `BUILD SUCCESSFUL in 1m 38s`. `kit/jvm-core` 138, `kit/jvm-mongo` 26 (9 skipped), `kit/jvm-ktor` 28, `monitor:backend` 66. No failure, no error. |

`kit/jvm-mongo` and `kit/jvm-ktor` both compile and pass. Each one builds `new
IngestEvent(sessionId, element, ts)` or overrides `append(List<IngestEvent>, String)`, and the
3-argument constructor keeps each call site correct.

---

# MAJOR 1 — FIXED. No store can get the raw client path.

I did not trust the new test of the pull request. I wrote my own test in my clone. It sends a
body with a raw path `/orders/tok-4f2a9b/receipt` and a raw host
`intranet.employer-name.example`, on a session start entry and on a click entry. It reads
**every record component by reflection**, so a component that a later commit adds also comes
under the check. All four of my tests pass.

**Each way out of the parser, with my measured output.**

| The way out | What I measured |
|---|---|
| `IngestEvent` components | `[sessionId=..., element=octo:session-start, ts=..., path=null, referrerHost=other]` |
| `IngestEvent.toString()` | `IngestEvent[element=octo:session-start, ts=2026-09-21T10:15:30Z]` |
| `StoredEvent` components | `[sessionId=..., element=..., ts=..., userId=user-1, path=null, referrerHost=other]` |
| `StoredEvent.toString()` | `StoredEvent[element=..., ts=..., userId=<redacted>]` |
| `InMemoryEventLogStore.events()` | Each `StoredEvent` holds `path=null` |
| `ParsedClick.toString()` | `ParsedClick[element=octo:session-start, ageMs=0]` |
| `ParsedIngestRequest.toString()` | `ParsedIngestRequest[clicks=[ParsedClick[element=..., ageMs=0], ...]]` |
| Each log line of a batch with a bad path, a bad host, and a reserved element | Two constant texts only. Neither one holds the path, the host, the session id, or a client element name. |
| Each exception message (`ELEMENT_PATTERN`, `SESSION_ID_NOT_UUID`, `NEGATIVE_AGE_MS`, `DUPLICATE_FIELD`, `INVALID_JSON`) with a raw path in the body | Each message is a constant text. No message holds the path or the session id. |
| A public accessor | `ParsedClick`, `ParsedIngestRequest`, and `IngestParser` are all package-private. An app outside `octometer.kit.core.ingest` cannot name them. |

`IngestPipeline.java:106` sets `String path = null;` for every click. The shape check of rule
C39 still runs, only to find an invalid value for the rule C41 warning. The raw value stays in
the `ParsedClick` and it goes no further. This matches the maintainer decision word for word,
and it matches rule C39 ("The server never stores the raw client value") and rule C42 ("Without
a route pattern list, the server stores no `path` field").

The `referrerHost` of my raw internal host became the literal `other`, never the raw host. That
is rule C40. Good.

## Point 3 — my judgement of the seam for #104

The seam is good. #104 changes one line: `IngestPipeline.java:106`, from `String path = null;` to
the match result of the route pattern list. Nothing else in the chain needs a change:

- `EventLogStore.append` keeps its signature.
- `StoredEvent` already holds the `path` component, and `StoredEvent.of` already copies it. I
  proved the copy with my own break (see the table below).
- `InMemoryEventLogStore` needs no change.
- #105 then adds `.append("path", event.path())` to `MongoEventLogStore.toDocument`, and that
  value is a match result by construction.

Two small notes, and neither one is a defect:

1. `IngestPipeline.process(String rawBody, Clock clock)` takes no settings. #104 needs the route
   pattern list here, so it adds a third parameter or a new component on `IngestSettings`. Both
   are small, and this module already has the compatibility-constructor pattern for a record.
2. The correction flattened the old nested `if` block into one condition. #104 must open the
   block again, because it needs the shape-check result for the match. This costs #104 three
   lines. I state it so that the author of #104 expects it.

# MAJOR 2 — FIXED. The match reads the ordered source list.

`EventFieldValidator.java:46` now holds `List.of("google.com", "bing.com")`, and
`matchReferrerHost` loops over it. `referrerHostSourceList()` returns a set copy.

**My proof.** I added a third entry `duckduckgo.com` to the list, then I ran the test class. The
new test `matchesEachEntryOfTheSourceListExactlyAndAsASubdomain` **passed** with three entries:
the method returned `duckduckgo.com` for `duckduckgo.com` and for `www.duckduckgo.com`. Only the
two closed-set tests failed, as they must. Before this correction the same experiment made the
match return `other` for the new entry, with no test failure. The drift is closed.

I also checked that the new order is behaviour-neutral. The loop now runs for both entries before
the Google pattern. No host can both end with `.bing.com` and match
`^([a-z0-9-]+\.)*google\.((com|co)\.[a-z]{2}|com|[a-z]{2})$`, so no result changes. The new
guard `rawValue.indexOf("google.") >= 0` is safe, because the pattern needs that literal text.

---

# My 9 MINOR findings of the first review

| # | Finding | State | My judgement |
|---|---|---|---|
| 1 | `contract/README.md` was not a test task input | **FIXED** | I edited the source list inside `contract/README.md` and ran `./gradlew :kit:jvm-core:test` with **no** `--rerun-tasks` flag. The task ran again and `theSourceListMatchesTheContractSourceList` FAILED. I also edited `contract/examples/ingest-valid-C39-path.json`, and `parsesTheValidPathFieldOfAClick` FAILED. The `inputs.dir` for the examples goes past my finding. Good. |
| 2 | No test capped the C38 warning at one | **FIXED** | `aBatchWithThreeReservedElementsWritesAtMostOneWarning`. I broke the guard and the test failed. |
| 3 | A wrong-type `path` value gives zero warnings | **OPEN** | I accept the reason. Rule C41 says "a maximum of one warning", so zero stays inside the bound. The behaviour is correct against the contract, and a change needs an owner decision. A separate issue is the right place. |
| 4 | `StoredEvent` kept no 4-argument constructor | **FIXED** | The constructor is back, and `theFourArgumentConstructorGivesNoPathAndNoReferrerHost` covers it. I broke it and the test failed. |
| 5 | Four `-ing` forms as a noun or a verb | **FIXED** | Each of the four lines that I named is rewritten. I found one more that I missed; see MINOR A below. |
| 6 | A 29-word test comment | **FIXED** | `EventFieldValidatorReferrerHostTest.java:100` is now two sentences. |
| 7 | The pull request text held an `-ing` form and a long sentence | **FIXED** | The text now reads "each call site still compiles with no change." |
| 8 | The pull request text claimed a Javadoc note that `isValidPath` does not carry | **FIXED** | The claim is gone. |
| 9 | Rows C32 and C41 of the contract example table read as out of date | **OPEN** | I accept the reason. The contract belongs to #101 and #122, and this pull request must not touch it. The maintainer opens the follow-up. |

## The breaks that I ran, to prove the new tests can fail

| I changed | The tests that failed |
|---|---|
| `IngestPipeline.java:106`, `String path = null;` to `String path = click.path();` | 11 tests, `aStoreNeverReceivesTheRawClientPathForAValidPath`, `aValidPathPassesTheShapeCheckButDoesNotReachTheEventRecord`, and my own store test included |
| `ParsedClick.toString()` prints the path | `aStoreNeverReceivesTheRawClientPathForAValidPath`, and my own log-and-text test |
| `StoredEvent.of` passes `null` in place of `event.path()` | `ofCopiesThePathFromTheIngestEvent` |
| The rule C38 warning guard removed | `aBatchWithThreeReservedElementsWritesAtMostOneWarning` |
| The 4-argument `StoredEvent` constructor sets a path | `theFourArgumentConstructorGivesNoPathAndNoReferrerHost` |
| A third entry in the source list | The two closed-set tests only; the new match test passed |
| The source list inside `contract/README.md` | `theSourceListMatchesTheContractSourceList`, with no flag |

Each new test can fail. No test is empty. No test is always green. I restored each file after
each experiment, and my clone held no change at the end.

---

# Two new MINOR findings

Neither one prevents this approval.

**MINOR A — one `-ing` form stays.** `EventFieldValidator.java:16` and `:17`: "the two new
methods return a value instead of **throwing**". I named three other lines of this file in my
first review, and the author fixed each one. I missed this one. Fix:

```
 * Unlike {@link #validateElement}, {@link #validateSessionId}, and
 * {@link #validateUserId}, the two new methods return a value. They do
 * not throw: rule C41 makes an invalid `path` or `referrerHost` value
 * non-fatal, and it does not stop the whole request.
```

**MINOR B — one test count in the pull request text.** The text says `kit/jvm-mongo` has 52
tests. My own run gives 26 tests with 9 skipped, over 5 test suites. Please correct the number,
or state how you counted it. The result itself is right: 0 failures and 0 errors.

---

# The other checks

- **Scope (PASS).** The three new commits touch 11 files, all under `kit/jvm-core`. There is no
  route pattern match (#104), no MongoDB field (#105), and no rate limit. `contract/README.md`
  stays unchanged.
- **Secrets (PASS).** No credential, no database file, and no build output. Each host name and
  each UUID is fake or public.
- **Strict and safe (PASS).** The new pre-check in `isValidPath` (`path.isEmpty() ||
  path.length() > 150`) is correct: a value of 151 chars always needs 151 bytes or more in
  UTF-8, so the early return never rejects a legal value. The byte check still runs for a value
  of 150 chars or fewer.
- **Design decision D15 (PASS).** Four records now hide each raw value in `toString()`:
  `IngestEvent`, `StoredEvent`, `ParsedClick`, and `ParsedIngestRequest`.
- **The comment on rule C33 and rule C38** (`IngestPipeline.java:76`) records the order that I
  measured in my first review. It is correct and it helps a later reader.

No BLOCKER stays. No MAJOR stays. Two MINOR findings are new, and two of my first nine stay open
with a reason that I accept. I approve.

I cannot use `gh pr review --approve`, because the pull request comes from the same account.

PR #130, opened 2026-09-21
feat(tracker): add the route pattern match and the path of a click

xamcross:
## Summary

The tracker gets a new option, `routes`. With this option, each click entry holds `path`: the matched route pattern, with a `*` segment kept as the real segment. The matcher is a pure function in its own file, `kit/tracker/src/path-match.ts`, with its own test file. The tracker reads `location.pathname` at the time of the click, not at the time of a later flush. The tracker also drops a `data-octo` value that starts with `octo:` (contract rule C38), and it fails closed on a bad `routes` value.

## Acceptance criteria

- [x] Without the option `routes` no entry holds `path`. Test: `adds no path field to a click entry without the routes option`.
- [x] `/history/42` gives `/history/:id`. Case in `contract/examples/C42-path-match-cases.json` and test `adds the matched path field to a click entry with the routes option`.
- [x] `/articles/ovdp-2026` gives `/articles/ovdp-2026`. Case `a star segment keeps a valid real value`.
- [x] `/ARTICLES/` gives `/articles`. Case `a trailing slash goes before the match, and the match ignores the letter case`.
- [x] The stored value uses the text of the pattern, never the text of the input. Case `uses the text of the pattern for a literal segment, never the text of the input`.
- [x] `/tokens/abc` gives `/other`. Case `a path with no matching pattern`.
- [x] `/articles/a/b` gives `/other`. Case `a path with a different segment count than each pattern`.
- [x] A `*` segment of 81 characters gives `/other`. Case `a star segment of 81 characters` (and an 80-character case that stays valid).
- [x] `/articles//x`, `/articles/.`, and `/articles/..` give `/other`, and the rule also holds at a `:name` position. Six cases in the shared file, plus the mutation proof of correction round 1 (BLOCKER 1).
- [x] The tracker never decodes a `%` escape before the match. Case `never decodes a percent escape before the match`.
- [x] A result of 150 bytes stays, a result of 151 bytes becomes `/other`, and a percent-encoded Cyrillic path above 150 bytes becomes `/other`. Cases `a result of exactly 150 bytes stays`, `a result of 151 bytes gives /other`, and `a percent-encoded path above 150 bytes at a full route match`.
- [x] The match cases are in `contract/examples/C42-path-match-cases.json`, the tracker test reads that file, and the rule table of `contract/README.md` has a row for it. `src/path-match.test.ts:3` reads the file. `contract/README.md:260` names it, with a new sentence on the scope of the file (correction round 2).
[...]

PR #143, opened 2026-09-21
feat(tracker): add the page lifecycle flush and the retry rule

### LINKED ISSUE #36: Add the page lifecycle flush and the retry rule to the tracker

xamcross:
## Summary

The tracker now sends its queue before the browser hides or unloads the page. It retries a failed batch one time, after a wait, and it never sends a batch after `stop()`. It splits a large batch into two or more requests, each below the body size limit.

A new file, `kit/tracker/src/batch.ts`, holds the pure request-split logic, with its own unit tests. `kit/tracker/README.md` states the keepalive quota as a sum, and what a retry can cause.

**Correction round 1** (this update): a run token stops a late retry after `stop()`. The retry now waits before it fires. The `pagehide` flush and the `visibilitychange` flush now both send a batch one time only; only the timer flush retries. A new test proves the 64 KB keepalive sum with 200 real entries, no spy. Six minor findings of the two reviews are fixed. See the PR comment for the full list.

## Acceptance criteria (issue #36, edit of 2026-09-22)

- [x] Each rule of steps 1 to 6 has a test with fake timers. Evidence: `kit/tracker/src/index.test.ts` (85 tests) and `kit/tracker/src/batch.test.ts` (9 tests), each with `vi.useFakeTimers()`.
- [x] A failed request throws no error into the page. Evidence: `sendBatch` ends each attempt in a `.catch()`. `flush()` runs inside `safeFlush()`. Each split batch has its own try/catch (MINOR 4 of the reliability review).
- [x] No second send after `stop()`. Evidence: a run token (`runId`) that `start()` and `stop()` each raise. `sendBatch` and `waitThenRetry` check the token before the second send and before the retry timer fires. Tests: "sends no retry after stop(), also when the first request settles after stop()", "sends no retry after stop() runs during the wait before the retry".
- [x] A test proves the second send waits for the delay, with fake timers, not before it. Evidence: "waits the full delay before it sends a batch again after a network error" and "...after a 429 response", each checking the call count one ms before, and at, the deadline.
- [x] The keepalive body of one lifecycle flush stays below 64 KB, as a **sum**, not as one body. Evidence: "holds each request and the whole keepalive flush below the limits, with 200 entries at the contract maximum" (`index.test.ts`), a real `pagehide` flush, no spy, measuring the sum of all four requests (near 58 KB).
- [x] No request body passes 15 000 bytes. Evidence: `batch.ts`, `splitIntoRequestBatches`, and the same 200-entry integration test above, which asserts every request body below that limit.
- [x] A batch whose encoded body passes 15 000 bytes goes in two or more requests, each below that limit. Evidence: `batch.test.ts` "splits a batch whose encoded body passes 15 000 bytes, each request below that limit".
- [x] 50 entries with a 150-byte `path` each give a body below the 16 KB limit of rule C18. Evidence: `batch.test.ts` "keeps 50 entries with a 150-byte path each in one request, below the 16 KB limit of rule C18" (14 563 bytes measured).
- [x] The `pagehide` flush, and the `visibilitychange` flush, each send a batch one time only. Only the timer flush retries. Evidence: "the single send of the lifecycle flush" describe block, and "keeps the retry of the timer flush, unlike the lifecycle flush".
- [x] `npm test` in `kit/tracker` passes. Evidence: 132 tests passed (see the issue comment for the earlier 121-test record; this round adds 11 tests).
- [x] A comment on this issue records the output of `npm test`. Evidence: posted on 2026-09-21, and referred to again in the PR comment of this round.

## The retry rule

One retry after a network error, a 5xx response, or a 429 response, after a wait: 500 ms plus a random value below 500 ms for a network error or a 5xx, 2000 ms plus a random value below 2000 ms for a 429. A drop after each other 4xx response. Never a third send. The `pagehide` flush and the `visibilitychange` flush each send a batch one time only, with no retry, because a browser can freeze the page after the hidden state and drop the connection. Only the timer flush retries.

## What a retry can cause

A maintainer comment on this issue states that the MongoDB store of the app writes a batch with one ordered `insertMany` call, without a transaction. A retry after a mid-batch failure can write the first part of the batch again, with new `_id` values that the monitor cannot drop as copies. This means a retry after a 5xx response can count some clicks two times. `kit/tracker/README.md`, section "The retry rule", states this for an app team. This issue adds no client-side event id; a fix needs a contract change, outside this issue.

## Test count and bundle size (correction round 1)

- `npm test`: 131 tests passed (3 test files: `index.test.ts`, `path-match.test.ts`, `batch.test.ts`).
- `npm run typecheck`: no error.
- `npm run build`: no error.
- Bundle size (gzip), `dist/*.js`, no source maps:
  - `index.js`: 11 198 bytes raw, 2 889 bytes gzip.
  - `path-match.js`: 4 047 bytes raw, 1 197 bytes gzip.
  - `batch.js`: 1 605 bytes raw, 577 bytes gzip.
  - Combined: 16 850 bytes raw, 4 663 bytes gzip.
- `npm pack --dry-run`: 17 files, 21.5 kB package, 84.0 kB unpacked, no test file.

## Decisions of this round

1. **The retry wait.** Decided by the maintainer on 2026-09-22: 500 ms plus jitter below 500 ms for a network error or a 5xx, 2000 ms plus jitter below 2000 ms for a 429. Written into design decision D24, beside the retry sentence.
2. **The `visibilitychange` flush no longer retries.** The maintainer edit of 2026-09-22 restricts both lifecycle listeners to one send. Only the timer flush, which runs while the page stays visible, retries.
3. **A byte-based split rarely triggers with contract-valid data.** Unchanged from the first round: the contract caps `element` at 100 characters (rule C4) and `path` at 150 bytes (rule C39). The new 200-entry integration test proves the invariant `MAX_QUEUE_SIZE / MAX_BATCH_ENTRIES * MAX_BODY_BYTES < 65536` holds for the contract maximum, with no spy.

## Files not committed

None. `dist/` and `node_modules/` stay out of the commit, per the existing `.gitignore`.

Closes #36

Generated with Claude Code


xamcross:
**Review (TypeScript library engineer): CHANGES REQUESTED**

I made my own clone, and I ran `gh pr checkout 143` in it (commit `f797430`). I ran each command
myself with Node v24.21.0 and npm 11.19.0. The work is careful, and the six steps each hold under a
mutation test. Two MAJOR findings stay open.

## The commands that I ran

| Command | Result |
| --- | --- |
| `npm ci` | OK, 0 vulnerabilities |
| `npm test` (three times) | `Test Files 3 passed (3)`, `Tests 121 passed (121)` each time |
| `npm test` with a shared jsdom (`--no-isolate`) | 121 passed |
| `npm test` with a shuffled order (seeds 7, 11, 42, `--no-isolate`) | 121 passed each time |
| `npm test` with the files in the reverse order | 121 passed |
| `npm run typecheck` | no error |
| `npm run build` | no error, 12 files in `dist/` |
| `npm pack --dry-run` | 17 files, 18.8 kB package, 72.9 kB unpacked, no test file |

Bundle size, from my own build, with `gzip -9`: `index.js` 9 846 raw / 2 661 gzip, `path-match.js`
4 047 / 1 215, `batch.js` 921 / 402. The three files together: 14 814 raw / 3 803 gzip.

## The acceptance criteria

- **Each rule of steps 1 to 6 has a test with fake timers: PASS.** I broke the code 18 times, one
  time or more for each step. Each mutation makes a test fail. The table is below.
- **A failed request throws no error into the page: PASS.** I ran four mixed failure orders that the
  test file does not hold: a 500 and then a network error, a network error and then a 503, a 500 and
  then a 400, and a 302. Each one sends 2, 2, 2, and 1 requests. No run gave an unhandled rejection.
- **The keepalive body stays below 64 KB: PASS, with a thin margin.** See MAJOR 2.
- **No request body passes 15 000 bytes: PASS for each value that the tracker can build.** The
  element holds a maximum of 100 characters (`index.ts:50`), and the path holds a maximum of 150
  bytes (`path-match.ts:25`). See MINOR 1 for the one case outside that set.
- **A batch above 15 000 bytes goes in two requests, each below the limit: PASS at the unit level**
  (`batch.test.ts:40`). See MAJOR 2 for the integration level.
- **50 entries with a 150-byte path give a body below 16 KB: PASS.** I measured 14 563 bytes myself,
  through the public API, with a 100-character element and a 150-byte path.
- **`npm test` passes: PASS.** 121 tests, three runs.
- **A comment records the output of `npm test`: PASS.** The comment of 2026-09-21 on #36 records the
  same count, 121.

## The mutation test

Each line names one mutation of the source and the count of tests that fail for it. No mutation
stays green.

| # | Mutation | Failing tests |
| --- | --- | --- |
| 1 | Remove the `pagehide` listener (step 1) | 4 |
| 2 | Flush on `visible` in place of `hidden` (step 2) | 3 |
| 3 | `NORMAL_RETRY_COUNT = 0` (step 3) | 6 |
| 4 | Retry each 4xx response (step 3) | 1 |
| 5 | `NORMAL_RETRY_COUNT = 2`, a third send (step 3) | 2 |
| 6 | `PAGEHIDE_RETRY_COUNT = 1` (step 4) | 2 |
| 7 | `MAX_QUEUE_SIZE = 201` (step 5) | 1 |
| 8 | `queue.pop()` in place of `queue.shift()` (step 5) | 1 |
| 9 | `MAX_BODY_BYTES = 1e9`, no byte split (step 6) | 2 |
| 10 | Measure the entries only, without the frame and the session id (step 6) | 1 |
| 11 | `MAX_BATCH_ENTRIES = 51` (rule C17) | 2 |
| 12 | Remove `referrerPolicy: 'no-referrer'` (#130) | 1 |
| 13 | `Date.now()` in place of `performance.now()` (#130) | 2 |
| 14 | The `routes` option fails open with a bad entry (#130) | 2 |
| 15 | Remove the `try/catch` around the matcher (#130) | 1 |
| 16 | `keepalive: false` on the lifecycle flush | 1 |
| 17 | `stop()` keeps the click listener | 29 |
| 18 | `start()` without the guard against a second call | 20 |

Step 5 has no new test, and the author says so. I judge the old test
(`index.test.ts:343`) as good: mutation 7 and mutation 8 each make it fail. It holds the count, the
oldest entry, and the newest entry.

Mutation 17 and mutation 18 also answer the question of a leaked listener: a leak of the click
listener breaks 29 tests at once. The suite stays green with a shared jsdom document, with three
shuffled orders, and with the files in the reverse order. I found no leak of a listener, of a fake
timer, of `document.visibilityState`, or of the global `fetch` stub.

## The new module `src/batch.ts`

I ran my own tests against the module.

- The measure is exact. For a Cyrillic path, an emoji path, and a lone surrogate,
  `encodedBodyBytes` gives the same number as the UTF-8 byte count of the real request body: 130,
  117, 111, and 117 bytes. `JSON.stringify` writes an escape for a lone surrogate, so the measured
  string and the sent string stay equal. The key order of `batch.ts:63` and `index.ts:293` is the
  same, thus the measure and the body never differ.
- 60 entries with a 300-character Cyrillic path give 3 requests of 14 763, 14 783, and 9 023 bytes.
  No entry is lost. No entry goes into two requests. No request is empty.
- No loop without an end: the `for` loop is bounded by the click count, and the retry is bounded by
  `retriesLeft`.

## Findings

### MAJOR 1 — a retry goes out after `stop()`

`kit/tracker/src/index.ts:286-313` (`sendBatch`), with `kit/tracker/src/index.ts:131`.

`stop()` is the consent switch of the tracker: design decision D24 starts the tracker only after the
consent signal, and `stop()` empties the queue and removes the `sessionStorage` key. A request that
is in flight keeps its own promise chain, thus the retry runs after `stop()`.

Evidence, from a test that I wrote and ran in my clone: one click, the flush timer fires, `fetch`
holds the request open, the app calls `stop()`, then the request fails with a network error. The
mock records **2 calls**. The second call starts after `stop()`, and it carries the click entry and
the session id that `stop()` removed from `sessionStorage`. No test in the pull request covers this.

Fix. Give each run a token, and stop a retry of an old run.

```ts
let runId = 0;
// in start(), after `started = true`:  runId += 1;
// in stop(), before the listener removal:  runId += 1;

function flush(options: FlushOptions): void {
  ...
  const token = runId;
  for (const batch of batches) {
    void sendBatch(id, batch, options.keepalive, options.retryCount, token);
  }
}

function sendBatch(id: string, batch: ClickPayload[], keepalive: boolean, retriesLeft: number, token: number): Promise<void> {
  ...
  const canRetry = retriesLeft > 0 && token === runId;
  // use canRetry in the .then() branch and in the .catch() branch
}
```

Add one test: "sends no retry after `stop()`". Hold the first `fetch` open, call `stop()`, reject
the request, and expect exactly one call.

### MAJOR 2 — no test measures the byte limit with real worst-case data

`kit/tracker/src/index.test.ts:1274-1295`.

The test "sends one request for each batch that the splitter gives, each request below 15 000 bytes"
replaces `splitIntoRequestBatches` with a spy that gives two batches of one small click. The loop
check is real, but the last assertion, `bodyBytes < 15000`, cannot fail: each mocked body holds about
100 bytes. The pull request names this test as the evidence for two acceptance criteria, and for
those two it shows nothing.

The margin is thin, and this is why the gap matters. I measured the worst case that the contract
allows, through the public API, with no spy: 200 entries, a 100-character element, a 150-byte path,
and a `pagehide` flush. The result is **4 requests of 14 563 bytes each, 58 252 bytes in total, with
`keepalive: true` on each one, and 200 entries in total**. One request keeps only 437 bytes below the
15 000-byte limit, and the group keeps 7 284 bytes below the 64 KB keepalive quota. A later field, or
a longer `ageMs` value, crosses one of the two limits, and no test in this pull request fails for it.

Fix. Keep the spy test for the loop, drop its byte assertion, and add one integration test with no
spy. This test runs green on the present code.

```ts
it('holds each request and the whole keepalive flush below the limits, with 200 entries at the contract maximum', () => {
  const longSegment = 'a'.repeat(149);
  window.history.pushState({}, '', `/${longSegment}`);
  const element = 'b'.repeat(100);
  tracker = createTracker({ endpoint: ENDPOINT, routes: [`/${longSegment}`], flushIntervalMs: 1000000 });
  tracker.start();
  for (let i = 0; i < 200; i += 1) {
    const host = document.createElement('div');
    host.setAttribute('data-octo', element);
    document.body.appendChild(host);
    clickElement(host);
  }
  vi.advanceTimersByTime(600000);
  window.dispatchEvent(new Event('pagehide'));

  const encoder = new TextEncoder();
  const sizes = fetchMock.mock.calls.map((call) => encoder.encode(String((call[1] as RequestInit).body)).length);
  const entries = parseCalls(fetchMock).reduce((sum, call) => sum + call.body.clicks.length, 0);
  expect(entries).toBe(200);
  for (const size of sizes) {
    expect(size).toBeLessThan(15000);
  }
  expect(sizes.reduce((sum, size) => sum + size, 0)).toBeLessThan(64 * 1024);
});
```

### MINOR 1 — one entry above the limit gives a request above the limit

`kit/tracker/src/batch.ts:48-53`, with `kit/tracker/README.md:154-158`.

A click whose own body passes 15 000 bytes forms its own group. I measured a request of 15 200 bytes
for such an entry. The README says each request body stays below the limit, and for this one case it
does not. The tracker cannot build such an entry today, because the element holds 100 characters and
the path holds 150 bytes, so I keep this at MINOR. The choice to keep the click is correct.

Fix. Add one sentence to the doc comment of `splitIntoRequestBatches` and to the README section: one
entry that passes the limit on its own goes out alone, above the limit, and the server answers 400
under rule C18.

### MINOR 2 — the retry after a 429 response has no delay

`kit/tracker/src/index.ts:299-301`.

The retry starts at once, inside the same promise chain. For a network error and for a 5xx response
this is acceptable. For a 429 response the server states a rate limit, and an immediate retry meets
the same limit and adds load. Decision 1 of the pull request states the open point, and the issue
names no delay, thus I do not block on it.

Fix. Open a follow-up issue for a delay before the retry, and for the `Retry-After` header.

### MINOR 3 — the README states the keepalive limit as a per-request rule

`kit/tracker/README.md:57-60`.

The browser counts the sum of the bodies of each keepalive request that is in flight for the
document against the 64 KB quota, not one body alone. One `pagehide` flush of a full queue starts
four requests at the same time. My measure gives 58 252 bytes in total.

Fix. State the rule as a sum, and give the worst case: a full queue of 200 entries gives four
requests and 58 252 bytes.

### MINOR 4 — the splitter encodes the whole candidate body for each click

`kit/tracker/src/batch.ts:44-53`.

`encodedBodyBytes` runs one time for each click, and each run encodes the whole group again. The cost
grows with the square of the entry count. A full queue of 200 entries encodes about 1.4 MB, and the
`pagehide` flush pays this cost while the page unloads.

Fix. Measure each entry one time, keep a running sum, and add the fixed frame cost of the session id
and of the separators.

### MINOR 5 — a stored session id passes without a check (before this pull request)

`kit/tracker/src/index.ts:347-354`.

`readStoredSessionId` gives the value of the `sessionStorage` key to the request with no check
against contract rule C5. A page that writes a long value under that key makes each request body pass
the 16 KB limit of rule C18, and the server answers 400 for each one. I measured two requests of
20 053 bytes each for a 20 000-byte stored id. The splitter cannot repair this, because the frame
alone passes the limit. This behaviour comes from an earlier issue, and it is outside the scope
here.

Fix. Open a follow-up issue: check a stored session id against rule C5, and create a new id when the
stored value breaks the rule.

### MINOR 6 — two README sentences pass the length rule

`kit/tracker/README.md:57-60` and `kit/tracker/README.md:154-158`.

The sentence about the 64 KB limit holds 27 words. The sentence about the split holds 38 words.
ASD-STE100 gives a maximum of 25 words for a description. The text "well under the browser limit" is
also a figure of speech.

Fix. Cut each of the two sentences into two sentences. Write "below" in place of "well under".

- The retry rule sends exactly two requests for each retryable failure order, and one request for a
  400 response and for a 302 response. A third send is not reachable: the inner call of the retry
  always settles, so the outer `.catch` cannot start a further send.
- The `pagehide` flush sends one time only. The `visibilitychange` flush keeps the retry. Decision 2
  of the pull request agrees with the issue text.
- `stop()` removes the click listener, the `pagehide` listener, and the `visibilitychange` listener.
  A second `start()` adds each one only one time.
- Each rule of #130 stays: the `routes` option fails closed, the guard of rule C38 drops each `octo:`
  value, the matcher runs inside a `try/catch`, `ageMs` uses the monotonic clock, and
  `referrerPolicy` is `no-referrer`.
- No console line holds a path, a pattern, a `data-octo` value, or a session id. The new code adds no
  console line. The warning of `prepareRoutes` names the index of the entry only.
- Decision 3 of the pull request is correct. I confirm the number 14 563 with my own measure, thus
  the byte split never runs with data that the contract allows.
- The public API adds no export and no option. `batch.ts` stays outside the `exports` map of the
  package, so an app cannot import it. `dist/batch.js` is in the tarball, so the import of
  `dist/index.js` resolves.
- The diff holds six files. It holds no secret, no database file, and no build output.
- The commit body and the code comments follow ASD-STE100.

Please fix MAJOR 1 and MAJOR 2. I approve after that.

xamcross:
**Review (browser reliability engineer): CHANGES REQUESTED**

I made my own clone. I ran `gh pr checkout 143` in it. I ran each command myself. I also ran the
built tracker in a real Chrome browser, against a local server that records each request. I
deleted the clone and I stopped the server.

## The acceptance criteria

|---|---|---|---|
| 1 | Each rule of steps 1 to 6 has a test with fake timers | PASS | I broke the code six times. Each break fails a test. See "The break test" below. |
| 2 | A failed request throws no error into the page | PASS | I broke four things: a rejected `fetch`, a `fetch` that throws at the call, an absent `fetch`, and a `JSON.stringify` that throws. No case throws into the page. No case writes a console line. |
| 3 | The keepalive body stays below 64 KB | PASS, with a wrong reason | Chrome, a full queue of 200 entries at the contract maximums: 4 requests, 14371 + 14463 + 14463 + 14463 = 57760 bytes. The quota is 65536 bytes. The margin is 7776 bytes. See MAJOR 2. |
| 4 | No request body passes 15 000 bytes | PASS, with one exception | `batch.ts:38`. One single click above the limit forms its own request. The tracker cannot make such a click today, because rule C4 caps `element` at 100 characters and `path-match.ts:25` caps `path` at 150 bytes. See MINOR 3. |
| 5 | A batch above 15 000 bytes goes in two requests | PASS | I raised `MAX_BODY_BYTES` to 60000. The test "splits a batch whose encoded body passes 15 000 bytes" fails. |
| 6 | 50 entries with a 150-byte `path` give a body below 16 KB | PASS | Chrome measured 14463 bytes. My own calculation gives 14513 bytes with a 5-digit `ageMs`. Both stay below 16384. |
| 7 | `npm test` in `kit/tracker` passes | PASS | My own run: `Test Files 3 passed (3)`, `Tests 121 passed (121)`. `npm run typecheck` gives no error. |
| 8 | A comment on the issue records the `npm test` output | PASS | The comment is on issue #36. |

No secret, no database file, and no build output is in the diff. The diff holds 6 files. The
language of the README, of each comment, and of the commit body follows ASD-STE100.

---

## BLOCKER

None.

---

## MAJOR 1 — the retry after a 429 lands in the same rate-limit window

**File:** `kit/tracker/src/index.ts:299-302`

**The defect.** The code retries at once, inside the same promise chain. Decision D20 gives the
rate limit a **fixed window of 60 seconds**. Decision D43 gives the anonymous key 300 requests
and 900 click entries in that same window. An immediate retry falls in the same window in more
than 99.99 percent of cases. The counter is already above the limit there. Thus the second
request gets a second 429, and the tracker drops the batch. The retry cannot succeed.

The cost is not zero. A limiter that counts each request, also a rejected one, moves further
above its limit. A full queue gives 4 split requests in one moment (`index.ts:259-261` starts
each one in a `for` loop). A 429 answer then makes 8 requests in one moment, and it consumes 400
click entries of the 900 that D43 allows each minute. A 429 means "slow down". This code speeds
up.

The 5xx case has the same defect with a different reason. An app that answers 500 needs seconds,
not milliseconds. An Atlas failover takes 10 to 30 seconds. A retry after 0 ms meets the same
failed server.

**The rule that I state, with exact numbers.** The design gives no delay, thus I give one:

- A network error or a 5xx response: retry one time after `500 ms`, plus a random value in
  `[0, 500)` ms.
- A 429 response: retry one time after `2000 ms`, plus a random value in `[0, 2000)` ms.
- The `pagehide` flush keeps 0 retries. A timer does not survive a page close.
- The jitter stops a group of tabs from retrying in one moment.

The design already holds this pattern. Decision D6 uses `min(interval * 2^failures, 300 s)` with
a jitter of 10 percent for the monitor poll.

**Which text must hold the rule.** The contract has priority, and the contract says nothing about
a retry delay. Rule C19 names only the status codes. No text holds a `Retry-After` header today.
Thus nothing conflicts. The delay is a client timing rule, thus **decision D24 must hold it**,
beside the sentence "It retries a batch one time after a network error, a 5xx, or a 429". The
issue #36 text must gain the delay in step 3. The tracker README must state it for an app team.
Add a contract rule beside C19 only when each client of the contract must obey the same delay.

**The exact fix.** Give `sendBatch` a delay before the second send. Use `setTimeout` with the
numbers above. Keep the `pagehide` path at 0 retries, thus no timer is needed there. Add a test
with fake timers that proves the second send starts after the delay, and not before it.

---

## MAJOR 2 — the README states a wrong reason for the 64 KB keepalive rule

**File:** `kit/tracker/README.md`, the section "The page lifecycle flush"

**The text today:** "A `keepalive` request stays below 64 KB, well under the browser limit,
because the tracker also holds each request body below the limit of the next section."

**The defect.** The reason is wrong, and the words "well under" are wrong. A browser limits the
**sum** of the bodies of each `keepalive` request in flight for one page. It does not limit each
single body. A rule about one body cannot prove a rule about the sum.

**My measurement in Chrome.**

- One `keepalive` body of 70000 bytes: the `fetch` promise **rejects** with
  `TypeError: Failed to fetch`. It does not throw at the call.
- Five `keepalive` requests of 15000 bytes at the same time: four succeed, the fifth rejects with
  the same error.
- The tracker, a full queue of 200 entries at the contract maximums, one real `pagehide`:
  4 requests of 14371, 14463, 14463 and 14463 bytes. The sum is **57760 bytes**. The quota is
  65536 bytes. The margin is **7776 bytes**, which is 12 percent.

The criterion holds. It holds by arithmetic only. `MAX_QUEUE_SIZE / MAX_BATCH_ENTRIES` gives 4
requests, and `4 * MAX_BODY_BYTES` gives 60000 bytes. No line of code holds this invariant, and
no test asserts the sum. A later issue that raises the queue cap to 400, or that raises
`MAX_BODY_BYTES` to the full 16384, passes the quota and no test fails.

The margin is also not the tracker's alone. The quota belongs to the page, not to the tracker. An
app that sends its own `keepalive` request, or a `navigator.sendBeacon`, at the same page end
takes from the same 64 KB. One of the four tracker requests then rejects, and its 50 clicks
disappear with no console line.

**The exact fix.**

1. Replace the sentence. State that the browser counts the sum of each in-flight `keepalive`
   body for one page. Give the measured worst case of about 58 KB for a full queue of 200
   entries. Tell the app team that its own `keepalive` request or `sendBeacon` at page end shares
   the same 64 KB budget.
2. Add a comment at `index.ts:58` that names the invariant:
   `MAX_QUEUE_SIZE / MAX_BATCH_ENTRIES * MAX_BODY_BYTES` must stay below 65536.
3. Add one test that builds a full queue of 200 entries at the contract maximums, fires
   `pagehide`, and asserts that the sum of each request body stays below 65536 bytes. This test
   holds the invariant for a later issue.

---

## MAJOR 3 — the single-send rule of step 4 does not protect the batch at page end

**File:** `kit/tracker/src/index.ts:119-124`

**The defect.** Step 4 gives the `pagehide` flush 0 retries, because a second send can count
clicks two times. In the Chrome page lifecycle a page reaches the terminated state only through
the hidden state. Thus on a tab close and on a navigation the `visibilitychange` flush runs
first, and it empties the queue. The `pagehide` flush then finds nothing.

I proved the second part with my own test: after a `visibilitychange` flush, a `pagehide` event
gives **1 request in total, not 2**. The protection of step 4 applies to an empty queue.

The batch that carries the clicks at page end goes out from the `visibilitychange` flush, **with
the retry enabled**. The hidden state is the exact state where a browser freezes a page and drops
a connection. The page sees a network error, also when the request reached the store. The retry
then writes the same clicks a second time. Decision 2 of the author follows the issue text, but
it puts the retry on the one flush that carries the highest risk.

**What I could not measure.** The Chrome window of the extension stayed occluded in my run.
`document.visibilityState` was `hidden` for the whole life of the page, thus no
`visibilitychange` event fired. I state the order from the documented lifecycle model, and not
from my own measurement.

**The exact fix.** Choose one, and state the choice in the README:

- (a) Give the `visibilitychange` flush the same 0 retries as `pagehide`. Then no batch at page
  end goes out two times, and step 4 means what it says. The timer flush keeps the retry, because
  it runs while the page is visible.
- (b) Keep the retry, and add this sentence to the README section "The retry rule": "The
  `visibilitychange` flush keeps the retry. A browser can freeze the page after the hidden state
  and drop the connection, thus this flush carries the largest risk of a double count."

I prefer (a). The maintainer must decide.

---

## The answers to the review questions

**Question 2, the double count.** The README states the cause, and that is good. It does not
state the size, the frequency, or what a reader sees. Two paths give a double count:

1. An `insertMany` fails in the middle. The app answers 500. The retry writes the first part
   again. This needs a document-level write failure, thus it is rare.
2. The first request reaches the store and commits, and the browser still reports a network
   error, because the connection dropped after the send and before the answer. At page end this
   is **not** rare. This path is open on the `visibilitychange` flush and on the timer flush. It
   is closed on the `pagehide` flush. See MAJOR 3.

With 500 visitors each day, the app gets about 1000 lifecycle requests each day. I have no
measurement of the network-error rate at page end, thus I give no number for the frequency. The
size of one event is at most one batch: 50 clicks of one session.

**What the monitor shows.** Contract rule C24 drops a copy only for the same `_id`. A retry
writes new `_id` values. Thus the monitor counts each copy. The click count of the affected
elements is too high, and the click count of the session row in "Anonymous sessions" (D44) is too
high. Worse: `flush()` computes `ageMs` one time, at `index.ts:253-254`, and `sendBatch` reuses
the same array. The server computes `ts` as `receivedAt - ageMs` (rule C14). On the retry
`receivedAt` is later and `ageMs` is the same, thus the two copies carry **different** `ts`
values. No later query can find the copies and remove them. The data is wrong for ever.

**What the README must tell an app team.** Add these four lines:

1. A click count can be too high after a retry. It is never too low from this cause.
2. The size of one such error is at most 50 clicks of one session.
3. The copies carry a different `_id` and a different `ts`. No later query can remove them.
4. Do not use the click count for billing, or for any figure that must be exact.

**Is the retry of a 5xx worth that cost?** For a network error, yes. Without a retry the loss is
certain, and the double count is an accident. For a 5xx, the value is the lowest of the three.
A 500 from this app is the exact signal of a mid-batch `insertMany` failure. Thus the retry is
not merely at risk of a copy; it is the case that the maintainer described. With no delay
(MAJOR 1) the retry also meets the same failed server. My judgement: keep the 5xx retry only with
the delay of MAJOR 1, and state in the README that a 500 answer is the case with the highest
chance of a copy.

**Question 4, the events.** I tested each case. The results:

- `pagehide` with `persisted: true` (the back/forward cache): Chrome fired it on a real
  navigation. The tracker sent its 4 requests, and each one arrived. The queue empties. After a
  restore, a new click starts a new timer and goes out. Correct.
- `visibilitychange` to hidden, then `pagehide`: 1 request in total. No batch goes out two times.
  Correct.
- A tab in the background that comes back: no entry stays in the queue. The pending timer always
  drains the queue later. Correct.
- A real tab close: `pagehide` with `persisted: false`, 1 request with 2 clicks, delivered.
- The code adds **no** `unload` listener and **no** `beforeunload` listener. I asserted this on
  `window` and on `document`. The back/forward cache stays open, and Chrome proved it: my run
  gave `persisted: true`.

**Question 5, the second send.** The second send **does** use `keepalive: true`. `sendBatch`
passes its `keepalive` argument to the recursive call at `index.ts:301` and at `index.ts:307`. I
asserted the flag on the second `fetch` call. A second send that never runs leaves a clean state:
the queue is already empty, the timer is untouched, and no flag stays set. The next visible phase
sends only the new click, with no copy of the lost batch. The batch is lost in silence, and that
is the accepted trade.

**Question 6, no error reaches the page.** Four cases, each one broken by me:

| Case | Result |
|---|---|
| A rejected `fetch` | No throw. The `.catch()` at `index.ts:305` holds it. Each split batch still goes out. |
| A `fetch` that throws at the call | No throw into the page. `safeFlush` holds it. But see MINOR 4. |
| An absent `fetch` | No throw into the page. |
| A `JSON.stringify` that throws | No throw into the page. No request at all, because the split measures the body first. |
| Each case | No `console.warn` and no `console.error`. This matches the README. |

**Question 7, the real browser check.** A Chrome browser was available. I built the tracker, I
served `dist/` and a test page from a local server on port 7532, and I recorded each request. The
results are in MAJOR 2, in MAJOR 3, and in question 4 above. **What stays without a measurement:**
the `visibilitychange` flush in a real browser and its order against `pagehide` (the window stayed
occluded, thus the page was hidden for its whole life and the event never fired); a freeze and a
resume on a mobile browser; Safari; Firefox.

**Question 8, the break test.** `npm ci` and `npm test` pass: 121 tests. I broke the code six
times, and I restored each file:

|---|---|---|
| 1 | `PAGEHIDE_RETRY_COUNT` 0 to 1 | 2 tests fail |
| 2 | The visibility condition inverts | 3 tests fail |
| 3 | The `pagehide` flush drops `keepalive` | 1 test fails |
| 4 | Each failed status retries, also a 400 | 1 test fails |
| 5 | **The retry drops `keepalive`** | **0 tests fail** — see MINOR 1 |
| 6 | `MAX_BODY_BYTES` 15000 to 60000 | 1 test fails |
| 7 | `MAX_QUEUE_SIZE` 200 to 500 | 1 test fails |

No test is empty. No test is always green. Break 5 is the one gap.

---

## MINOR findings

**MINOR 1 — no test holds `keepalive` on the second send.**
`kit/tracker/src/index.test.ts:1208`. I changed both recursive calls of `sendBatch` to
`keepalive: false`, and all 76 tests still passed. The code is correct; the test is missing. Fix:
in the test "retries a batch from the visibilitychange listener, unlike pagehide", assert that
the second `fetch` call has `init.keepalive === true`.

**MINOR 2 — a lifecycle flush leaves the pending timer.**
`kit/tracker/src/index.ts:244-262`. `flush()` empties the queue and it clears no timer. I asserted
that `clearTimeout` never runs on a `visibilitychange` flush. Two results follow. The stale timer
fires one extra time with an empty queue. The next click gets no timer of its own, because
`scheduleFlush` at `index.ts:218-221` returns early, thus that click goes out earlier than
`flushIntervalMs`. No entry is lost. Fix: in `flush()`, after `queue = []`, add
`if (timerId !== null) { clearTimeout(timerId); timerId = null; }`.

**MINOR 3 — the file comment of `batch.ts` contradicts the function comment.**
`kit/tracker/src/batch.ts:7-8` states "Each request body stays below the 16 KB limit of contract
rule C18, with a safety margin". The comment of `splitIntoRequestBatches` at
`kit/tracker/src/batch.ts:34-36` states the exception: one single click above the limit forms its
own group. Fix: add the exception to the file comment.

**MINOR 4 — a synchronous throw of the first `fetch` abandons each later split batch.**
`kit/tracker/src/index.ts:259-261`. The `for` loop starts each `sendBatch` inside one try/catch of
`safeFlush`. When the first `fetch` throws at the call, the loop stops, and each later batch never
goes out. Its entries already left the queue. My test: 120 clicks give 3 batches, and only 1
`fetch` call runs. The impact is small, because Chrome **rejects** for the quota case and for a
content security policy; it does not throw at the call (I measured this). Fix: put the
`void sendBatch(...)` call inside its own try/catch inside the loop.

---

## Scope and the design

The change does what the issue asks, and no more. Each name, each number, and each rule agrees
with decision D24 and with rules C13, C17, C18, and C19. `package.json` adds `src/batch.ts` to
`files`, which is correct for the `npm pack` of decision D25. The README removes the old "Out of
scope" text correctly.

I approve after MAJOR 1, MAJOR 2, and MAJOR 3 close.

xamcross:
Correction round 1 of this pull request. It answers the TypeScript review
(https://github.com/xamcross/octometer/pull/143#issuecomment-5769368832) and the
reliability review (https://github.com/xamcross/octometer/pull/143#issuecomment-5769393367),
and the issue #36 edit of 2026-09-22.

For each item below, I mutated the code, ran the test, saw the failure, then restored the
code. I did this myself, in my own clone.

## MAJOR findings

**TypeScript review MAJOR 1 — a retry goes out after `stop()`. Fixed.**
`kit/tracker/src/index.ts:134` adds a run token, `runId`. `start()` (line 145) and `stop()`
(line 165) each raise it. `sendBatch` (line 343) checks the token before the second send
(lines 363, 369). `waitThenRetry` (line 384) checks it again after the wait (line 394), because
`stop()` can run during the delay. Tests: "sends no retry after stop(), also when the first
request settles after stop()" and "sends no retry after stop() runs during the wait before the
retry", both in `index.test.ts`. Each mutation of either check makes one of these tests fail.

**TypeScript review MAJOR 2 and reliability review MAJOR 2 — no test proves the byte limit
with real worst-case data. Fixed.**
`index.test.ts`, "holds each request and the whole keepalive flush below the limits, with 200
entries at the contract maximum": 200 entries at the contract maximum sizes, one real
`pagehide` flush, no spy. It asserts each request body below 15 000 bytes and the sum below
64 KB. The old spy-based test keeps the loop check only; I removed its meaningless byte
assertion. `kit/tracker/src/index.ts:58-66` adds the invariant comment,
`MAX_QUEUE_SIZE / MAX_BATCH_ENTRIES * MAX_BODY_BYTES < 65536`, beside `MAX_QUEUE_SIZE`.

**Reliability review MAJOR 1 — the retry has no delay. Fixed.**
`kit/tracker/src/index.ts:84-94` adds the four delay constants. `retryDelayMs` (line 508)
gives 500 ms plus jitter below 500 ms for a network error or a 5xx, 2000 ms plus jitter below
2000 ms for a 429. `waitThenRetry` (line 384) applies the wait with `setTimeout`. Tests: "waits
the full delay before it sends a batch again after a network error" and "...after a 429
response", each checking the call count one ms before, and at, the exact deadline, with
`Math.random` mocked to 0 (the random value is replaceable this way, for the test).

**Reliability review MAJOR 3 — the retry of the `visibilitychange` flush carries the largest
risk. Fixed with option (a), the maintainer's preferred choice.**
`kit/tracker/src/index.ts:82` renames the constant to `LIFECYCLE_RETRY_COUNT` (0), and line 154
gives the `visibilitychange` flush this count too, the same as `pagehide`. Only the timer flush
(line 262, `TIMER_RETRY_COUNT`) still retries. Tests: the "the single send of the lifecycle
flush" describe block, and "keeps the retry of the timer flush, unlike the lifecycle flush".

**Decision (issue #36 edit, D24 written).** The delay rule and the lifecycle rule are now in
`docs/superpowers/specs/2026-09-21-octometer-design.md`, decision D24, beside the retry
sentence. Only that part of D24 changed.

## MINOR findings — 6 of 6 fixed

1. **An entry above the byte limit alone (TS review MINOR 1).** `kit/tracker/src/batch.ts:47-76`
   drops such a click and writes one console warning for the call, with no click text in it.
   Test: "drops one click that alone passes the byte limit, with one console warning", and
   "writes only one warning for one call, also with two oversized clicks", in `batch.test.ts`.
2. **A lifecycle flush leaves the pending timer (reliability review MINOR 2).**
   `kit/tracker/src/index.ts:290-296` clears the timer inside `flush()`. Test: "clears the
   pending timer, so a later click waits its own flushIntervalMs", in `index.test.ts`.
3. **The splitter re-encodes the whole group for each click (TS review MINOR 4).**
   `kit/tracker/src/batch.ts:47-83` now measures each click one time, with a running byte
   count and a fixed frame cost, instead of re-encoding the whole group again. Verified by the
   existing correctness tests (order, count, split points). A timing assertion was not added,
   because it would be flaky in CI; I checked the linear cost by reading the new loop.
4. **A synchronous throw of `fetch` abandons later split batches (reliability review MINOR 4).**
   `kit/tracker/src/index.ts:305-313` gives each batch its own try/catch inside the loop. Test:
   "still sends a later batch when an earlier batch throws at the fetch call", in
   `index.test.ts`.
5. **A stored session id passes with no check (TS review MINOR 5).**
   `kit/tracker/src/index.ts:95-96, 424-436` adds `SESSION_ID_PATTERN` (the C5 rule, the same
   36-character UUID pattern as the server's `EventFieldValidator`) and checks a stored value
   against it before use. Test: "creates a new sessionId when the stored value breaks the UUID
   rule of rule C5".
6. **README and comment sentences over 25 words (TS review MINOR 6).** Rewritten in
   `kit/tracker/README.md` and in `kit/tracker/src/index.ts` and `src/batch.ts`, wherever a
   sentence passed 25 words, including several pre-existing ones in files this round touches.

Three more MINOR findings of the reliability review are resolved as a side effect of the MAJOR
fixes above, not by a separate change:

- **MINOR 1, no test holds `keepalive` on the second send.** The scenario itself is gone: the
  `visibilitychange` flush no longer retries (MAJOR 3), so no lifecycle retry with `keepalive`
  exists to test. The timer flush, which does retry, never sets `keepalive: true`.
- **MINOR 3, the file comment of `batch.ts` contradicts the function comment.** Both comments
  now state the drop-and-warn behaviour of the MINOR 1 (TS review) fix, so they agree.
- **The wrong keepalive-limit reason and the missing sum test (MAJOR 2 of this review).** Fixed
  above, with the README rewrite and the 200-entry integration test.

## Numbers

- `npm ci`: 0 vulnerabilities.
- `npm test`: 131 tests passed (85 in `index.test.ts`, 9 in `batch.test.ts`, 37 in
  `path-match.test.ts`).
- `npm run typecheck`: no error.
- `npm run build`: no error.
- Bundle size (gzip): `index.js` 11 198 / 2 889, `path-match.js` 4 047 / 1 197, `batch.js`
  1 605 / 577. Combined: 16 850 bytes raw, 4 663 bytes gzip.
- `npm pack --dry-run`: 17 files, 21.5 kB package, 84.0 kB unpacked, no test file.
- CI: all 5 checks pass (Frontend, JVM modules, Monitor backend on Windows, Tracker, Secret
  scan).

Commit: f615e9c. The pull request description now lists the new acceptance criteria of the
issue #36 edit as a checklist, with evidence for each one.

xamcross:
**Review (second review, TypeScript library engineer): APPROVE**

I made a second clone, and I ran `gh pr checkout 143` in it (commit `f615e9c`, on the merge
`8cec099`). I read the issue #36 body of 2026-09-22 and the maintainer comment first. I ran each
command and each mutation myself. Both MAJOR findings of my first review are closed, and each of my
six MINOR findings is closed. Three new MINOR findings stay open. None of them blocks this pull
request.

## The commands that I ran

| Command | Result |
| --- | --- |
| `npm ci` | OK, 0 vulnerabilities |
| `npm test` (three times) | `Tests 131 passed (131)`, 3 files, each time |
| `npm test` with a shared jsdom (`--no-isolate`) | 131 passed |
| `npm test` with a shuffled order (seeds 7 and 42) | 131 passed each time |
| `npm test` with the files in the reverse order | 131 passed |
| `npm run typecheck` | no error |
| `npm run build` | no error |
| `npm pack --dry-run` | 17 files, 21.5 kB package, 84.0 kB unpacked, no test file |

Bundle size, from my own build, with `gzip -9`: `index.js` 11 198 raw / 2 907 gzip, `path-match.js`
4 047 / 1 215, `batch.js` 1 605 / 586. The three files together: 16 850 raw / 4 200 gzip. The count
of 131 tests agrees with the author.

## MAJOR 1 — a retry after `stop()`. Closed.

`kit/tracker/src/index.ts:134` adds the run token. `start()` (line 145) and `stop()` (line 165) each
raise it. The token guards the retry at line 363, at line 369, and again in the timer callback of
`waitThenRetry` (line 394).

My own evidence, from two tests that I wrote and ran in my clone:

- I held the first `fetch` open, called `stop()`, then failed the request. The mock records **1
  call**. Before the fix it recorded 2 calls.
- I failed the first request with a 503 response, let the retry wait start, then called `stop()`
  during the wait. The mock records **1 call**.

I also broke the fix, one check at a time:

| Mutation | Result |
| --- | --- |
| Remove the token check of `waitThenRetry` (line 394) | 1 test fails: "sends no retry after stop() runs during the wait before the retry" |
| Remove the token check at line 363 | 131 pass |
| Remove the token check at line 369 | 131 pass |
| Remove the token check of the flush timer (line 258) | 131 pass |

The behaviour is correct, and one test holds it. The check of `waitThenRetry` carries the whole
guard today, because each retry goes through the wait. The three other checks are defence, and the
check at line 258 cannot run at all, because `stop()` clears that timer first (line 167). Keep them;
they cost nothing. This is an observation, not a finding.

## MAJOR 2 — no test measures the byte limit with real data. Closed.

`kit/tracker/src/index.test.ts:1500` holds the new test: 200 entries, a 100-character element, a
150-byte path, one real `pagehide` flush, and no spy on the splitter. It asserts the entry count,
each body below 15 000 bytes, the sum below 65 536 bytes, and `keepalive: true` on each request.
`kit/tracker/src/index.ts:58-66` names the invariant
`MAX_QUEUE_SIZE / MAX_BATCH_ENTRIES * MAX_BODY_BYTES < 65536`. The spy test at line 1474 keeps the
loop check only, and its meaningless byte assertion is gone. My own measure of the same case gives 4
requests of 14 563 bytes, 58 252 bytes in total. See MINOR 2 below for one gap that stays.

## My six MINOR findings of the first review

| --- | --- | --- | --- |
| 1 | One entry above the byte limit gives a request above the limit | Closed | `batch.ts:56-66` drops the click and writes one warning with no click text. Mutation: a byte limit of 1e9 makes 3 tests fail. |
| 2 | No delay before the retry after a 429 response | Closed | `index.ts:92-94, 508-513`. Mutation `RETRY_DELAY_429_MS = 500` makes the boundary test fail. |
| 3 | The README states the keepalive limit as a per-request rule | Closed | The README now states the sum, the measured 58 KB, and the shared budget of the page. |
| 4 | The splitter re-encodes the whole group for each click | Closed | `batch.ts:51-79` keeps a running count. I compared it against a full re-measure for 40 random click sets, with Cyrillic text, an emoji, and Japanese text: the same split each time, the worst body 14 984 bytes. |
| 5 | A stored session id passes with no check | Closed | `index.ts:96, 429-432`. Mutation: the removal of the check makes the new test fail. |
| 6 | Two README sentences pass 25 words | Closed | The two sentences now hold 21 and 17 words. |

## The delay rule of the new step 3

The wait is `500 + Math.random() * 500` for a network error and for a 5xx response, and
`2000 + Math.random() * 2000` for a 429 response (`index.ts:508-513`). This agrees with the issue
text and with decision D24.

- The tests replace the random value with `vi.spyOn(Math, 'random').mockReturnValue(0)`, so the
  jitter is replaceable. `retryDelayMs` reads the global `Math.random`, and `afterEach` restores it.
- The boundary holds. `index.test.ts:1195` checks 499 ms and 500 ms. `index.test.ts:1217` checks
  1999 ms and 2000 ms. Mutation `RETRY_DELAY_MS = 0` makes the first fail. Mutation
  `RETRY_DELAY_429_MS = 500` makes the second fail.
- The range holds. With `Math.random` at 0.999 I measured the second send at 1000 ms after the
  failure, and not at 998 ms. See MINOR 1 for the missing test of that bound.
- Each lifecycle flush sends one time only (`index.ts:82, 151, 154`). Mutation
  `LIFECYCLE_RETRY_COUNT = 1` makes 3 tests fail.

## The design document

`git diff f615e9c~1 f615e9c -- docs/.../2026-09-21-octometer-design.md` gives 8 changed lines, all
inside decision D24, and only the retry sentence and the lifecycle sentence. The other file changes
of the branch come from the merge commit `8cec099` of `main`. The scope is correct.

## The new findings

### MINOR 1 — the size of the random part has no test

`kit/tracker/src/index.ts:86` and `kit/tracker/src/index.ts:94`.

Each delay test mocks `Math.random` to 0, thus each test measures the fixed part only. I set
`RETRY_JITTER_MS` to 50 000, a wait of up to 50 seconds, and the whole suite stays green: 131
passed. A wrong jitter constant reaches a user with no test failure.

Fix. Add one test with a high random value. This test passes on the present code, and it fails with
the mutation above.

```ts
it('keeps the retry delay below one second, also with the largest random value', async () => {
  vi.spyOn(Math, 'random').mockReturnValue(0.999);
  fetchMock.mockRejectedValueOnce(new Error('network error')).mockResolvedValue({ ok: true, status: 204 });
  // ... one click, then advance the flush timer ...
  await vi.advanceTimersByTimeAsync(998);
  expect(fetchMock).toHaveBeenCalledTimes(1);
  await vi.advanceTimersByTimeAsync(2);
  expect(fetchMock).toHaveBeenCalledTimes(2);
});
```

### MINOR 2 — the keepalive sum test holds the number 200 itself

`kit/tracker/src/index.test.ts:1510`.

The test loop clicks 200 times, which is the value of `MAX_QUEUE_SIZE` today. I raised
`MAX_QUEUE_SIZE` to 400, and this test stayed green, because it still sends only 200 clicks. The
suite does catch the change, through the separate queue-cap test at line 343, but the test that
carries the invariant of the new acceptance criterion does not.

Fix. Click 400 times in this test. The queue cap then decides the entry count, and a later change of
`MAX_QUEUE_SIZE` makes this test fail. The assertion `totalClicks` becomes the cap itself, thus keep
the byte assertions and drop the exact count check, or export the constant.

### MINOR 3 — no test holds `MAX_BODY_BYTES` under the 16 KB limit of rule C18

`kit/tracker/src/batch.ts:30`, with `kit/tracker/src/batch.test.ts:52-61`.

The byte tests compare against the constant itself, so they stay green for each value of the
constant. I set `MAX_BODY_BYTES` to 16 300, above the 16 KB limit of rule C18, and the suite stays
green: 131 passed.

Fix. Add one line to `batch.test.ts`:

```ts
it('keeps the body limit under the 16 KB limit of contract rule C18', () => {
  expect(MAX_BODY_BYTES).toBeLessThan(16 * 1024);
});
```

## What I checked again and found correct

- The tests stay stable: three plain runs, one run with a shared jsdom document, two shuffled runs,
  and one run with the files in the reverse order. Each run gives 131 passed. The new async retry
  tests add no flake and leave no timer behind.
- The incremental byte count of `batch.ts` is exact. My reference comparison gives the same split as
  a full re-measure, for 40 random click sets with non-ASCII text.
- No entry is lost, no entry goes into two requests, and no request is empty, for the 200-entry case
  at the contract maximum.
- A lifecycle flush now clears the pending timer (`index.ts:290-296`), and each split batch has its
  own `try/catch` (`index.ts:306-312`). A mutation of either one makes its test fail.
- The new console warning of `batch.ts:61` holds no click text, no path, no pattern, and no session
  id. No other console line is new.
- The commit `f615e9c` touches the design document and `kit/tracker` only. It holds no secret, no
  database file, and no build output. The commit body and the new comments follow ASD-STE100.

The three MINOR findings above are test guards, not defects of the shipped behaviour. I approve this
pull request. Please add the three tests in this round or in a follow-up issue.

xamcross:
**Review (second review, browser reliability engineer): APPROVE**

I made a new clone, and I ran `gh pr checkout 143` in it. I checked the commits `8cec099` and
`f615e9c`. I broke the code for each fix, I saw the test fail, and I restored each file. I also
ran the built tracker in a real Chrome browser, against a local server that answers 429 and 500
on command. My clone is deleted. My server is stopped. My tab is closed.

`npm ci`: 0 vulnerabilities. `npm test`: **131 tests passed** (3 files). `npm run typecheck`: no
error. `npm run build`: no error.

---

## Each MAJOR finding closes

### MAJOR 1 — the retry delay. CLOSED.

`kit/tracker/src/index.ts:84-94` holds the four constants. `retryDelayMs` at
`kit/tracker/src/index.ts:507-512` gives the two waits. `waitThenRetry` at
`kit/tracker/src/index.ts:384-401` applies the wait with `setTimeout`.

My break: I made `retryDelayMs` return `0` for both branches. Two tests fail:
"waits the full delay before it sends a batch again after a network error" and "...after a 429
response". A second break, the 429 branch with the short delay, fails the 429 test alone. Thus
the 2000 ms is held on its own, and not only the 500 ms.

My own probe measured the upper end. With `Math.random` at 0.999999, a network error waits below
1000 ms, and a 429 waits below 4000 ms. `setTimeout` truncates the float value, thus the real
bounds are `[500, 1000)` ms and `[2000, 4000)` ms.

**Real Chrome, my own measurement:**

| Case | The measured wait | The design range |
|---|---|---|
| A 429 response from the timer flush | **3991 ms** | 2000 to 4000 ms |
| A 500 response from the timer flush | **999 ms** | 500 to 1000 ms |

### MAJOR 2 — the README and the sum of the keepalive bodies. CLOSED.

`kit/tracker/README.md` now states the rule correctly: a browser limits the **sum** of the
bodies of each `keepalive` request in flight for one page. It gives the measured worst case of
about 58 KB. It also warns the app team that the quota belongs to the page, and that an own
`keepalive` request or a `sendBeacon` call at the page end takes from the same budget. Each of
my three fix items is present.

The invariant comment stands beside the constant, at `kit/tracker/src/index.ts:58-66`:
`MAX_QUEUE_SIZE / MAX_BATCH_ENTRIES * MAX_BODY_BYTES < 65536`.

The new test at `kit/tracker/src/index.test.ts:1500` is real. It builds 200 entries at the
contract maximum sizes, it runs one real `pagehide` flush with no spy on the splitter, and it
asserts each body below 15000 bytes and the sum below 65536. My own probe gives the same numbers:
4 requests of 14563 bytes, a sum of 58252 bytes, a margin of 7284 bytes.

### MAJOR 3 — the retry of the `visibilitychange` flush. CLOSED, with option (a).

`kit/tracker/src/index.ts:82` gives `LIFECYCLE_RETRY_COUNT = 0`. Line 151 and line 154 give this
count to both listeners, each with `keepalive: true`. Only the timer flush keeps a retry
(`TIMER_RETRY_COUNT`, line 75), and it never sets `keepalive`.

My break: I gave the `visibilitychange` flush `TIMER_RETRY_COUNT`. The test "sends a batch only
one time from visibilitychange too, also after a network error" fails.

My own probe confirms the whole rule. After a network error, each lifecycle flush sends one
request, with `keepalive: true`. The timer flush sends two requests, each with `keepalive: false`.

**Real Chrome, my own measurement.** A navigation with a full queue of 200 entries, and a 500
answer to the first request: **4 ingest requests, not 5**. All 200 clicks arrive one time. The
event was `pagehide` with `persisted: true`, thus the back/forward cache stays open. A real tab
close gave `pagehide` with `persisted: false` and 1 request with 3 clicks.

---

## Each MINOR finding closes

| # | The finding | Result | My evidence |
|---|---|---|---|
| 1 | No test holds `keepalive` on the second send | CLOSED by removal | The case no longer exists. A lifecycle flush never retries, and the timer flush never sets `keepalive`. My probe asserts both. |
| 2 | A lifecycle flush leaves the pending timer | FIXED | `index.ts:290-296` clears the timer. My break removed the block: the test "clears the pending timer, so a later click waits its own flushIntervalMs" fails. My probe: after a flush at 1000 ms, a later click now waits its full 5000 ms. |
| 3 | The file comment of `batch.ts` contradicts the function comment | FIXED | `batch.ts:1-13`. The splitter now drops an oversized click, thus "Each request body stays below the 16 KB limit" is true with no exception. The header names the guard. My break kept such a click: 2 tests in `batch.test.ts` fail. |
| 4 | A synchronous throw of the first `fetch` abandons the later batches | FIXED | `index.ts:305-313` gives each batch its own try/catch. My break removed it: the test "still sends a later batch when an earlier batch throws at the fetch call" fails. My probe: 120 clicks give 3 batches, the first `fetch` throws at the call, and `fetch` still runs 3 times. |

---

## Decision D24 on the branch

The two new sentences are correct and exact. They match the rule that I stated, and they match
the code:

- "The wait is 500 ms plus a random value below 500 ms for a network error or a 5xx. The wait is
  2000 ms plus a random value below 2000 ms for a 429, because the rate limit uses a fixed
  60-second window (D20, D43)."
- "Only the timer flush retries. The `pagehide` flush and the `visibilitychange` flush send a
  batch one time only. A browser can freeze the page after the hidden state and drop the
  connection."

The numbers, the reason, and the decision references D20 and D43 are right. Only the retry part
of D24 changed, as the author states.

---

## The real browser check

A Chrome browser with the extension was available. I built the tracker, I served `dist/` and a
test page from a local server on port 7533, and I recorded each request with its size and its
time. The results are in MAJOR 1 and in MAJOR 3 above. **What stays without a measurement:** the
`visibilitychange` flush in a real browser (the extension window stayed occluded again, thus the
page was hidden for its whole life and the event never fired); a freeze and a resume on a mobile
browser; Safari; Firefox.

---

## Three new MINOR findings, none of them blocking

**NEW MINOR A — the jitter has no test.** `kit/tracker/src/index.ts:509, 511`. I removed
`Math.random() * RETRY_JITTER_MS` and `Math.random() * RETRY_JITTER_429_MS` from both branches.
**0 of 131 tests fail.** The two delay tests mock `Math.random` to 0, thus they cannot see the
jitter. Decision D24 states the random value, thus a test must hold it. Fix: add one test with
`Math.random` mocked to a value near 1, and assert the upper bound of each wait.

**NEW MINOR B — the sum test does not hold the invariant against a change of the cap.**
`kit/tracker/src/index.test.ts:1510` writes `i < 200`, and not `MAX_QUEUE_SIZE`. I raised
`MAX_QUEUE_SIZE` to 400: the sum test still passes, because it queues only 200 entries. The
comment at `index.ts:58-66` states the invariant, and no test holds it. Fix: export
`MAX_QUEUE_SIZE`, and build the loop from it.

**NEW MINOR C — decision D24 keeps the old per-body wording.** The design document, line 391,
still reads "A keepalive body stays below 64 KB". This is the same framing that the README
corrected. I did not raise this in my first review, thus I do not block on it. Fix, in a later
issue: use the same sentence as the README.

---

## Scope, secrets, and the language

The diff of `f615e9c` touches the tracker, its tests, its README, and decision D24 only. No
secret, no database file, and no build output is in the diff. The language of the README, of each
new comment, and of the commit body follows ASD-STE100.

PR #127, opened 2026-09-21
feat(monitor): serve the level 1 totals at GET /api/apps

xamcross:
**Summary.** This pull request adds `GET /api/apps`. The route runs three
read statements on the read connection and merges each result with the
`app` rows. Each statement uses one covering index, per the review of
issue #109 on pull request #121. An app without events gets zero totals
and the derived status `NEVER_POLLED`.

**Acceptance criteria.**
- [x] A test seeds events of 2 apps with `NULL` user ids and asserts the
  three numbers. Evidence: `AppTotalsRouteTest.two apps with NULL user
  ids give the right clicks, unique users, and unique sessions`, PASS.
- [x] `EXPLAIN QUERY PLAN` shows a covering index for each statement.
  Evidence: `AppTotalsRouteTest.EXPLAIN QUERY PLAN shows a covering
  index for each of the three statements`, PASS. The plans:
  `SCAN event USING COVERING INDEX event_agg` (clicks and unique
  users), `SCAN event USING COVERING INDEX event_session` (unique
  sessions).
- [x] A row with `kind = 1` does not change `clicks` and `uniqueUsers`,
  and its session counts in `uniqueSessions`. Evidence:
  `AppTotalsRouteTest.a session-start row does not change clicks or
  uniqueUsers, but its session counts`, PASS.
- [x] An app without events returns zeros and `NEVER_POLLED`. Evidence:
  `AppTotalsRouteTest.an app without events returns zeros and
  NEVER_POLLED`, PASS.
- [x] Each row of the response holds the nine fields of step 4.
  Evidence: `AppTotalsRouteTest.each row holds the nine fields, and a
  time field is UTC ISO 8601 with milliseconds`, PASS.
- [x] Each time field is UTC ISO 8601 with milliseconds. Evidence: the
  same test asserts `2023-11-14T22:13:20.123Z` and
  `2023-11-14T22:14:20.000Z`.
- [x] `./gradlew :monitor:backend:test` passes. Evidence: `BUILD
  SUCCESSFUL`, 76 tests, 0 failures, 0 errors. I ran this command
  myself with `JAVA_HOME=C:/Users/xamcr/.jdks/ms-21.0.9`.
- [x] A comment on the issue records the three query plans. See the
  comment on issue #18.

**Confirm results.**
- The field names and the types of `AppTotalsRow` agree with the row
  type of issue #20. Source: `git show
  origin/issue-20-app-table:monitor/frontend/src/app/apps/app-row.ts`.
[...]

Issue #113, opened 2026-09-21
Serve the anonymous sessions and the sessionId filter


xamcross:
**Blocked by #110.** The statement reads the column `kind` of #109 and the rows that #110 marks, and the third filter joins the route of #51. You can write the response model now.

**Goal.** `GET /api/apps/{appId}/sessions?anonymous=true&firstPath=<path>&page=` lists the anonymous sessions. The elements route gets the third filter `sessionId`.

**Implementation steps.**
1. Add the route `GET /api/apps/{appId}/sessions?anonymous=true&firstPath=<path>&page=` to `monitor/backend`. One page holds 50 rows.
2. Read the start rows with `kind = 1` and `user_id IS NULL`, in the order `startTime DESC, sessionId ASC`. The `firstPath` value is a bound parameter.
3. Give each row the session id, the first path, the source, the start time, the click count, and the user id of a later sign-in. The source is the `referrerHost` value, or `(direct)` when the field is absent.
4. Show the path `(unknown)` and the time of the first click for a session without a start row.
5. Add the third filter `sessionId=<uuid>` to `GET /api/apps/{appId}/elements`. Exactly one of the three filters is allowed, else return 400.
6. Write the tests of the acceptance criteria.

**Acceptance criteria.**
- [ ] The order is `startTime DESC, sessionId ASC`, and one page holds a maximum of 50 rows.
- [ ] `firstPath` filters the rows, and the value is a bound parameter.
- [ ] Each row holds the session id, the first path, the source, the start time, the click count, and the user id.
- [ ] A row without `referrerHost` gives the source `(direct)`.
- [ ] The click count counts only rows with `kind = 0`.
- [ ] A session with a later sign-in shows that user id.
- [ ] A session without a start row shows `(unknown)` and the time of its first click.
- [ ] `EXPLAIN QUERY PLAN` shows the index `event_start`.
- [ ] A request to the elements route with exactly one of `userId`, `anonymous=true`, and `sessionId` passes. Each other request gives 400.
- [ ] The elements statement with `sessionId` holds the literal `kind = 0` and groups by `element`.
- [ ] `EXPLAIN QUERY PLAN` shows the index `event_session` for that statement.
- [ ] Each new route has the Host check of D12, it sends `Cache-Control: no-store`, and it holds no CORS header.
- [ ] The captured log holds no query string and no raw path.
- [ ] `./gradlew :monitor:backend:test` passes.
- [ ] A comment on this issue records the query plans.

**Related.**
[...]

Issue #50, opened 2026-09-21
Serve the level 2 totals with pages and a filter


xamcross:
**Goal.** `GET /api/apps/{appId}/users?page=&q=` returns one row for each user of an app: the clicks, the sessions, and the unique elements.

**Implementation steps.**
1. Add the route `GET /api/apps/{appId}/users` to `monitor/backend`.
2. Write the level 2 statement of section 6. It groups the events of one app by `user_id`, and it holds the literal `kind = 0`.
3. Add the filter `q` on the user id.
4. Add the pages of 50 rows. The response holds `page`, `pageCount`, and `rows`.
5. Return 404 for an unknown app id.
6. Write the tests of the acceptance criteria.

**Acceptance criteria.**
- [x] The order is `clicks DESC, user_id ASC`.
- [x] The server clamps a page above the last page, and the response holds `page` and `pageCount`.
- [x] `q` filters the user id with one `LIKE`, and the value is a bound parameter.
- [x] Each row holds `userId`, `clicks`, `sessions`, and `uniqueElements`. The `NULL` user is one row with `userId: null`.
- [x] `EXPLAIN QUERY PLAN` shows the index `event_agg`.
- [x] An unknown app id gives 404.
- [x] The captured log holds no query string.
- [x] `./gradlew :monitor:backend:test` passes.
- [x] A comment on this issue records the test result and the `EXPLAIN QUERY PLAN` output.

**Related.** #31 owns the format of the call log. #109 gives the column `kind` and the partial index `event_agg`.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, D13, D15, and section 6.

xamcross:
Body edit on 2026-09-21: #9 closed. No open blocker remains, thus the first line and the label `blocked` are removed. The work can start.

xamcross:
Body edit of 2026-09-21: rule M4 of the design change. The level 2 statement counts only a click, thus it holds the literal `kind = 0`. Issue #111 makes that change, thus a Related line names it. The steps stay as they are.

The maintainer made this edit. The source is the design brief in the first comment of #101.

xamcross:
[...]

PR #144, opened 2026-09-22
feat(monitor): serve the level 3 element totals

### LINKED ISSUE #51: Serve the level 3 totals

xamcross:
## Summary

This pull request adds the route `GET /api/apps/{appId}/elements`. It answers the level 3 totals: one row for each element that one user touched, with the clicks, the sessions, and the last interaction time. The route lives in its own file and its own package, `octometer.monitor.elements`, so it shares no file with issue #50. The level 3 statement filters with `user_id IS ?` and holds the literal `kind = 0`, so SQLite keeps the covering index `event_agg`.

**Correction round 1** (after the SQL review and the security review) fixes two MAJOR findings and six MINOR findings. See the comment below for the full list.

## Acceptance criteria

- [x] A request with exactly one of `userId` and `anonymous=true` passes. Each other request gets 400.
  Evidence: `ElementsRouteTest` — `userId alone is a valid filter`, `anonymous=true alone is a valid filter`, `no parameter is invalid`, `userId and anonymous together are invalid`, `a blank userId is invalid`, `anonymous=false is invalid`, `sessionId alone is invalid, because issue #113 owns that filter`, `a repeated userId is invalid, even though the key appears once`, `anonymous=true and anonymous=false together are invalid`, and the HTTP test `GET with no filter gets 400, not 500, and writes no ERROR log line`.
- [x] The SQL uses `user_id IS ?`.
  Evidence: `ElementsRoute.kt`, the constant `ELEMENT_TOTALS_SQL`. The test `EXPLAIN QUERY PLAN shows the covering index event_agg for the level 3 statement` runs this exact text.
- [x] The anonymous rows are correct.
  Evidence: `anonymous=true reads the rows with user_id IS NULL, and never a named user's rows`.
- [x] The response holds `rows` with `element`, `clicks`, `sessions`, and `lastInteractionAt`.
  Evidence: `GET with userId answers 200 with the rows field and the four JSON names`. `sessions` now has its own defending test; see the correction round comment, MAJOR 1 of the SQL review.
- [x] `lastInteractionAt` is UTC ISO 8601 with milliseconds.
  Evidence: `lastInteractionAt is the UTC ISO 8601 form of the newest click of the element, with milliseconds`.
- [x] The order is `clicks DESC, element ASC`.
  Evidence: `the order is clicks DESC then element ASC`.
- [x] An unknown app id gives 404.
  Evidence: `an unknown app id gives AppNotFound` (function level) and `GET for an unknown app id gets 404` (HTTP level).
  Evidence: `GET with no filter gets 400, not 500, and writes no ERROR log line` and `a request with a userId query string writes no log line that holds that value`. The second test now reads `captureLogEvents` (every log level), not `captureErrorLogEvents`; see the correction round comment, MAJOR 1 of the security review.
  Evidence: two runs in sequence after the correction round and the merge of `origin/main`, both green. See the test count below.
  Evidence: posted after this pull request opens, and again after the correction round.

## Extra evidence

- A non-numeric app id gives 400: `GET with a non-numeric app id gets 400`.
- A hostile user id (a quote character and a SQL keyword) stays a bound parameter and never breaks the statement: `a user id with a quote character stays a bound parameter, and never breaks the statement`.
- An element with every character of the allowed pattern (`Aa9_.:-Zz0`) reads back unchanged: `an element with every character of the allowed pattern reads back unchanged`.
- A session-start row (`kind = 1`) never shows as a row, and `octo:session-start` never appears: `a session start row (kind = 1) never shows as a row, and octo-session-start never appears`.
- The route reads the app row and the level 3 statement inside one `database.read { }` block (`loadElementTotals`), so both statements share one call. Pull request #127 (issue #18) merged to `main` on `6154083`, and it made `SqliteDatabase.read { }` one read transaction; the doc comment of `loadElementTotals` now states this.
- The new HTTP test `GET with userId answers with Cache-Control no-store, no CORS header, and application-json` locks the response headers of this route.
- The new HTTP test `an element with HTML text reaches the response as one JSON string value` locks the JSON round trip of a hostile element text.
- The new HTTP test `an app id with a leading zero or a leading plus sign still reaches the same app` records the accepted forms `01` and `+1`.

## The query plan (recorded per the review rule)

`EXPLAIN QUERY PLAN` on `ELEMENT_TOTALS_SQL` gives three lines:

1. `SEARCH event USING COVERING INDEX event_agg (app_id=? AND user_id=?)`
2. `USE TEMP B-TREE FOR count(DISTINCT)`
3. `USE TEMP B-TREE FOR ORDER BY`

The base row scan uses the covering index `event_agg`, on the two leading equality terms `app_id` and `user_id`. SQLite adds two more steps for the `COUNT(DISTINCT session_id)` term and for the `ORDER BY` on the aggregate `clicks`. Both steps run on the small, already-filtered row set of one app and one user.

## Test count

`ElementsRouteTest` holds 27 tests after the correction round. `./gradlew :monitor:backend:test` passes for the whole module, 206 tests in total, run two times in sequence after the merge of `origin/main`.

- The route rejects the parameter `sessionId` with 400 today. Issue #113 owns that filter; the issue's own comment history confirms this split.
- The 400 message is "Send exactly one of userId or anonymous=true." The 404 message is "The app is not registered." The 400 message for a non-numeric app id is "The app id must be a whole number." These three sentences match the fixed-sentence pattern of the existing registry routes.
- The route does its own timestamp formatting, and does not import a formatter from the level 1 route (issue #18, pull request #127), because the two routes must not share a file, even now that pull request #127 is on `main`.
- MAJOR 2 of the security review (a bad percent escape gives 500) is not fixed here. Issue #148 owns `Application.kt`; this pull request makes no change there. See the correction round comment.

Closes #51

Generated with Claude Code


xamcross:

I made my own clone. I ran `gh pr checkout 144` in it. I ran each command and each probe myself. I did not trust the pull request text.

`./gradlew :monitor:backend:test` ran one time on the branch. The result: **190 tests, 0 failures, 0 skipped, 18 test classes**. `ElementsRouteTest` holds 22 of the 190 tests. The five CI checks pass.

---

## 1. The acceptance criteria

| Criterion | Result | My own evidence |
| --- | --- | --- |
| Exactly one of `userId` and `anonymous=true` passes; each other request gets 400 | PASS, with one gap | The seven `parseFilter` tests pass. A repeated parameter still passes; see MINOR 2. |
| The SQL uses `user_id IS ?` | PASS | `ElementsRoute.kt:58`. The value is a bound parameter at `ElementsRoute.kt:132`. |
| The anonymous rows are correct | PASS | My probe seeded 100 000 rows, half with `user_id IS NULL`. `anonymous=true` answered 250 rows; a named user answered the other 250. |
| `rows` holds `element`, `clicks`, `sessions`, `lastInteractionAt` | PASS | Live body: `{"rows":[{"element":"checkout.save","clicks":1,"sessions":1,"lastInteractionAt":"2023-11-14T22:13:20.000Z"}]}` |
| `lastInteractionAt` is UTC ISO 8601 with milliseconds | PASS | The same body, and the test at `ElementsRouteTest.kt:113`. |
| The order is `clicks DESC, element ASC` | PASS | `ElementsRoute.kt:59`, plus the passing test at `ElementsRouteTest.kt:128`. |
| An unknown app id gives 404 | PASS | My probe: app id `-1` and app id `0` each gave `404 {"error":"The app is not registered."}`. |
| The captured log holds no query string | PASS at run time, FAIL as a test | See MAJOR 1. The behaviour is correct today. The test cannot fail. |
| `./gradlew :monitor:backend:test` passes | PASS | 190 tests, 0 failures. |
| A comment on the issue records the test result | PASS | The comment is on issue #51. |

---

## 2. Check 1 — injection

The app id goes through `toLongOrNull()`. The filter value goes to `setString` or to `setNull`. No input reaches the SQL text.

I sent each value over HTTP against a live server. No answer was 500. No body held a stack trace or a class name.

| Input | Answer |
| --- | --- |
| `userId=o'brien`, `x'--`, `x'; DROP TABLE event; --` | 200 `{"rows":[]}` |
| `userId=%25`, `%25user%25`, `_`, `user_1` | 200 `{"rows":[]}` (no LIKE, thus no wildcard) |
| `userId=%00a` (a NUL) | 200 |
| `userId=a%0Ab`, `a%0D%0AX-Injected:%201` | 200, and the answer held no injected header |
| `userId=` + 10 000 characters | 200 over the test host; a real Netty engine answers `400 An HTTP line is larger than 4096 bytes.` |
| `userId=<a valid UUID>` | 200 `{"rows":[]}` |
| `userId=x' UNION ALL SELECT name,1,1,1 FROM app --` | 200 `{"rows":[]}` |
| `appId=1e3`, `9223372036854775808`, `1;`, `1.0`, `1 OR 1=1` | 400 `The app id must be a whole number.` |
| `appId=-1`, `appId=0` | 404 `The app is not registered.` |
| `appId=01`, `appId=+1` | 200 (see MINOR 6) |

After the hostile requests, a plain request still answered 200. The table `event` survives.

One input breaks the rule: a bad percent escape gives 500. See MAJOR 2.

## 3. Check 2 — the leak of the user id through a log

**The short answer: no log holds the user id today, and the design accepts the user id in the query string, but the guard is one test that cannot fail, and issue #57 can break the rule.**

The long answer:

- The monitor installs no `CallLogging` plugin. I captured **every** level of the root logger for three requests to this route. The result was five `INFO` lines of the Ktor start-up. No line held `example.com`, no line held `userId`, and no line held the query string.
- No proxy exists. The server binds to `127.0.0.1` (D12), thus no proxy writes an access log of this URL.
- D13 puts the user id in the query string on purpose. The design accepts it. The monitor shows the data to the owner on a loopback address. I agree with the decision inside that model.
- The residual risk is a browser and a future log. The query string stays in the browser history of the owner. `Cache-Control: no-store` stops a cached copy; I saw the header on the 200, the 400, the 404 and the 500 of this route. Issue #57 adds a file log with rotation. The default Ktor `CallLogging` format prints the URI with the query string. That change would write `userId=<personal data>` to a file on disk, against D15.
- The fix for this pull request: make the sentinel test real (MAJOR 1). The fix for issue #57: print the method, the path and the status only, never the query string, and add a test that proves it.

## 4. Check 3 — the answer

I read each header of a live 200 answer:

- `Cache-Control: no-store` — present.
- `X-Content-Type-Options: nosniff` — present.
- `Content-Security-Policy` — present, with the exact text of D12.
- `Access-Control-Allow-Origin`, `Access-Control-Allow-Credentials`, `Vary` — each one absent.
- The Host check of D12 holds: a request with `Host: evil.example` gets 403, and a request with no Host header gets 403.
- `Content-Type: application/json`, with no charset. See MINOR 1.

An element text with HTML goes out as a JSON string only. The live body was:
`{"rows":[{"element":"<script>alert(1)</script>",...},{"element":"a\"b",...}]}`
The quote character is escaped. The content type is JSON, and `nosniff` stops a browser sniff. No body held HTML.

## 5. Check 4 — the reserved element

I seeded two `kind = 1` rows with the element `octo:session-start`, one for a named user and one for an anonymous session. Neither row showed. The literal `kind = 0` at `ElementsRoute.kt:58` keeps them out. PASS.

I also seeded a `kind = 0` row with that same element name. It **does** show. This is not a defect of this route; the route follows section 6 exactly. See MINOR 5 for the reason and for the follow-up.

## 6. Check 5 — the cost

I built one app with 100 000 rows, 500 distinct elements, 1 000 sessions, half of the rows anonymous.

- `EXPLAIN QUERY PLAN` with a literal value, for both filter forms:
  `SEARCH event USING COVERING INDEX event_agg (app_id=? AND user_id=?)`, then two temp B-trees. No full table scan, on the named form and on the anonymous form.
- A named user: 250 rows in 33 ms. The anonymous form: 250 rows in 22 ms. A user id that matches no row: 0 rows in under 1 ms.
- Over HTTP: 200 in 120 ms, 24 705 bytes.
- A page size beyond the design: `?userId=user-1&page=1&pageSize=1000000` is ignored, and the answer stays the same. D13 gives level 3 no pages, thus an answer with no limit follows the design. The row count has the distinct element count of one user as its ceiling, not the row count of the table. I accept it.

## 7. Check 6 — do the tests fail when I break the code?

I broke the code four times, I ran the tests, and I restored each file. `git status` is clean now.

| The break | The result |
| --- | --- |
| The bound parameter: the user id goes into the SQL text | FAIL, as it must: `a user id with a quote character stays a bound parameter, and never breaks the statement` |
| A log line: `log.info("elements request {}", call.request.queryParameters.toString())` in the route | **No test of the branch failed.** 201 tests ran; only my own probe class failed. See MAJOR 1. |
| The `Cache-Control` header: the guard no longer sets it | FAIL, as it must: 6 tests of `RequestGuardTest`, `RequestGuardRawSocketTest` and `ApplicationErrorHandlingTest`. **No test of `ElementsRouteTest` failed.** See MINOR 3. |
| The reserved element: `AND kind = 0` goes out of the statement | FAIL, as it must: `a session start row (kind = 1) never shows as a row...` and `EXPLAIN QUERY PLAN shows the covering index event_agg...` |

---

## The findings

### MAJOR 1 — the sentinel log test cannot fail

`monitor/backend/src/test/kotlin/octometer/monitor/elements/ElementsRouteTest.kt:302`, with `monitor/backend/src/test/kotlin/octometer/monitor/LogCapture.kt:21`.

The test `a request with a userId query string writes no log line that holds that value` reads the list of `captureErrorLogEvents`. That helper keeps only `Level.ERROR`. An `INFO` line or a `WARN` line with the user id passes the test.

My evidence: I added one line to `ElementsRoute.kt` that logs the whole query string at `INFO`. I ran `./gradlew :monitor:backend:test --tests "octometer.monitor.*"`. Every test of the branch passed. D15 asks for a sentinel test on the captured log; this one does not search the captured log.

The pull request text claims "Two tests confirm the captured log holds no query string and no user id." One of the two tests cannot fail. Please correct the text too.

The fix. Add a helper to `LogCapture.kt`:

```kotlin
/** Runs [block] and returns its result with every log event of any level. */
suspend fun <T> captureAllLogEvents(block: suspend () -> T): Pair<T, List<ILoggingEvent>> {
    val root = LoggerFactory.getLogger(Logger.ROOT_LOGGER_NAME) as Logger
    val appender = ListAppender<ILoggingEvent>()
    appender.start()
    root.addAppender(appender)
    try {
        return block() to appender.list.toList()
    } finally {
        root.detachAppender(appender)
        appender.stop()
    }
}
```

Then change the test at line 302 to `captureAllLogEvents`, and assert two facts:

```kotlin
assertTrue(events.none { it.formattedMessage.contains("the-secret-user-id") })
assertTrue(events.none { it.formattedMessage.contains("userId=") })
```

Keep `captureErrorLogEvents` for the ERROR count at line 254.

### MAJOR 2 — a bad percent escape in the query string gives 500

`monitor/backend/src/main/kotlin/octometer/monitor/elements/ElementsRoute.kt:69`, with `monitor/backend/src/main/kotlin/octometer/monitor/Application.kt:87`.

`call.parameters` merges the path parameters and the query parameters. Ktor decodes the query string there. A bad escape makes Ktor throw `BadRequestException`. The catch-all `exception<Throwable>` of `Application.kt` turns it into 500.

My evidence, over a raw socket against a real Netty server:

```
GET /api/apps/1/elements?userId=%zz HTTP/1.1
-> HTTP/1.1 500 Internal Server Error
   {"error":"The server had an internal error."}
   log: [ERROR] octometer.monitor.Application :: An unhandled exception reached the server. BadRequestException
```

Two rules break. Lesson 4 of the backend brief says only an invalid request gives 400; a client mistake must never give 500. Lesson 3 and MAJOR 1 of the third Ktor review say a client mistake writes no ERROR line. An ERROR line of a client mistake also hides a real fault from the owner.

The body holds no data and no stack trace, thus nothing leaks. That is why this is MAJOR and not BLOCKER.

I state plainly that the defect is on `main` too: `GET /api/health?x=%zz` gives the same 500. This route is the first route whose design puts user input in the query string, thus it is the first route a person reaches with a hand-typed or a copied URL. One handler fixes every route:

```kotlin
// In Application.kt, inside install(StatusPages), before exception<Throwable>:
exception<BadRequestException> { call, _ ->
    call.respond(HttpStatusCode.BadRequest, ErrorBody("The request is not valid."))
}
```

Add the import `io.ktor.server.plugins.BadRequestException`. Add one test: a raw socket sends `?userId=%zz`, and it asserts 400 plus an empty ERROR list. Open a separate issue for the health route if you prefer a smaller diff here; please say which way you choose.

### MINOR 1 — the JSON content type carries no charset

The live answer holds `Content-Type: application/json`, with no `charset=utf-8`. RFC 8259 registers no charset parameter for `application/json`, and `X-Content-Type-Options: nosniff` stops a browser sniff, thus the risk today is near zero. Add the charset only if the review checklist demands it. My judgement: keep the current value, and record the reason in the pull request text.

### MINOR 2 — a repeated query parameter passes

`monitor/backend/src/main/kotlin/octometer/monitor/elements/ElementsRoute.kt:109`.

`parseFilter` counts the keys, not the values. Two live results:

- `?userId=user-1&userId=user-2` gives 200 with the rows of `user-1`.
- `?anonymous=true&anonymous=false` gives 200 with the anonymous rows.

D13 asks for exactly one of the three parameters. A request with two values of one name is not one parameter. No privilege rises, because the monitor has no login, thus this is MINOR.

The fix:

```kotlin
val values = listOf("userId", "anonymous", "sessionId").sumOf { parameters.getAll(it)?.size ?: 0 }
if (values != 1) return null
```

Add two tests, one for each live result above.

### MINOR 3 — the new route asserts no response header of its own

`monitor/backend/src/test/kotlin/octometer/monitor/elements/ElementsRouteTest.kt`.

D13 names three facts for each route of D44: the Host check of D12, `Cache-Control: no-store`, and no CORS header. I confirmed all three at run time on this route. `RequestGuardTest` proves them on `/api/health` only. When I removed the header from the guard, no test of `ElementsRouteTest` failed.

The fix: add one HTTP test on `/api/apps/{appId}/elements` that asserts `no-store`, asserts a null `Access-Control-Allow-Origin`, and asserts 403 for `Host: evil.example`.

### MINOR 4 — no test proves the HTML safety of an element text

The element text comes from a page of the app. The answer must carry it as a JSON string. The behaviour is correct; no test locks it. Add one test: seed the element `<script>alert(1)</script>`, read the body, and assert that the JSON parse gives that exact string.

### MINOR 5 — a `kind = 0` row named `octo:session-start` shows

This route follows section 6 exactly, thus this is not a defect of this pull request. I report it because check 4 of my task names the reserved element.

`EventStore.insertEvents` (`store/EventStore.kt:15`) names no `kind` column, thus every row that the store writes today gets the default `kind = 0`. Migration `002_first_page.sql` adds the column with the default 0, and it writes no backfill for the rows of version 1.0. Contract rule C38 says a reader of version 1.0 counts `octo:session-start` as a click.

My evidence: a seeded `kind = 0` row with that element name shows as a row with one click.

The fix belongs to the reader issue, not here. Please open one issue: the reader sets `kind = 1` for `octo:session-start`, and migration `003` backfills `UPDATE event SET kind = 1 WHERE element = 'octo:session-start'`.

### MINOR 6 — two forms of one app id both pass

`?appId=01` and `?appId=+1` each reach the app 1, because `toLongOrNull` accepts both forms. The monitor has no login, thus nothing rises. Note it in the pull request text, or reject a value that does not match `^(0|[1-9][0-9]*)$`.

### MINOR 7 — one comment sentence is too long

`monitor/backend/src/test/kotlin/octometer/monitor/elements/ElementsRouteTest.kt:172`.

"Both steps run on the small, already-filtered row set of one app and one user, not on the whole table, so the covering index still does the heavy filter work." That sentence has 30 words. Rule 8 of the backend brief sets 25 words as the maximum. Split it into two sentences.

---

## The scope and the hygiene

- The diff adds three files and changes no other file: `ApiRoutes.kt` (+2 lines), `ElementsRoute.kt`, `ElementsRouteTest.kt`. The route lives in its own package, as the brief asks.
- The route adds one line to `ApiRoutes.kt` and no line to `Application.kt`, as the brief asks.
- The test uses `testDataDir()`, thus it leaves no folder in the temp folder of the machine.
- The commit body and the comments read as Simplified Technical English, with the one exception of MINOR 7.

## The summary

The SQL is safe. The index plan is right. The reserved `kind = 1` row stays out. The headers are right. Nothing leaks a user id today.

Two things must change before an approval:
1. Make the sentinel log test able to fail (MAJOR 1).
2. Give a bad percent escape a 400, with no ERROR line (MAJOR 2).

I deleted my clone after this review.

xamcross:
**Maintainer decisions after the security review (2026-09-22).**

- MAJOR 1 (the sentinel log test cannot fail): correct in this pull request. Pull request #147 adds a helper for each log level to `LogCapture.kt`. The correction round of this pull request uses a helper with the same name and the same body, thus the merge of the two branches gives no conflict of substance.
- MAJOR 2 (a bad percent escape gives 500): the defect is on `main` for each route, with the cause in `Application.kt`. Issue #148 owns the fix. This pull request does not change `Application.kt`, because seven open branches change `monitor/backend` today.
- The MINOR on a stored `octo:session-start` row with `kind = 0`: the reader of #110 sets `kind = 1` for that element. No change here.
- The MINOR on a repeated query parameter (`anonymous=true&anonymous=false` gives 200): correct it in this pull request. Exactly one filter parameter, one time.

The correction round starts after the SQL review.

xamcross:
**Review (SQL and API engineer): CHANGES REQUESTED**

I made my own clone. I ran `gh pr checkout 144` in it. Each line below holds evidence that I saw.
The build reports `sqlite_version() = 3.53.4`. The default time zone of the machine is
`Europe/Kiev`, thus each UTC result below is a real proof.

## The acceptance criteria

1. **Exactly one of `userId` and `anonymous=true`, else 400 — PARTIAL.** I sent 13 requests to
   `GET /api/apps/{appId}/elements`, plus one for an unknown app id and one for a non-numeric
   app id. No answer is 500. Each 400 body holds the fixed sentence
   `{"error":"Send exactly one of userId or anonymous=true."}`. The answers: no parameter 400,
   `anonymous=false` 400, `anonymous=TRUE` 400, `anonymous=` 400, `anonymous` 400, `userId=`
   400, `userId` plus `anonymous=true` 400, `sessionId=<uuid>` 400, `colour=red` 400,
   `userId=user-1` 200, `userId=user-1&colour=red` 200. One answer is wrong:
   `?userId=user-1&userId=user-2` gives 200. See MAJOR 2.
2. **The SQL uses `user_id IS ?` — PASS.** `ElementsRoute.kt:58`.
3. **The anonymous rows are correct — PASS.** I seeded two apps, two users with the same
   element, two anonymous clicks in one session, and two session-start rows. `anonymous=true`
   gives one row: `banner.close`, `clicks=2`, `sessions=1`. A named user's row never appears.
4. **The response holds `rows` with the four names — PASS.** The body of a 200 answer:
   `{"rows":[{"element":"checkout.save","clicks":2,"sessions":2,"lastInteractionAt":"2023-11-14T22:13:20.500Z"}, ...]}`.
5. **`lastInteractionAt` is UTC ISO 8601 with milliseconds — PASS.** The value above is UTC on a
   machine with the zone `Europe/Kiev`. `TIMESTAMP_FORMATTER` (`ElementsRoute.kt:63`) holds
   `withZone(ZoneOffset.UTC)`, thus no test depends on the zone of the machine. I also replaced
   `ZoneOffset.UTC` with `ZoneId.systemDefault()`; the test of the form then failed.
6. **The order is `clicks DESC, element ASC` — PASS.** My rows came back as
   `checkout.save` (2 clicks), `Aa9_.:-Zz0` (1 click), `a.b` (1 click).
7. **An unknown app id gives 404 — PASS.** `GET /api/apps/999999/elements?anonymous=true` gives
   404 with `{"error":"The app is not registered."}`. A non-numeric app id gives 400 with
   `{"error":"The app id must be a whole number."}`.
8. **The captured log holds no query string — PASS for the route.** The route writes no log
   line. The reach of the test is smaller than D15 asks; see MINOR 4.
9. **`./gradlew :monitor:backend:test` passes — PASS.** I ran it two times in sequence, the
   second run after `cleanTest`. Both runs: 190 tests, 0 failures, 0 skipped. `ElementsRouteTest`
   holds 22 of the 190 tests. The count agrees with the pull request text.
10. **A comment on the issue records the test result — PASS.** The comment is on issue #51.

## The counts (my own seed data)

- `clicks` counts a row with `kind = 0` only. Proof: I replaced `COUNT(*) AS clicks` with
  `COUNT(DISTINCT session_id) AS clicks`; the test `a second app never mixes its rows into the
  first app's totals` failed.
- A session start row (`kind = 1`) shows no row, and it changes no count. My seed holds a start
  row for a named user and a start row for an anonymous session. Neither one appears in a
  result.
- An element with each character of the pattern of rule C33 (rule C33 names the pattern of rule
  C4, `[A-Za-z0-9_.:-]+`) reads back unchanged. My seed used `Aa9_.:-Zz0`.
- A second app never mixes its rows into the first app's totals.
- `sessions` holds `COUNT(DISTINCT session_id)` in the SQL text, but no test defends the value.
  See MAJOR 1.

## The query plan

I ran `EXPLAIN QUERY PLAN` on `ELEMENT_TOTALS_SQL` myself, with a bound app id and with two
forms of the second parameter.

With a bound user id, and also with a bound null, the plan is the same:

```
SEARCH event USING COVERING INDEX event_agg (app_id=? AND user_id=?)
USE TEMP B-TREE FOR count(DISTINCT)
USE TEMP B-TREE FOR ORDER BY
```

This agrees with the pull request text, line for line. `user_id IS ?` with a bound value keeps
the covering index `event_agg` on this build.

Two more plans, for the record:

- The same statement without the `kind` filter falls to a different index:
  `SEARCH event USING COVERING INDEX event_session (app_id=?)`, plus
  `USE TEMP B-TREE FOR GROUP BY`. The literal thus earns its place.
- The same statement with `AND kind = ?` and a bound `0` gives the plan of `event_agg` too, on
  this build. See MINOR 2.

## The route pattern and the shared files

- The route follows `AppTotalsRoute.kt` of pull request #127: the same `internal const val` SQL
  text, the same `TIMESTAMP_FORMATTER`, the same `fun Route.x(database: SqliteDatabase)` shape,
  and one `database.read { }` block for each statement of the handler. PASS.
- The route shares no file with issue #50. It lives in the new package
  `octometer.monitor.elements`. PASS.
- `ApiRoutes.kt` grows by one import and one line. PASS.
- `loadElementTotals` runs the app check and the level 3 statement in one `database.read { }`
  block (`ElementsRoute.kt:96`). PASS. Pull request #127 makes that block one read transaction;
  that change is not on `main` yet.
- The diff holds three files, all of them source files. It holds no secret, no database file,
  and no build output.

## The tests

I broke the code 15 times, and I ran `ElementsRouteTest` after each break. Eleven breaks gave
the right failure:

| The break | The test that failed |
| --- | --- |
| `ORDER BY clicks ASC` | `the order is clicks DESC then element ASC` |
| drop the literal `kind = 0` | the plan test, and the session-start test |
| `user_id = ?` in place of `user_id IS ?` | `anonymous=true reads the rows with user_id IS NULL...` |
| `MIN(ts)` in place of `MAX(ts)` | `lastInteractionAt is the UTC ISO 8601 form...` |
| drop `.SSS` from the pattern | `lastInteractionAt is the UTC ISO 8601 form...` |
| `ZoneId.systemDefault()` in place of UTC | `lastInteractionAt is the UTC ISO 8601 form...` |
| accept each `anonymous` value | `anonymous=false is invalid` |
| `presentKeys < 1` in place of `!= 1` | `userId and anonymous together are invalid` |
| the app row always exists | `an unknown app id gives AppNotFound`, and the 404 test |
| `COUNT(DISTINCT session_id) AS clicks` | `a second app never mixes its rows into the first app's totals` |
| an ERROR log line with the user id | `a request with a userId query string writes no log line...` |

Four breaks passed each test:

- `COUNT(session_id) AS sessions`, and `COUNT(DISTINCT event_id) AS sessions`. Only the plan
  test failed. See MAJOR 1.
- An INFO log line that holds the user id. See MINOR 4.
- `GROUP BY element, kind` in place of `GROUP BY element`. This break changes no result,
  because the filter holds `kind = 0`. It is not a defect of a test.

**The temporary folders.** I listed `%TEMP%` before a clean run of `ElementsRouteTest` and after
it. The run added 0 folders. `testDataDir()` registers each root, and the shutdown hook deletes
it. No database and no folder stays open after a test. PASS.

## The findings

### MAJOR 1 — no test defends the value of `sessions`

`monitor/backend/src/test/kotlin/octometer/monitor/elements/ElementsRouteTest.kt:49`

I replaced `COUNT(DISTINCT session_id) AS sessions` with `COUNT(DISTINCT event_id) AS sessions`
in `ElementsRoute.kt:56`. Only the query plan test failed. Each behaviour test passed. I then
replaced it with `COUNT(session_id) AS sessions`, with the same result. The reason: no test
gives one user two clicks in one session and then asserts `sessions`. In the test at line 49,
`clicks` and `sessions` both equal 2, thus the two columns are not separate. The `sessions`
value of the level 3 row is a criterion of this issue, and only a plan string defends it today.

**The fix.** In `two users of one app give one row for their shared element, with the right
clicks and sessions`, add a third click of `user-1` in the session `s1`, with a new event id.
Then assert `assertEquals(3L, row.clicks)` and `assertEquals(2L, row.sessions)`.

### MAJOR 2 — a repeated `userId` gives 200, not 400

`monitor/backend/src/main/kotlin/octometer/monitor/elements/ElementsRoute.kt:109`

`parseFilter` counts the keys that the query string holds, and not the values. `Parameters`
reports one key for `?userId=user-1&userId=user-2`, thus `presentKeys` is 1. Line 116 then
reads `parameters["userId"]`, which returns the first value. I sent that request and I got 200
with the rows of `user-1`. The route drops the second value without a word. D13 asks for
exactly one of the three parameters, else 400, and the criterion of this issue repeats it. A
caller who sends two user ids gets the data of one of them, and no signal about the other.

**The fix.** Count the values, not the keys:

```kotlin
val presentValues = listOf("userId", "anonymous", "sessionId")
    .sumOf { key -> parameters.getAll(key)?.size ?: 0 }
if (presentValues != 1) return null
```

Add a test: `parseFilter(Parameters.build { append("userId", "a"); append("userId", "b") })`
returns null.

One note on the grade. If you read D13 as a rule about the name of the parameter only, then this
finding is a MINOR. I read it as a rule about the request, because the route must never choose
between two user ids without a word.

### MINOR 1 — an assertion that cannot fail

`monitor/backend/src/test/kotlin/octometer/monitor/elements/ElementsRouteTest.kt:157`

`tableStillExists` reads `executeQuery("SELECT COUNT(*) FROM event").use { it.next() }`. A
`COUNT(*)` result always holds one row, thus `next()` always returns true. A dropped table makes
the statement throw, and the test then errors before the assertion. The assertion adds nothing.

**The fix.** Read the count and assert the number:
`assertEquals(2, count, "the event table keeps both rows")`.

### MINOR 2 — a comment states a reason that this build does not support

`monitor/backend/src/main/kotlin/octometer/monitor/elements/ElementsRoute.kt:52`

The comment says "A bound parameter does not give the same plan on each SQLite build". I ran the
same statement with `AND kind = ?` and a bound `0` on this build. The plan is the same
`SEARCH event USING COVERING INDEX event_agg (app_id=? AND user_id=?)`. The measured reason for
the literal is different: a statement with **no** `kind` filter falls to `event_session` plus a
temp B-tree for the GROUP BY.

**The fix.** Replace the second sentence with the measured fact, for example: "A statement
without the `kind` filter falls to the index `event_session`. Rule M4 keeps the literal in the
SQL text."

### MINOR 3 — a redundant assertion

`monitor/backend/src/test/kotlin/octometer/monitor/elements/ElementsRouteTest.kt:101`

Line 99 asserts one row, and line 100 asserts that its element is `checkout.save`. Line 101 then
cannot fail. **The fix.** Delete line 101.

### MINOR 4 — the log sentinel reads only the ERROR level

`monitor/backend/src/test/kotlin/octometer/monitor/elements/ElementsRouteTest.kt:302`

`captureErrorLogEvents` keeps an event of the ERROR level only (`LogCapture.kt:21`). I added an
INFO line with the user id to the handler; each test passed. I added the same line at the ERROR
level; the test failed. D15 says a log line never holds a user id, at any level. The route
writes no log line today, thus the behaviour is right, and only the reach of the test is small.

**The fix.** Add a helper beside `captureErrorLogEvents` that returns each captured event, and
assert on that list here. `LogCapture.kt` is a shared file, thus add the helper; do not change
the old one.

### MINOR 5 — two comment sentences are longer than 25 words

- `monitor/backend/src/main/kotlin/octometer/monitor/elements/ElementsRoute.kt:105` — the second
  sentence holds 26 words, and a semicolon joins two statements. Split it.
- `monitor/backend/src/test/kotlin/octometer/monitor/elements/ElementsRouteTest.kt:172` — the
  last sentence holds 30 words, and "the heavy filter work" is a figure of speech. Split it, and
  write "the covering index still selects the rows".

### MINOR 6 — no test asserts the headers of the new route

The brief asks each new route for a test of `Cache-Control: no-store` and of the absent CORS
header. I checked both myself: the answer holds `Cache-Control: no-store`, and it holds no
`Access-Control-Allow-Origin`. `installRequestGuard` sets the header for each `/api/` path, and
`RequestGuardTest` covers that rule. **The fix** is optional: add one assertion of the two
headers to `GET with userId answers 200...`.

## The merge check (a fact, not a finding)

- `origin/main` (`3186569`): the branch already sits on it. `git merge origin/main` reports
  "Already up to date".
- The branch of pull request #127 (`git fetch origin pull/127/head:pr127`): one conflict, in
  `monitor/backend/src/main/kotlin/octometer/monitor/ApiRoutes.kt`. The conflict holds two
  hunks: the import line, and the call line in `apiRoutes`. Each side keeps its own line, thus
  the fix is a two-line merge. I aborted both merges.

## The verdict

The SQL is right, the plan is right, and the route follows the pattern of #127. Two items stop
an approval: MAJOR 1 (the `sessions` value has no test that can fail) and MAJOR 2 (a repeated
`userId` gives 200). Fix the two, and I approve.

xamcross:
**Correction round 1**, after the SQL review, the security review, and the maintainer decisions.

Commits: `1c52949` (the correction round), `5e2eef2` (the merge of `origin/main`, pull request #127, commit `6154083`).

## MAJOR findings

1. **SQL MAJOR 1 — no test defended the value of `sessions`.**
   File: `monitor/backend/src/test/kotlin/octometer/monitor/elements/ElementsRouteTest.kt:57`.
   Fixed. The test now gives `user-1` a third click in the same session `s1`, and asserts `clicks = 3` and `sessions = 2`. I replaced `COUNT(DISTINCT session_id)` with `COUNT(session_id)` in `ElementsRoute.kt`, ran the test, and saw it fail. I restored the code and ran the test again; it passed.

2. **SQL MAJOR 2, and the matching security MINOR — a repeated `userId` or `anonymous` value passed.**
   File: `monitor/backend/src/main/kotlin/octometer/monitor/elements/ElementsRoute.kt:115`.
   Fixed. `parseFilter` now counts each value of `userId`, `anonymous`, and `sessionId` with `parameters.getAll(key)?.size`, not each distinct key. Two new tests: `a repeated userId is invalid, even though the key appears once` and `anonymous=true and anonymous=false together are invalid`. I reverted the count to `parameters.contains(it)`, ran both tests, and saw them fail. I restored the fix and ran the tests again; they passed.

3. **Security MAJOR 1 — the log sentinel test read the ERROR level only.**
   Files: `monitor/backend/src/test/kotlin/octometer/monitor/LogCapture.kt` (new helper `captureLogEvents`), `monitor/backend/src/test/kotlin/octometer/monitor/elements/ElementsRouteTest.kt:418`.
   Fixed. `LogCapture.kt` gets the helper `captureLogEvents`, with the same name and the same body as pull request #147 (I read `pr147:monitor/backend/src/test/kotlin/octometer/monitor/LogCapture.kt` with `git show`), so the two branches merge with no conflict of substance. The test now reads `captureLogEvents` and asserts on every level, plus a second assertion on the literal `userId=`. I added an INFO line that logged the query string to the route, ran the test, and saw it fail. I removed the line and ran the test again; it passed.

4. **Security MAJOR 2 — a bad percent escape gives 500.**
   Not fixed here. Per the maintainer decision, the cause sits in `Application.kt`, on `main`, for every route. Issue #148 owns the fix. This pull request makes no change to `Application.kt`, because seven open branches change `monitor/backend` today.

## MINOR findings

- SQL MINOR 1 (a dead assertion, `ElementsRouteTest.kt:157` at review time): fixed. The test now reads the row count and asserts the number, `assertEquals(2, count, ...)`, in place of an assertion on `ResultSet.next()` that could never fail.
- SQL MINOR 2 (the comment at `ElementsRoute.kt:52` stated an unmeasured reason): fixed. The comment now states the measured fact: a statement without the `kind` filter falls to the index `event_session`.
- SQL MINOR 3 (a redundant assertion, `ElementsRouteTest.kt:101` at review time): fixed. Deleted; the row's single-element assertion already covers it.
- SQL MINOR 4 (the log sentinel's reach was smaller than D15 asks): fixed by the correction of security MAJOR 1 above; the two findings shared one cause.
- SQL MINOR 5 / security MINOR 7 (two comment sentences over 25 words, `ElementsRoute.kt:105` and `ElementsRouteTest.kt:172` at review time): fixed. Both are split into shorter sentences; the idiom "the heavy filter work" is now "the covering index still selects the rows".
- SQL MINOR 6 / security MINOR 3 (no header test on the new route): fixed. New test `GET with userId answers with Cache-Control no-store, no CORS header, and application-json`.
- Security MINOR 1 (the JSON content type carries no charset): not corrected, by the reviewer's own judgement ("keep the current value, and record the reason"). Recorded in the pull request text: the other routes of this module give no charset either.
- Security MINOR 2: this is SQL MAJOR 2 under a different grade; corrected as MAJOR 2 above.
- Security MINOR 4 (no test of an HTML element text as a JSON string): fixed. New test `an element with HTML text reaches the response as one JSON string value`.
- Security MINOR 5 (a `kind = 0` row named `octo:session-start` shows): not changed here. This belongs to issue #110, which sets `kind = 1` for that element at read time.
- Security MINOR 6 (the app id forms `01` and `+1` both pass): fixed. New test `an app id with a leading zero or a leading plus sign still reaches the same app` records the accepted behaviour; the monitor has no login, thus no privilege rises.

## The merge

`origin/main` moved to `6154083` (pull request #127, issue #18) while this pull request was open. I merged it with a merge commit, `5e2eef2`. The one conflict sat in `ApiRoutes.kt`, at the import line and the route line, exactly where the SQL reviewer's own merge check predicted it. Both lines of both sides are in the file now: `appTotals(services.database)` and `elementsRoute(services.database)`. `SqliteDatabase.read { }` is one read transaction as of that merge; the doc comment of `loadElementTotals` now states it.

## The tests

`./gradlew :monitor:backend:test` ran two times in sequence after the correction round and the merge, with JDK 21. Both runs: green, 206 tests in the module, 27 of them in `ElementsRouteTest` (was 22 before the correction round).

## CI

Frontend, JVM modules, Monitor backend on Windows, Secret scan, and Tracker all pass on commit `5e2eef2`.

xamcross:

I made a new clone. I ran `gh pr checkout 144` in it. I ran each check and each experiment myself. I restored each file after each experiment, and `git status` is clean now.

The head is `5e2eef2`. `./gradlew :monitor:backend:test` ran one time with JDK 21. The result: **206 tests, 0 failures, 0 skipped, 20 test classes**. `ElementsRouteTest` holds 27 of the 206 tests. The five CI checks pass on `5e2eef2`.

---

## MAJOR 1 — the log sentinel: CONFIRMED FIXED

I added this line to `ElementsRoute.kt`, above the `parseFilter` call:

```kotlin
org.slf4j.LoggerFactory.getLogger("octometer.monitor.elements")
    .info("elements request {}", call.request.queryParameters.toString())
```

I ran `./gradlew :monitor:backend:test --tests "octometer.monitor.*"`. The result:

```
ElementsRouteTest > a request with a userId query string writes no log line that holds that value() FAILED
    org.opentest4j.AssertionFailedError: Expected value to be true.
206 tests completed, 1 failed
```

The same experiment on the first round passed every test. The sentinel test can fail now. I removed the line, and I restored the file.

The test at `ElementsRouteTest.kt:426` reads `captureLogEvents`, and it asserts two facts: no event holds `the-secret-user-id`, and no event holds the literal `userId=`. The second assertion also catches a line that prints the query string with a different user id. I accept the form.

## The helper matches pull request #147

I ran `git fetch origin pull/147/head:pr147`, then `git show pr147:monitor/backend/src/test/kotlin/octometer/monitor/LogCapture.kt`, then a `diff` against the file of this branch. The two files are identical: the same two function names, the same two bodies, the same two doc comments, and the same order. The only difference is the line ending of the `git show` output. The merge of the two branches gives no conflict of substance, as the maintainer decision asks.

## MAJOR 2 — out of scope, accepted

The maintainer moved the 500 of a bad percent escape to issue #148. I read issue #148: it is open, and its title states the fix. This pull request changes no line of `Application.kt`. I accept the split. The cause sits on `main` for every route, thus a fix here would help this one route only.

---

## The 7 MINOR findings of the first review

| # | The finding | The state | My judgement |
| --- | --- | --- | --- |
| 1 | The JSON content type carries no charset | Kept, with a recorded reason | **Accepted.** My first review said "keep the current value, and record the reason". The shared `ContentNegotiation` gives every route of the module the same value, thus the route matches its siblings. The new test asserts the exact string `application/json`, thus a silent change of the value now breaks a test. RFC 8259 registers no charset parameter, and `nosniff` stops a browser sniff. |
| 2 | A repeated query parameter passed | Fixed | **Confirmed.** `parseFilter` counts each value with `getAll(key)?.size`. I sent each form over HTTP against a live server: `?userId=user-1&userId=user-2` gives 400, `?anonymous=true&anonymous=false` gives 400, and `?anonymous=false&anonymous=true` gives 400. A valid single filter still gives 200. |
| 3 | No header test on the new route | Fixed | **Confirmed.** I made the guard drop `Cache-Control`. The new test `GET with userId answers with Cache-Control no-store, no CORS header, and application-json` failed, and the other 26 tests passed. I restored the guard. |
| 4 | No test of an HTML element text | Fixed | **Accepted.** The test seeds `<script>alert(1)</script>`, and it asserts the exact string after a JSON parse, plus the content type and `nosniff`. See NEW MINOR 2 for one wrong sentence in its comment. |
| 5 | A `kind = 0` row named `octo:session-start` shows | Moved to issue #110 | **Accepted, with one request.** Issue #110 is open, and step 3 sets `kind = 1` for that element in the reader. The route follows section 6 exactly, thus no change belongs here. One gap stays: migration `002` adds the column with the default 0, and it writes no backfill. Please add one acceptance criterion to #110: a migration sets `kind = 1` for each stored row with the element `octo:session-start`. Contract rule C38 says a reader of version 1.0 counted that element as a click, thus a stored row of that form can exist. |
| 6 | The app id forms `01` and `+1` both pass | Recorded by a test | **Accepted.** The monitor has no login, thus no privilege rises. The new test records the behaviour, and it names the reason. A later hardening must change that test on purpose; that is the correct cost. |
| 7 | One comment sentence over 25 words | Fixed for the named sentence | **Accepted.** The sentence at `ElementsRouteTest.kt:172` is two sentences now, each under 16 words. The idiom "the heavy filter work" is gone. See NEW MINOR 1 for four new sentences. |

---

## Two new MINOR findings. Neither one blocks the merge.

### NEW MINOR 1 — four new comment sentences pass 25 words

Rule 8 of the backend brief sets 25 words as the maximum for a comment sentence. I counted the words of each new comment sentence. Four pass the limit:

- `ElementsRoute.kt:113` — "`Parameters` reports one key for `?userId=a&userId=b`, thus a count of the keys missed the second value, and the route answered 200 with one of the two user ids." 28 words. It also joins two statements with "and".
- `ElementsRouteTest.kt:290` — "MINOR 6 of the SQL review, and MINOR 3 of the security review: D13 asks each route of D44 for the Host check, `Cache-Control: no-store`, and no CORS header." 29 words.
- `ElementsRouteTest.kt:293` — "The JSON content type carries no charset parameter on the other routes of this module either (`HealthRouteTest` strips parameters before its own comparison), so this route matches the existing form." 30 words.
- `ElementsRouteTest.kt:318` — "The security review confirmed at run time that the body carries no raw HTML tag, that the content type stays `application/json`, and that `X-Content-Type-Options: nosniff` stops a browser sniff." 29 words.

The fix: split each one into two sentences. The correction round fixed the two sentences that I named, and it added four more, thus I record the pattern and not only the lines.

### NEW MINOR 2 — one comment states a fact that I did not find

`monitor/backend/src/test/kotlin/octometer/monitor/elements/ElementsRouteTest.kt:319`.

The comment says "the body carries no raw HTML tag". That is not my finding. JSON escapes the quote character, and it does not escape `<` or `>`. The live body of my first review was:

```
{"rows":[{"element":"<script>alert(1)</script>","clicks":1,...}]}
```

The tag is in the body, inside a JSON string value. The safety comes from two other facts: the content type is `application/json`, and `X-Content-Type-Options: nosniff` stops a browser sniff to HTML. A later reader of this comment could believe the serializer removes the tag, and could then drop the `nosniff` header as needless.

The fix: replace that clause. For example: "the body carries the tag inside a JSON string value, and the content type plus `nosniff` stop a browser from a render of it".

---

## The scope and the hygiene

- Commit `1c52949` changes three files: `ElementsRoute.kt`, `LogCapture.kt`, and `ElementsRouteTest.kt`. No other file changes.
- The merge `5e2eef2` keeps both lines of `ApiRoutes.kt`: `appTotals(services.database)` and `elementsRoute(services.database)`. I read the file myself.
- The doc comment of `loadElementTotals` now names the read transaction of pull request #127. That statement is true after the merge.
- The dead assertion of SQL MINOR 1 is a real number assertion now: `assertEquals(2, count, ...)`.

## The verdict

No BLOCKER of this pull request stays. No MAJOR of this pull request stays. MAJOR 2 belongs to issue #148, by the maintainer decision. The two new MINOR findings touch comment text only, and they change no behaviour.

**APPROVE.** Please correct the two new MINOR findings in the same push as any other small change, or in a follow-up commit before the merge. I do not need a third review for them.

I deleted my clone after this review.

xamcross:
**Review (second review, SQL and API engineer): APPROVE**

I made a second clone, and I ran `gh pr checkout 144` in it. I checked the commits `1c52949`
and `5e2eef2`. Each line below holds evidence that I saw. The build reports
`sqlite_version() = 3.53.4`. The default time zone of the machine is `Europe/Kiev`.

## MAJOR 1 — the `sessions` count now has a test that can fail. CONFIRMED.

`ElementsRouteTest.kt:57` gives `user-1` a third click in the session `s1`, and it asserts
`clicks = 3` and `sessions = 2`. I broke the statement two times:

| The break in `ElementsRoute.kt:57` | The test that failed |
| --- | --- |
| `COUNT(session_id) AS sessions` | `two users of one app give one row...` — `expected: <2> but was: <3>`, plus the plan test |
| `COUNT(DISTINCT session_id)` to `COUNT(DISTINCT event_id)` | the same two tests, with the same message |

The first break was invisible to each behaviour test before this round. Now a behaviour test
catches it. I restored the file after each break, and `git status` reports a clean tree.

My own seed data agrees: three clicks of one user in two sessions give `clicks=3, sessions=2`.

## MAJOR 2 — the filter counts each value. CONFIRMED.

`ElementsRoute.kt:118` reads `sumOf { key -> parameters.getAll(key)?.size ?: 0 }`. I sent 16
requests to the route. Each repeated parameter gets 400 with the fixed sentence
`{"error":"Send exactly one of userId or anonymous=true."}`:

| The request | The answer |
| --- | --- |
| `?userId=a&userId=b` | 400, the fixed sentence |
| `?anonymous=true&anonymous=false` | 400, the fixed sentence |
| `?anonymous=true&anonymous=true` | 400, the fixed sentence |
| `?userId=a&userId=a` | 400, the fixed sentence |
| `?sessionId=x&sessionId=y` | 400, the fixed sentence |

The maintainer decision asks for exactly one filter parameter, one time. The route meets it.

I also reverted the count to `parameters.contains(it)`. The two new tests then failed:
`a repeated userId is invalid, even though the key appears once` and
`anonymous=true and anonymous=false together are invalid`. I restored the file.

No regression on the other forms: `?userId=a` 200, `?anonymous=true` 200,
`?userId=a&colour=red` 200, and 400 for no parameter, `anonymous=false`, `anonymous=TRUE`,
`anonymous=`, `userId=`, `userId` plus `anonymous=true`, `sessionId=<uuid>`, and `colour=red`.
No answer is 500.

## The six MINOR findings of my first review

| --- | --- | --- | --- |
| 1 | A dead assertion on `ResultSet.next()` | FIXED | `ElementsRouteTest.kt:163` reads the row count, and line 171 asserts `assertEquals(2, count, ...)`. |
| 2 | The comment stated an unmeasured reason | FIXED | `ElementsRoute.kt:51` now says a statement without the `kind` filter falls to `event_session`. This is the fact that I measured. |
| 3 | A redundant assertion | FIXED | The old line 101 is gone. The test at `ElementsRouteTest.kt:99` keeps two strong assertions. |
| 4 | The log sentinel read the ERROR level only | FIXED | `LogCapture.kt:33` adds `captureLogEvents`, and the test at `ElementsRouteTest.kt:417` reads it. I added an INFO line with the query string; the test failed. I added a WARN line with the user id; the test failed too. Both breaks passed the old form. |
| 5 | Two comment sentences over 25 words | FIXED | `ElementsRoute.kt:107` holds three short sentences. `ElementsRouteTest.kt:186` holds two sentences of 15 and 12 words. The figure of speech "the heavy filter work" now reads "the covering index still selects the rows". |
| 6 | No header test on the new route | FIXED | The new test `GET with userId answers with Cache-Control no-store, no CORS header, and application-json` asserts the three facts. It passes. |

The new helper `captureLogEvents` is byte for byte the helper of pull request #147. I compared
it with `git show pr147:monitor/backend/src/test/kotlin/octometer/monitor/LogCapture.kt`. The
KDoc names issue #61, because that pull request owns the helper. The two branches thus merge
with no conflict of substance.

## The merge with `main`

- `origin/main` is now `6154083` (pull request #127). `git merge-base --is-ancestor origin/main
  HEAD` succeeds, and `git merge origin/main` reports "Already up to date".
- `ApiRoutes.kt` holds both routes: `appTotals(services.database)` at line 20, and
  `elementsRoute(services.database)` at line 21. Both import lines are present. CONFIRMED.
- The doc comment of `loadElementTotals` (`ElementsRoute.kt:87`) names the read transaction of
  pull request #127. CONFIRMED.
- `SqliteDatabase.read { }` now runs `BEGIN` and `COMMIT` around the block. The app check and
  the level 3 statement of `loadElementTotals` thus read one snapshot.
- I sent `GET /api/apps` in the merged tree; it answers 200. The level 1 route of #127 still
  works beside the new route.
- The diff against `origin/main` holds four files, all of them source files: `ApiRoutes.kt`,
  `ElementsRoute.kt`, `LogCapture.kt`, and `ElementsRouteTest.kt`. It holds no secret, no
  database file, and no build output.

## The tests

`./gradlew :monitor:backend:test` ran two times in sequence, each run after `cleanTest`, with
the toolchain of JDK 21. Both runs: **206 tests, 0 failures, 0 errors, 0 skipped.**
`ElementsRouteTest` holds 27 of them, and `AppTotalsRouteTest` of #127 holds 8. The count agrees
with the number that the implementer reports.

I listed `%TEMP%` before a clean run of `ElementsRouteTest` and after it. The run added 0
folders. No database and no temporary folder stays open after a test.

## The query plan (a second measurement)

The SQL text does not change in this round. I ran `EXPLAIN QUERY PLAN` again, with a bound user
id and with a bound null. Both give the same three lines:

```
SEARCH event USING COVERING INDEX event_agg (app_id=? AND user_id=?)
USE TEMP B-TREE FOR count(DISTINCT)
USE TEMP B-TREE FOR ORDER BY
```

## New MINOR findings of this round (they do not block the approval)

The corrections add new comment text. Four sentences break a language rule. Fix them in a later
round, or leave them; they change no behaviour.

1. `ElementsRoute.kt:113` — the sentence that starts "`Parameters` reports one key" holds 28
   words, and "and" joins two statements. **Fix.** Split it: "`Parameters` reports one key for
   `?userId=a&userId=b`. A count of the keys thus missed the second value. The route then
   answered 200 with one of the two user ids."
2. `ElementsRouteTest.kt:294` — the sentence that starts "The JSON content type carries no
   charset" holds 30 words. **Fix.** Split it at "either".
3. `ElementsRouteTest.kt:319` — the sentence that starts "The security review confirmed at run
   time" holds 29 words, and it joins three clauses. **Fix.** Use a vertical list of three
   items.
4. `ElementsRouteTest.kt:347` — "instead of silently allowing it" uses an `-ing` form as a noun.
   **Fix.** Write "and it does not hide the behaviour". The words "locked the fact" at line 292
   and "locks" at line 322 are figures of speech; write "asserts" in both places.

One record, with no action: the body of the commit `1c52949` says "two MINOR findings", and the
same sentence then lists four items. Do not amend a pushed commit. The pull request text holds
the correct list.

## The verdict

APPROVE. Both MAJOR findings of my first review are fixed, and I saw each fix fail when I broke
the code. Each of my six MINOR findings is fixed. The merge with `main` keeps both routes and
the read transaction. Both test runs are green with 206 tests. No BLOCKER and no MAJOR stays.

Issue #142, opened 2026-09-21
Remove each temporary test folder of the monitor backend after a run


xamcross:
## Goal

Each test of the monitor backend removes its temporary folder, thus no test secret file stays in `%TEMP%` after a green run.

## Observed on 2026-09-22

The fourth security review of #125 found this after a green test run (https://github.com/xamcross/octometer/pull/125#issuecomment-5769037345 names the review; the point is in the report of the reviewer to the maintainer):

- The test classes `MonitorServicesTest` and `ApplicationLifecycleTest` make their own temporary folders. They do not use the helper `registerTempRoot` of the other test classes.
- Folders with the names `octometer-monitor-services-test-*` and `octometer-lifecycle-broken-secrets-*` stay in `%TEMP%` after a green run. Each one holds a test secret file. The values are test values, not real secrets.

## Implementation steps

1. Change `MonitorServicesTest` to make each temporary folder through `registerTempRoot`, or through `@TempDir` of JUnit.
2. Change `ApplicationLifecycleTest` in the same way.
3. Search each other test class of `monitor/backend` for `createTempDirectory`, and correct each one that has no cleanup.
4. Run `./gradlew :monitor:backend:test` two times, and count the folders with the prefix `octometer-` in `%TEMP%` before and after each run.

## Acceptance criteria

- [x] The count of folders with the prefix `octometer-` in `%TEMP%` is the same before and after a green test run.
- [x] No test class of `monitor/backend` calls `createTempDirectory` without a cleanup.
- [x] `./gradlew :monitor:backend:test` passes with JDK 21.
- [x] A comment on this issue records the two counts.

## Related

- #15 added the two test classes (#125). This issue starts after #125 merges.

xamcross:
Pull request #169 fixes this issue.

## The two counts

Two forced runs of `./gradlew :monitor:backend:test` with JDK 21, after the fix and after a merge of `main`, each `BUILD SUCCESSFUL`, 316 tests, 0 failures, 0 errors:

- Run 1: 43 folders with the prefix `octometer-` in `%TEMP%` before the run, 43 after.
- Run 2: 43 folders before the run, 45 after.

Other agents run the same test suite on their own clones of this repository at the same time, on this machine. Their runs add and remove their own `octometer-` folders in the shared `%TEMP%` folder, so a raw count between two points in time can move even when no test leaks a folder. Run 2 shows this: a short time after the run, the folder set settled back to the same 43 names as before the run.

[...]

Issue #57, opened 2026-09-21
Write the log to files with rotation


xamcross:
**Blocked by #28.** The log rules need the poll status of #28. The Logback file config can start now.

**Goal.** The monitor writes its log to files in `logs\`, with fixed limits for the size and the age.

**Implementation steps.**
1. Confirm the name of the Logback rotation policy and its properties with the Logback documentation before you use them.
2. Add a file appender with rotation to the Logback config of `monitor/backend`.
3. Set the limits: 10 MB for each file, 14 days, and 200 MB in total.
4. Write the files to `logs/`, a sibling of `dataDir` (D3).
5. Write one log line for a status change and one for a failure. A good poll writes no line.
6. Write the tests of the acceptance criteria in `monitor/backend`.

**Acceptance criteria.**
- [ ] The files go to `logs/` in the parent folder of `dataDir`.
- [ ] The limits are 10 MB, 14 days, and 200 MB.
- [ ] A good poll writes no line.
- [ ] A status change writes one line.
- [ ] `./gradlew :monitor:backend:test` passes.
- [ ] A comment on this issue records the test result and the file list of `logs/` after one run.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, D3, D34, and D37.

xamcross:
Note on 2026-09-22 from the maintainer, from the privacy review of #147 (issue #61) and the security review of #144 (issue #51): https://github.com/xamcross/octometer/pull/147#issuecomment-5769624408 and https://github.com/xamcross/octometer/pull/144#issuecomment-5769526769

The routes of level 2, level 3, and the erasure carry a user id in the query string or in the path (design decision D13). The monitor writes no request log today, and each reviewer confirmed that no log line holds a user id. When this issue adds a file log, it must not install a request log with the default format of Ktor, because that format prints the query string. A request log, when one comes, prints the method, the status, and the route pattern only. The steps of this issue do not change.

PR #145, opened 2026-09-22
feat(monitor): serve the level 2 user totals

### LINKED ISSUE #50: Serve the level 2 totals with pages and a filter

xamcross:
## Summary

This pull request adds `GET /api/apps/{appId}/users`: the level 2 totals
of D13. Each row holds `userId`, `clicks`, `sessions`, and
`uniqueElements` for one user of one app. The route supports the filter
`q` and the page size of 50 rows, per section 6 of the design.

The count statement and the page statement run in one
`database.read { }` block, on the covering index `event_agg`. The
literal `kind = 0` sits in the SQL text of each statement, not as a
bound parameter. Section 6 of the design states the reason: "Level 2
and level 3 keep the filter `kind = 0`, else SQLite ignores the index
`event_agg`."

## Acceptance criteria

- [x] The order is `clicks DESC, user_id ASC`.
  Evidence: `UserTotalsRouteTest.the order is clicks DESC then user_id ASC` passes. A new test, `two users with the same click count break the tie on user_id ASC`, seeds a tie in clicks and asserts the exact order. Changing `user_id ASC` to `user_id DESC` made this new test fail, then the fix was restored.
- [x] The server clamps a page above the last page, and the response holds `page` and `pageCount`.
  Evidence: `a page above the last page is clamped to the last page`, `page 2 gives the remaining 10 rows`, and `more than one page gives page 1 with 50 rows and the right pageCount` all pass. A mutation test (the clamp line removed) made the clamp test fail, then the fix was restored.
- [x] `q` filters the user id with one `LIKE`, and the value is a bound parameter.
  Evidence: `q treats the percent sign and the underscore as literal characters, not wildcards` passes, with a decoy row that a naive unescaped LIKE would also match. A new test, `a hostile q value is a bound parameter, thus the rows stay unchanged and the answer is 200`, sends a quote and a `DROP TABLE` payload through `q` over HTTP and asserts 200 with the row unchanged. Concatenating the value into the SQL text of `countUserGroups` and `readUserTotalsPage`, for a moment, made both of these tests fail, then the fix was restored.
- [x] Each row holds `userId`, `clicks`, `sessions`, and `uniqueElements`. The `NULL` user is one row with `userId: null`.
  Evidence: `the NULL user is one row with userId null, and two apps do not mix their users` passes. A new test, `a non-empty q drops the row with userId null`, proves that a filter removes that row, and states the reason in a code comment and in the KDoc of the route.
- [x] `EXPLAIN QUERY PLAN` shows the index `event_agg`.
  Evidence: `EXPLAIN QUERY PLAN shows the covering index event_agg for the count and the page statement` passes. The real plan text of each of the four statements (the count statement and the page statement, with and without the filter q):
  - count (no filter): `CO-ROUTINE (subquery-1)`, `SEARCH event USING COVERING INDEX event_agg (app_id=?)`, `SCAN (subquery-1)`
  - page (no filter): `SEARCH event USING COVERING INDEX event_agg (app_id=?)`, `USE TEMP B-TREE FOR count(DISTINCT)`, `USE TEMP B-TREE FOR count(DISTINCT)`, `USE TEMP B-TREE FOR ORDER BY`
  - count (with q): the same three lines as the count statement above.
  - page (with q): the same four lines as the page statement above.

  A mutation test (the literal `kind = 0` removed from the page statement) made this test, and two others, fail, then the fix was restored. A new test, `the count statement applies q, thus pageCount reflects the filtered row count`, seeds 60 users (55 with a shared text, 3 of those with a rarer text) and asserts `pageCount: 2` unfiltered, `pageCount: 2` with the wide filter, and `pageCount: 1` with the narrow filter. Making the filtered count statement ignore `q`, for a moment, made this new test fail, then the fix was restored.
- [x] An unknown app id gives 404.
  Evidence: `an unknown app id gives 404 with a fixed sentence` and `GET of an unknown app id gives 404 over HTTP` pass, with the fixed sentence "The app is not registered." A mutation test (the exists check replaced with `if (false)`) made both tests fail, then the fix was restored.
  Evidence: the route writes no log line of any level. A rewritten test, `the captured log holds no line with the filter text, at every log level`, uses the new helper `captureLogEvents` (every log level, not only ERROR) and asserts that the joined text of every captured event holds no filter text. Adding a `log.info` call with the value of `q`, for a moment, made this test fail, then the fix was restored. The earlier version of this test read only ERROR events and held an unreachable assertion loop; both are gone. A bad `appId` and a bad `page` each give a fixed sentence with no echo of the input; see `a non-numeric app id gives 400, not 500` and `a page of 0 or a negative page gives 400`. A mutation test (the positive-page check removed) made the page test fail, then the fix was restored.
  Evidence: ran twice, both green, 206 tests total in the module (27 in `UserTotalsRouteTest`). `./gradlew build` ran once, green. See the issue comment for the full command output.
- [x] A comment on this issue records the test result and the `EXPLAIN QUERY PLAN` output.
  Evidence: posted as a separate issue comment.

## Correction round 1

The SQL review and the security review of this pull request found five
MAJOR gaps and several MINOR gaps. Each one is fixed. See the review
comment below this pull request for the full list, the fix, and the
test evidence for each finding.

## Confirm steps

None. Issue #50 names no "Confirm" step.

## Decisions made where the issue was silent

- The route sits in its own file and package, `octometer.monitor.users.UserTotalsRoute`, per the shared backend brief for parallel work. `ApiRoutes.kt` gained two lines: one for this route, one for the level 1 route of pull request #127.
- `q` matches a substring of the user id (`%q%`), not a prefix or an exact match. The issue names no shape for the match; a substring search fits a filter box best.
- An empty result still gives `pageCount: 1` and `page: 1`, so the page fields always name a real, existing page.
- A page count of zero groups still gives `pageCount: 1`, for the same reason.
- An empty `page` and an empty `q` now behave the same way: each one means the default (page 1, no filter).
- A `q` above 200 characters, or a `q` with a control character, gives 400 with a fixed sentence.

Closes #50

Generated with Claude Code


xamcross:

I made my own clone of the repository. I ran `gh pr checkout 145` in it. I ran each
check below myself with JDK 21. I did not trust the pull request text.

The threat model: a user id is personal data. This route returns a list of user ids,
and `q` searches in them. The monitor binds to a loopback address, and it has no
login (D12). I judge the route inside that model.

## The acceptance criteria

1. The order is `clicks DESC, user_id ASC`. **PASS.** The test
   `the order is clicks DESC then user_id ASC` passes in my run.
2. The page clamp, with `page` and `pageCount`. **PASS.** On my set of 100 000 rows,
   `?page=2147483647` gave 200 with `page = 90` and `pageCount = 90`.
3. `q` filters the user id with one `LIKE`, and the value is a bound parameter.
   **PASS in the code. FAIL in the tests.** See MAJOR 1.
4. Each row holds `userId`, `clicks`, `sessions`, and `uniqueElements`. The `NULL`
   user is one row. **PASS.**
5. `EXPLAIN QUERY PLAN` shows the index `event_agg`. **PASS.** I ran the four
   statements on a table of 100 000 rows, after `ANALYZE`. Each plan holds
   `SEARCH event USING COVERING INDEX event_agg (app_id=?)`.
6. An unknown app id gives 404. **PASS.**
7. The captured log holds no query string. **PASS in the code. FAIL in the tests.**
   See MAJOR 2.
8. `./gradlew :monitor:backend:test` passes. **PASS.** One run gave `BUILD SUCCESSFUL`.
   The XML results hold 186 tests, 0 failures, 0 errors, and 0 skipped tests.
9. A comment on the issue records the test result and the plan output. **PASS.**

## Check 1: injection through `q` — PASS

I sent each value below to `loadUserTotals` and to the route over HTTP:
`'`, `--`, `;`, `'; DROP TABLE event; --`, `%`, `_`, `\`, `%%%`, a NUL, a line break,
`%_` in one string, and 10 000 characters.

- Every answer was 200 with a valid page, or 400. No answer was 500. No body held a
  stack trace.
- The `event` table still held its 7 rows after the whole battery.
- `%` alone matched only the user id `50%_off`. `_` alone matched only the same row.
  The decoy row `50Xoff` never matched. Thus `%` and `_` match only themselves.
- `\` (the escape character of the code) matched only the user id `back\slash`.
- Over the real Netty server, a `q` of 10 000 characters gives 400 with the body
  `An HTTP line is larger than 4096 bytes.`

## Check 2: cost — PASS, with numbers

I seeded one app with 100 000 event rows. 90 000 rows hold `kind = 0`. They hold
4 500 distinct user ids and 20 000 session ids. I ran `ANALYZE`. Each number below is
the range of three runs on this machine, after one warm call:

| The call | The time | The result |
|---|---|---|
| no filter, page 1 | 55 ms to 65 ms | `pageCount = 90`, 50 rows |
| `q=%` | 27 ms to 33 ms | 0 rows |
| `q=a` (one character) | 71 ms to 78 ms | `pageCount = 90`, 50 rows |
| `q=user-0` | 91 ms to 98 ms | `pageCount = 90`, 50 rows |
| `page=2147483647` | 56 ms to 62 ms | clamped to page 90 |
| `q` of 10 000 characters | 31 ms to 36 ms | 0 rows |

The maximum is 98 ms. Every value stays far below one second.

No plan holds `SCAN event`. Each of the four statements holds
`SEARCH event USING COVERING INDEX event_agg (app_id=?)`. Thus there is no full table
scan.

The `COUNT(*)` subquery does run on each request, and it reads each index entry of the
app. D14 accepts that: "SQL at query time, with the indexes and statements of
section 6. No rollup table and no result cache in version 1." At 90 000 clicks the
cost is near 60 ms, thus one second arrives near 1.5 million clicks. The pilot app has
fewer than 500 visits each day (D42), thus the route has much spare capacity.

One note, and not a finding: `SqliteDatabase.read { }` holds one mutex for the whole
call. A page call of 98 ms blocks each other read for that time.

## Check 3: the app id — PASS

| The app id | The answer |
|---|---|
| `1e3` | 400 `{"error":"The app id must be a whole number."}` |
| `-1` | 404 `{"error":"The app is not registered."}` |
| `01` | 200, a valid page of app 1 |
| `9223372036854775808` | 400 |
| `1;` | 400 |
| the id of a second app | 200, with the rows of that app only |

Two apps never mix their users. The `appId` value is a bound `Long` in each statement.

## Check 4: leaks — PASS

I put a Logback appender on the root logger. It kept every level, and not only ERROR.
I then sent 20 requests: a normal call, a hostile `q`, a `q` with a NUL, a `q` with a
line break, a bad app id, a bad page, and one request that forced a 500.

- The run wrote 6 log events. Five are start-up lines of the test server and of
  `MonitorServices`. One is the ERROR line
  `An unhandled exception reached the server. SQLiteException`.
- No log line held a user id, the text of `q`, the query string, or the path. I
  searched the joined text for `user@example.com`, `example.com`, `DROP TABLE`,
  `script`, `q=`, and `/api/apps`. Each search gave false.
- Each error body is a fixed sentence. No body echoed the input.

**What a request log would hold.** `Application.kt` installs `ContentNegotiation`,
`StatusPages`, and the request guard. It installs no `CallLogging` plugin. Thus Ktor
writes no line for a request today, and no log holds `q`.

A later issue that installs `CallLogging` with the default format writes
`"${status}: ${httpMethod.value} - ${uri}"`. The `uri` value holds the whole query
string. The log then holds `q=<part of a personal id>`, for example
`q=user@example.com`. **The design does not accept that.** D15 says: "A log line never
holds a user id or a connection string."

The monitor binds to `127.0.0.1`, and it runs behind no proxy (D12), thus no proxy
access log exists today. A proxy in front would record the query string in its access
log, and D15 forbids that too. Issue #57 adds the file log with rotation; that issue
must keep the query string out of the file.

MAJOR 2 below turns this risk into a test failure, and not into a review finding.

## Check 5: the answer — PASS

I read the raw bytes from a real Netty server on a free port.

- `Cache-Control: no-store` is present on 200, on 400, on 404, and on 500.
- No answer holds `Access-Control-Allow-Origin`. The server installs no CORS plugin.
- A wrong Host header gives 403 on this route. Mutation E below proves that a test
  covers it.
- `Content-Type: application/json`, with no charset parameter. That is correct:
  RFC 8259 defines JSON as UTF-8, and it registers no charset parameter. The answer
  also holds `X-Content-Type-Options: nosniff` and the CSP of D12, thus a browser
  cannot read the body as HTML.
- A user id of `<script>alert(1)</script>` goes out as a JSON string:
  `{"userId":"<script>alert(1)</script>","clicks":1,...}`. The body holds no HTML
  element. A frontend issue must still bind the value as text, and not as HTML.

One note, and not a finding: a request line above 4096 bytes gets a Netty answer of
400 with `Content-Type: text/plain; charset=utf-8` and no `Cache-Control` header. The
decoder makes that answer before the request guard runs. The body names the limit
only; it echoes no input. This behaviour belongs to every route, and not to this one.

## Check 6: the tests — two gaps

I broke the code five times. I ran the test each time. I saw the result. I restored
the code each time.

|---|---|---|
| A | The `q` value goes into the SQL text, and not into a bound parameter | **No test failed.** |
| B | `ESCAPE '\'` removed from each filtered statement | `q filters the user id with a substring match, and % and _ are literal, not wildcards` FAILED |
| C | A log line with the `q` value added to the route | **No test failed**, in the whole module |
| D | `Cache-Control: no-store` skipped for a path that ends with `/users` | `GET answers with page, pageCount, rows, and Cache-Control no-store, and no CORS header` FAILED |
| E | The Host check skipped for a path that ends with `/users` | `a wrong Host header gets 403 on this route too` FAILED |

## Check 7: the test run

`./gradlew :monitor:backend:test`, one run, JDK 21: `BUILD SUCCESSFUL`. The XML results
hold **186 tests**, 0 failures, 0 errors, and 0 skipped tests. `UserTotalsRouteTest`
holds 18 tests. The pull request text is correct on both numbers.

---

## Findings

### MAJOR 1 — no test defends the bound parameter

`monitor/backend/src/test/kotlin/octometer/monitor/users/UserTotalsRouteTest.kt`
(the whole class; the nearest test starts at line 140).

The acceptance criterion says: "`q` filters the user id with one `LIKE`, and the value
is a bound parameter." The code is correct. No test covers the second half.

Evidence: I changed `countUserGroups` (`UserTotalsRoute.kt:154`) and
`readUserTotalsPage` (`UserTotalsRoute.kt:169`). Each one put the pattern into the SQL
text with `sql.replace("LIKE ? ESCAPE '\'", "LIKE '$likePattern' ESCAPE '\'")`. The
escape function stayed in place. `./gradlew :monitor:backend:test --tests
"octometer.monitor.users.UserTotalsRouteTest"` then gave `BUILD SUCCESSFUL`. With that
same mutation, a `q` of `'` and a `q` of `'; DROP TABLE event; --` each threw
`SQLiteException`, and the route answered 500. The current `%` and `_` test does not
catch the mutation, because the escape function hides each wildcard first.

The fix: add one test to `UserTotalsRouteTest`.

- Seed one app with two users.
- Send each of `'`, `'; DROP TABLE event; --`, `;`, and `--` as `q` over HTTP.
- Assert 200 and the page shape for each value.
- Read `SELECT COUNT(*) FROM event` after the calls. Assert the same row count as
  before the calls.

### MAJOR 2 — the sentinel test of D15 cannot fail for this route

`monitor/backend/src/test/kotlin/octometer/monitor/users/UserTotalsRouteTest.kt:269-284`,
with `monitor/backend/src/test/kotlin/octometer/monitor/LogCapture.kt:21`.

Two defects join here:

1. `captureErrorLogEvents` keeps only `Level.ERROR` events (`LogCapture.kt:21`). An
   INFO line or a DEBUG line with the filter text never reaches an assertion.
2. Line 280 asserts `errorEvents.isEmpty()`. The loop of lines 281 to 283 then reads an
   empty list, thus `assertFalse(event.formattedMessage.contains("alice-secret-id"))`
   never runs. That assertion is unreachable.

Evidence: I added
`log.info("The user page call holds q={} for app {}.", filter, appId)` after
`UserTotalsRoute.kt:77`. `./gradlew :monitor:backend:test --tests
"octometer.monitor.*"` gave `BUILD SUCCESSFUL`. The whole module passes while the route
writes the raw `q` value and the app id to the log.

The pull request text claims this criterion with the words "and no filter text". That
half of the claim does not hold. D15 asks for a real guard: "A sentinel test searches
the captured log."

The fix:

- Add `captureLogEvents` beside `captureErrorLogEvents` in `LogCapture.kt`. It returns
  each event of each level.
- In the test, join the `formattedMessage` value of each captured event.
- Assert that the joined text holds neither the `q` value nor the user id of the seeded
  row.
- Delete the unreachable loop of lines 281 to 283.

### MINOR 1 — `q` has no length limit

`monitor/backend/src/main/kotlin/octometer/monitor/users/UserTotalsRoute.kt:130`

`escapeLikePattern` can double the length of `q`. A `q` of 25 000 `%` characters makes a
pattern of 50 002 characters. The default `SQLITE_MAX_LIKE_PATTERN_LENGTH` is 50 000.
SQLite then throws `SQLiteException` ("LIKE or GLOB pattern too complex"), and the route
answers 500. I saw that 500 through the Ktor test engine.

The real server does not reach this state today. Netty answers 400 first, because the
request line passes 4096 bytes. The fix protects the route if a later change raises the
Netty limit, or if a later issue moves the filter into a request body.

The fix: reject a long value in the route handler, after line 77.

```kotlin
private const val MAX_FILTER_LENGTH = 200
// ...
if (filter != null && filter.length > MAX_FILTER_LENGTH) {
    call.respond(HttpStatusCode.BadRequest, ErrorBody("The filter q is too long."))
    return@get
}
```

### MINOR 2 — a NUL character in `q` drops the rest of the filter

`monitor/backend/src/main/kotlin/octometer/monitor/users/UserTotalsRoute.kt:130`

`GET /api/apps/1/users?q=%00` gave 200 with every user of the app. SQLite ends a text
value at the first NUL, thus the pattern becomes `%`, and it matches each row. A filter
that silently becomes "no filter" surprises the owner. This is not a leak, because the
owner already sees each row without a filter.

The fix: reject a `q` that holds a character below ` `. Answer 400 with the
sentence "The filter q holds a control character."

### MINOR 3 — the test folder skips the registered helper

`monitor/backend/src/test/kotlin/octometer/monitor/users/UserTotalsRouteTest.kt:35`

The field calls `Files.createTempDirectory("octometer-user-totals-test-")` directly.
`tearDown` deletes the folder, but a throw in `setUp`, or a killed JVM, leaves the
folder in `%TEMP%`. Issue #142 exists for that reason.

The fix: call `registerTempRoot(tempDir)` in the field initializer, or build the folder
with `testDataDir()`.

### MINOR 4 — `page=` with an empty value gives 400

`monitor/backend/src/main/kotlin/octometer/monitor/users/UserTotalsRoute.kt:89`

D13 writes the route as `GET /api/apps/{appId}/users?page=&q=`. An empty `q` works:
`toLikePattern` treats it as no filter. An empty `page` gives 400, because `parsePage`
reads `""` and `toIntOrNull` gives null. A form and a URL builder both send an empty
value often. The two parameters must behave the same way.

The fix: change line 90 to `if (raw.isNullOrBlank()) return 1`.

---

## The other checks of the reviewer rules

- The diff holds three files. It holds no secret, no database file, and no build output.
- The scope matches the issue. `ApiRoutes.kt` gained one import and one line, per the
  backend brief. The route sits in its own package, thus it touches the smallest count
  of shared lines.
- The code agrees with the design: the field names of D13, the page size of 50, the
  order `clicks DESC, user_id ASC`, and the literal `kind = 0` of rule M4 in the SQL
  text.
- The comments and the commit body use short sentences and the active voice.

## The verdict

No BLOCKER. Two MAJOR findings stay, thus I request changes. The route code passed every
security check that I ran. The gap is in the tests: a later change can remove the bound
parameter, or add a log line with a user id, and the suite stays green. Fix MAJOR 1 and
MAJOR 2, and I approve.

xamcross:
**Review (SQL and API engineer): CHANGES REQUESTED**

I made my own clone, I ran `gh pr checkout 145`, and I used JDK 21. I ran each
command below myself. I claim no check that I did not see.

## The acceptance criteria of issue #50

| # | Criterion | Result |
|---|---|---|
| 1 | The order is `clicks DESC, user_id ASC` | PASS in the code, FAIL in the tests (MAJOR 1) |
| 2 | The server clamps a page above the last page; the response holds `page` and `pageCount` | PASS |
| 3 | `q` filters the user id with one `LIKE`, and the value is a bound parameter | PASS |
| 4 | Each row holds `userId`, `clicks`, `sessions`, `uniqueElements`; the `NULL` user is one row | PASS |
| 5 | `EXPLAIN QUERY PLAN` shows the index `event_agg` | PASS |
| 6 | An unknown app id gives 404 | PASS |
| 7 | The captured log holds no query string | PASS |
| 8 | `./gradlew :monitor:backend:test` passes | PASS |
| 9 | A comment on the issue records the test result and the plan | PASS |

**Criterion 1.** `UserTotalsRoute.kt:40` and `:44` hold `ORDER BY clicks DESC,
user_id ASC`. I seeded 60 users with one click each, plus `alice` with 3 clicks
and one anonymous group with 4 clicks. The rows came back as `null(4)`,
`alice(3)`, then the one-click users in alphabetical order. The SQL is right.
No test proves the second sort key; see MAJOR 1.

**Criterion 2.** I seeded 69 user groups. Page 1 gave 50 rows and page 2 gave 19
rows. Over HTTP `?page=999` gave `{"page":1,...}` on a one-page app. The clamp
agrees with D13, which asks for "a page clamp with `page` and `pageCount`".
`?page=2147483647` gives no overflow, because the clamp runs before the code
computes the offset.

**Criterion 3.** The `LIKE` value travels through `setString`. Each statement
holds `ESCAPE '\'`. My probes over HTTP: `?q=%25` returned only the user
`50%_off`, and `?q=_` returned only `50%_off`. A wildcard never escapes.

**Criterion 4.** My lab row for the anonymous group: `userId=null (wasNull=true)
clicks=4 sessions=3 uniqueElements=2`. The four `kind = 0` rows of that group sat
in 3 sessions with 2 distinct elements. One `kind = 1` row of a fourth session
changed no count. The user `bob`, who holds only a `kind = 1` row, got no row at
all. Two apps did not mix their users. D13 asks for `userId: string | null`, and
the code gives exactly that. D44 names the label `(anonymous)`; that label
belongs to the view, not to the API.

**Criterion 5.** I ran `EXPLAIN QUERY PLAN` on all four statements myself, on
SQLite 3.53.4, with the exact schema of `db/002_first_page.sql`:

```
count, no filter | count, with q:
  CO-ROUTINE (subquery-1)
  SEARCH event USING COVERING INDEX event_agg (app_id=?)
  SCAN (subquery-1)

page, no filter | page, with q:
  SEARCH event USING COVERING INDEX event_agg (app_id=?)
  USE TEMP B-TREE FOR count(DISTINCT)
  USE TEMP B-TREE FOR count(DISTINCT)
  USE TEMP B-TREE FOR ORDER BY
```

Each line agrees with the pull request text. The literal `kind = 0` is correct.
I also measured the alternative: with `kind = ?` and no usable bound value the
plan falls to `SEARCH event USING COVERING INDEX event_session` plus `USE TEMP
B-TREE FOR GROUP BY`. The partial index is then lost.

**Criterion 6.** Over HTTP: `/api/apps/999/users` gave 404 with `{"error":"The
app is not registered."}`. A 20 digit app id gave 400. An app id `-1` gave 404.

**Criterion 7.** I sent 26 odd requests through the route: an empty page, a
negative page, a page of `1e2`, a page above the maximum of `Int`, a repeated
page, a repeated `q`, a `q` with a percent sign, a `q` with a NUL, and a `q` of
10 000 characters. Each answer was 200, 400, or 404. No answer was 500. The
count of captured ERROR log events was 0 for every case. The module installs no
`CallLogging` plugin, thus no request line reaches the log.

**Criterion 8.** `./gradlew :monitor:backend:test`, two times in sequence, JDK
21: `BUILD SUCCESSFUL` both times. 186 tests, 0 failures, 0 errors, 0 skipped.
18 of these tests sit in `UserTotalsRouteTest`. The numbers of the pull request
text are correct.

## My mutation checks

I broke the code 15 times. Each time I ran `--tests
octometer.monitor.users.UserTotalsRouteTest --rerun-tasks`, and I restored the
file after the run. 13 mutations gave a red test. Two gave no red test at all:

| Mutation | Red tests |
|---|---|
| The page statement loses `kind = 0` | 3 |
| `user_id ASC` becomes `user_id DESC` | **0** |
| `clicks DESC` becomes `clicks ASC` | 1 |
| `escapeLikePattern` becomes the identity function | 1 |
| The page clamp is removed | 1 |
| The page size becomes 25 | 3 |
| `pageCount` loses the floor of 1 | 2 |
| `parsePage` accepts 0 and a negative page | 1 |
| `sessions` loses `DISTINCT` | 1 |
| `uniqueElements` loses `DISTINCT` | 1 |
| The app-exists check always says yes | 2 |
| The count statement drops the `q` placeholder | 3 |
| The `ESCAPE` clause is removed | 1 |
| The filtered count ignores `q`, and it keeps the placeholder | **0** |
| `OFFSET` is always 0 | 2 |

## Findings

### MAJOR 1 — no test proves the sort key `user_id ASC`

`monitor/backend/src/test/kotlin/octometer/monitor/users/UserTotalsRouteTest.kt:126`

I changed `user_id ASC` to `user_id DESC` in both page statements
(`UserTotalsRoute.kt:40` and `:44`). The result was `BUILD SUCCESSFUL`, 18
tests, 0 failures. No test sees the change. The order test seeds `alice` with 2
clicks and `bob` with 1 click, thus `ORDER BY user_id ASC` alone gives the same
list. No two users share a click count anywhere in the file.

The tie-break carries weight here. Most users hold one click, thus the second
key alone decides the page boundary. Section 6 of the design asks that each
`ORDER BY` ends with a unique column, for exactly that reason.

**Fix.** Add a test with two users that share a click count. Seed `zeta` and
`alpha` with 2 clicks each, plus `top` with 3 clicks. Assert the exact list
`listOf("top", "alpha", "zeta")`.

### MAJOR 2 — no test proves that the count statement applies `q`

`monitor/backend/src/main/kotlin/octometer/monitor/users/UserTotalsRoute.kt:30`

I replaced the filter of `USER_TOTALS_COUNT_FILTERED_SQL` with
`AND (user_id LIKE ? ESCAPE '\' OR 1 = 1)`. The placeholder stays, thus the
binding still works, and the filter does nothing. The result was `BUILD
SUCCESSFUL`, 18 tests, 0 failures.

A user hits this defect class fast. With 60 users and a `q` that matches 3 of
them, the response then holds `pageCount: 2` and 3 rows. The view of #51 then
offers a second page that is always empty. The shipped code is right; only the
proof is absent.

**Fix.** Add a test that seeds more than 50 users, sends a `q` that matches a
small subset, and asserts both `pageCount == 1` and the exact row count.

### MAJOR 3 — the comment of `loadUserTotals` states a false rule, and a stale fact

`monitor/backend/src/main/kotlin/octometer/monitor/users/UserTotalsRoute.kt:104`

The comment says: "the mutex of `read { }` still keeps the two statements
together, with no poll-loop write between them on the one reader connection".
That rule is false on this branch. The mutex serialises two readers. It holds
back no writer.

My measurement, on the schema of this branch: `reader.getAutoCommit()` is
`true`. I ran the count statement on one connection and got 69 groups. I
committed one insert on a second connection. I ran the same count statement on
the same first connection and got 70 groups. The second statement reads a new
snapshot.

The second problem is a stale fact. The comment says that #127 is not on `main`
yet. `origin/main` is now `6154083`, "feat(monitor): serve the level 1 totals at
GET /api/apps (#127)". `git show origin/main:.../SqliteDatabase.kt` holds
`read { }` with `BEGIN` and `COMMIT`.

**Fix.** Merge `origin/main` into the branch, as the shared brief asks. Then
replace the last sentence of the paragraph with: "`main` holds that change since
commit 6154083, thus the two statements read one snapshot."

I checked that this merge is safe. I merged `origin/main` into a throwaway
branch, I kept both route lines of `ApiRoutes.kt`, and I ran the full suite: 197
tests, 0 failures, 0 errors.

### MINOR 1 — a loop that never runs

`monitor/backend/src/test/kotlin/octometer/monitor/users/UserTotalsRouteTest.kt:280`

`assertTrue(errorEvents.isEmpty())` stands directly before `for (event in
errorEvents) { assertFalse(...) }`. The loop body never runs. The `assertFalse`
line is dead code.

**Fix.** Delete the loop. The emptiness assertion already proves the rule.

### MINOR 2 — the temporary folder skips the project helper

`monitor/backend/src/test/kotlin/octometer/monitor/users/UserTotalsRouteTest.kt:35`

The field calls `Files.createTempDirectory` directly. The shared brief and issue
#142 ask for `registerTempRoot` or for `@TempDir`. My two full runs left 0
folders in `%TEMP%`, thus I saw no leak. The risk stays: if `setUp` throws,
`tearDown` reads the `lateinit` field `database` first. That read throws
`UninitializedPropertyAccessException`, and the code never reaches
`tempDir.deleteRecursively()`.

**Fix.** Call `registerTempRoot(tempDir)` in `setUp`. The shutdown hook then
deletes the folder on a pass and on a failure.

### MINOR 3 — an empty `page` gives 400, an empty `q` does not

`monitor/backend/src/main/kotlin/octometer/monitor/users/UserTotalsRoute.kt:89`

My probe over HTTP:

- `?page=` gives `400 {"error":"The page must be a positive whole number."}`
- `?q=` gives `200` with the full first page

D13 writes the route as `GET /api/apps/{appId}/users?page=&q=`. A view that
builds the query string from an empty form field gets a 400 for the page, and a
normal answer for the filter. The two rules disagree.

**Fix.** Write `if (raw == null || raw.isEmpty()) return 1` in `parsePage`.

### MINOR 4 — a NUL in `q` cuts the `LIKE` pattern

`monitor/backend/src/main/kotlin/octometer/monitor/users/UserTotalsRoute.kt:130`

The driver stops a bound text value at the first NUL. My proof, with a bound
pattern:

- `'abcali' LIKE '%ali<NUL>xyz%'` gives 1
- `'abcalixyz' LIKE '%ali<NUL>xyz%'` gives 0

The pattern becomes `%ali`, thus the trailing `%` is lost. Over HTTP,
`?q=%00` returned every user of the app, because the pattern became `%`.
`?q=ali%00ce` returned an empty list. The value stays bound, thus no injection
is possible. The answer is wrong, and it is silent.

**Fix.** Give 400 with a fixed sentence when `q` holds the character
`'\u0000'`. A different correct fix removes each NUL before the escape step.

### MINOR 5 — the case rule of `q` is not stated and not tested

`monitor/backend/src/main/kotlin/octometer/monitor/users/UserTotalsRoute.kt:133`

SQLite folds the case of `LIKE` for ASCII letters only. My measurements:

- `'ALICE' LIKE '%alice%'` gives 1; over HTTP `?q=ALICE` returned the user `alice`
- `'АЛИСА' LIKE '%алиса%'` gives 0
- `'Алиса' LIKE '%Алиса%'` gives 1

A user id with a Cyrillic letter thus needs the same case in `q`. The view of
#51 shows a filter box, and a user of that box will meet this rule.

**Fix.** Add one sentence to the comment of `escapeLikePattern`. Add one test
with an ASCII pair and one test with a Cyrillic pair.

### MINOR 6 — a non-empty `q` drops the anonymous row

`monitor/backend/src/main/kotlin/octometer/monitor/users/UserTotalsRoute.kt:131`

`NULL LIKE '%a%'` gives `NULL`, thus the row with `user_id IS NULL` never
matches a filter. My lab: 69 groups with no filter, and 68 groups with the
pattern `%`. The rule is defensible, because a `NULL` holds no text. D44 gives
the row `(anonymous)` its own place in level 2, thus a reader needs the rule in
writing.

**Fix.** State the rule in the comment of `toLikePattern`. Add one test that
sends a `q` and asserts that the row with `userId: null` is absent.

### MINOR 7 — the comment cites a rule that the design does not hold

`monitor/backend/src/main/kotlin/octometer/monitor/users/UserTotalsRoute.kt:22`

The comment says "Rule M4 of the design (section 6)". `grep -rn "M4" docs/
contract/` finds no match. A reader cannot find that rule.

**Fix.** Cite the sentence of section 6: "Level 2 and level 3 keep the filter
`kind = 0`, else SQLite ignores the index `event_agg`."

### MINOR 8 — the comment says "blank", the code tests "empty"

`monitor/backend/src/main/kotlin/octometer/monitor/users/UserTotalsRoute.kt:127`

The comment says "A blank filter means 'no filter'". The code calls
`isNotEmpty()`. A `q` of one space is thus a real filter.

**Fix.** Write "An empty filter means 'no filter'", or call `isNotBlank()`.

### MINOR 9 — a new compiler warning

`monitor/backend/src/test/kotlin/octometer/monitor/users/UserTotalsRouteTest.kt:140`

Each build prints: `Name contains character(s) that can cause problems on
Windows: %`. The Windows CI job builds this module.

**Fix.** Rename the test, for example: `q treats the percent sign and the
underscore as literal characters`.

## Facts, not findings

- **The merges.** `origin/main`, the branch of #127, and the branch of #144 each
  give one conflict, and each conflict sits in `ApiRoutes.kt`. The conflict is
  one line inside `apiRoutes`. Both sides keep their own line.
- **`main` moved.** `origin/main` is now `6154083` (#127). The branch sits on
  `3186569`.
- **A repeated parameter.** Ktor takes the first value. `?page=1&page=abc` gives
  200, and `?page=abc&page=1` gives 400. `?q=alice&q=bob` filters on `alice`.
  The rule is stable, and no criterion names it.
- **The scope.** The diff holds 3 files. It adds no secret, no database file, and
  no build output. `ApiRoutes.kt` gains one import and one line, as the brief
  asks.
- **The CI.** All five checks pass.
- **The two decisions of the author.** A substring match for `q` agrees with
  D13, which names a "filter `q` on the user id" and no shape. An empty result
  with `pageCount: 1` agrees with the clamp rule of D13, because `page` then
  names a real page. I accept both decisions.

Fix MAJOR 1, MAJOR 2, and MAJOR 3. The two MAJOR test gaps need one new test
each. MAJOR 3 needs the merge with `origin/main` and a new comment sentence. The
MINOR findings are yours to take or to leave.

xamcross:
## Correction round 1

I read both reviews. I fixed each MAJOR finding and each MINOR finding.
I merged `origin/main` (commit 6154083) first, with a merge commit, and
kept both lines of `ApiRoutes.kt`.

### SQL review, MAJOR findings

**MAJOR 1 — no test proves the sort key `user_id ASC`.**
Fixed. New test: `UserTotalsRouteTest.kt:145`, `two users with the same
click count break the tie on user_id ASC`. It seeds a tie in clicks and
asserts the exact order. I changed `user_id ASC` to `user_id DESC` in
`UserTotalsRoute.kt:44` and `:48`. The new test failed. I restored the
code, and the test passed again.

**MAJOR 2 — no test proves that the count statement applies `q`.**
Fixed. New test: `UserTotalsRouteTest.kt:289`, `the count statement
applies q, thus pageCount reflects the filtered row count`. It seeds 60
users (55 with a shared text, 3 of those with a rarer text) and asserts
`pageCount: 2` unfiltered, `pageCount: 2` with the wide filter, and
`pageCount: 1` with the narrow filter. I changed the filter of
`USER_TOTALS_COUNT_FILTERED_SQL` (`UserTotalsRoute.kt:34`) to
`AND (user_id LIKE ? ESCAPE '\' OR 1 = 1)`. The new test failed. I
restored the code, and the test passed again.

**MAJOR 3 — the comment states a false rule, and a stale fact.**
Fixed. `UserTotalsRoute.kt:130` to `:139` now states that `main` holds
the one-read-transaction change since commit 6154083, and names the
mechanism (`BEGIN` before the block, `COMMIT` after it). New test:
`UserTotalsRouteTest.kt:502`, `a committed insert between the count
statement and the page statement changes neither result`. It gives a
new user 5 clicks so a visible row would sort first and push a real
user off page 1, then commits that insert on the writer thread inside
the read block. I forced a mid-block `COMMIT` and a new `BEGIN` inside
the test (a stand-in for a lost read transaction). The new test failed.
I removed that stand-in, and the test passed again. A direct mutation
of `SqliteDatabase.read { }` (removing its own `BEGIN`/`COMMIT`) did not
fail this test; the reader connection keeps one open transaction from
its first-ever statement until an explicit `COMMIT`, independent of
that wrapper, so this particular mutation cannot show the difference on
this driver. The stand-in above proves the mechanism instead.

### SQL review, MINOR findings — all fixed

- MINOR 1 (dead loop): gone with the log test rewrite of security MAJOR 2.
- MINOR 2 (temp folder): `UserTotalsRouteTest.kt:35` now calls `registerTempRoot`.
- MINOR 3 (empty page vs empty q): `UserTotalsRoute.kt:120`, `parsePage` now
  treats an empty string the same as a missing value. New test at
  `UserTotalsRouteTest.kt:330`.
- MINOR 4 (NUL in q): `UserTotalsRoute.kt:103` rejects a control character
  with 400. New test at `UserTotalsRouteTest.kt:354`.
- MINOR 5 (case rule of q): stated in the route KDoc, `UserTotalsRoute.kt:73`.
  New test at `UserTotalsRouteTest.kt:213`.
- MINOR 6 (anonymous row dropped by q): stated in the route KDoc
  (`UserTotalsRoute.kt:78`) and in the code comment of `toLikePattern`
  (`UserTotalsRoute.kt:159`). New test at `UserTotalsRouteTest.kt:197`.
- MINOR 7 (the "M4" citation): `UserTotalsRoute.kt:25` now quotes the
  sentence of design document section 6 directly.
- MINOR 8 ("blank" vs "empty"): `UserTotalsRoute.kt:159` now says "empty".
- MINOR 9 (Windows warning on `%`): the test at `UserTotalsRouteTest.kt:171`
  has a new name with no `%` character. The build log holds no warning.

### Security review, MAJOR findings

**MAJOR 1 — no test defends the bound parameter.**
Fixed. New test: `UserTotalsRouteTest.kt:393`, `a hostile q value is a
bound parameter, thus the rows stay unchanged and the answer is 200`.
It sends a quote and a `DROP TABLE` payload as `q` over HTTP, and
asserts 200 with the row unchanged afterward. I changed
`countUserGroups` and `readUserTotalsPage` (`UserTotalsRoute.kt:191` and
`:206`) to concatenate the pattern into the SQL text instead of binding
it. The new test failed. I restored the code, and the test passed
again.

**MAJOR 2 — the sentinel test cannot fail.**
Fixed. `monitor/backend/src/test/kotlin/octometer/monitor/LogCapture.kt`
gained `captureLogEvents` (every log level), the same helper and body as
pull request #147. The test at `UserTotalsRouteTest.kt:457`,
`the captured log holds no line with the filter text, at every log
level`, now uses it, and the unreachable assertion loop is gone. I added
a temporary `log.info` call with the value of `q` in the route handler.
The test failed. I removed that call, and the test passed again.

### Security review, MINOR findings — all fixed

- MINOR 1 (no length limit on q): `UserTotalsRoute.kt:99` rejects a `q`
  above 200 characters with 400. New test at `UserTotalsRouteTest.kt:374`.
  I disabled the check for a moment; the test failed, then passed again
  after I restored the check.
- MINOR 2 (NUL drops the rest of the filter): the same fix and the same
  test as SQL review MINOR 4 above.
- MINOR 3 (temp folder): the same fix as SQL review MINOR 2 above.
- MINOR 4 (empty page vs empty q): the same fix and test as SQL review
  MINOR 3 above.

## Test result

`./gradlew :monitor:backend:test`, run twice with JDK 21: `BUILD
SUCCESSFUL` both times, 206 tests, 0 failures, 0 errors (27 of the 206
sit in `UserTotalsRouteTest`). `./gradlew build`, run once: `BUILD
SUCCESSFUL`.

## CI result

All 5 checks passed after the push: Frontend, JVM modules, Monitor
backend on Windows, Secret scan, Tracker. The first run of "JVM
modules" failed once, on an unrelated pre-existing test
(`ApplicationLifecycleTest`, a real-server thread-count test that this
pull request does not touch). That test passed on the Windows job of
the same run, and passed twice in my own local run of the whole
backend suite. A rerun of the same commit gave a green result on every
check, which points to a flake of that Linux job, not a defect of this
change.

xamcross:

I made a new clone. I ran `gh pr checkout 145` in it. I confirmed commit `b116c93` on
top of the merge commit `0fdccde`. I ran each check below myself with JDK 21.

## MAJOR 1 (the bound parameter) — FIXED

The new test is `UserTotalsRouteTest.kt:393`, `a hostile q value is a bound parameter,
thus the rows stay unchanged and the answer is 200`.

- I changed `countUserGroups` and `readUserTotalsPage` for a moment. Each one put the
  pattern into the SQL text with `base.replace("LIKE ? ", "LIKE '" + likePattern + "' ")`.
  The named test FAILED. I restored the code with `git checkout --`.
- I sent `q='` over HTTP: 200, `{"page":1,"pageCount":1,"rows":[]}`.
- I sent `q='; DROP TABLE event; --` over HTTP: 200, the same shape.
- `SELECT COUNT(*) FROM event` gave 3 rows before the calls and 3 rows after them.

## MAJOR 2 (the log sentinel) — FIXED

The new test is `UserTotalsRouteTest.kt:457`, `the captured log holds no line with the
filter text, at every log level`. It uses `captureLogEvents`, and the unreachable loop
is gone.

- I added `log.info("The user page call holds q={} for app {}.", filter, appId)` to the
  route for a moment. The named test FAILED. I restored the code.
- `captureLogEvents` in `LogCapture.kt` matches pull request #147. I ran
  `git fetch origin pull/147/head:pr147`, then `git show
  pr147:monitor/backend/src/test/kotlin/octometer/monitor/LogCapture.kt`, then `diff`.
  The two files hold the same name, the same KDoc, and the same body. Only the line
  ending differs (LF from `git show`, CRLF in the working tree).

## The four MINOR findings — each one FIXED

| # | The finding | The new rule | My HTTP check |
|---|---|---|---|
| MINOR 1 | `q` has no length limit | above 200 characters gives 400 | `q` of 201 characters: 400 `{"error":"The filter q is too long."}`. `q` of 200 characters: 200 with a valid page. |
| MINOR 2 | a NUL in `q` drops the filter | a control character gives 400 | `q=a%00b`: 400 `{"error":"The filter q holds a control character."}` |
| MINOR 3 | the test folder skips the helper | the field registers the root | `UserTotalsRouteTest.kt:37` now calls `registerTempRoot(it)` |
| MINOR 4 | `page=` gives 400 | an empty `page` means page 1 | `page=`: 200 with `"page":1` and every row. `q=`: 200 with no filter. |

`parsePage` (`UserTotalsRoute.kt:120`) now reads `if (raw.isNullOrEmpty()) return 1`.
The route rejects a long `q` at `UserTotalsRoute.kt:99` and a control character at
`UserTotalsRoute.kt:103`. Each check runs before the store, thus a bad value never
reaches SQLite.

## The test run

`./gradlew :monitor:backend:test`, one run, JDK 21: `BUILD SUCCESSFUL`. The XML results
hold **206 tests**, 0 failures, 0 errors, and 0 skipped tests. `UserTotalsRouteTest`
holds 27 tests. The build log holds no Windows name warning.

## The verdict

No BLOCKER and no MAJOR stays. Each finding of my first review is closed, and I saw
each fix fail its own test before I restored the code. I approve.

xamcross:
**Review (second review, SQL and API engineer): APPROVE**

I made a new clone, I ran `gh pr checkout 145`, and I used JDK 21. HEAD is
`b116c93`, on top of the merge commit `0fdccde`. I ran each check below myself.

## The three MAJOR findings

**MAJOR 1 — the sort key `user_id ASC`. FIXED.**
I changed `ORDER BY clicks DESC, user_id ASC` to `user_id DESC` in both page
statements. Two tests went red:

- `two users with the same click count break the tie on user_id ASC`
- `a committed insert between the count statement and the page statement changes neither result`

I restored the file. The suite went green again.

**MAJOR 2 — `pageCount` under a filter. FIXED.**
I replaced the filter of `USER_TOTALS_COUNT_FILTERED_SQL` with
`AND (user_id LIKE ? ESCAPE '\' OR 1 = 1)`. The placeholder stays, thus the
binding still works. One test went red: `the count statement applies q, thus
pageCount reflects the filtered row count`. I restored the file.

**MAJOR 3 — the comment on the snapshot. FIXED, and the test is stronger than
you report.**

The comment at `UserTotalsRoute.kt:125` is correct now. It names commit
`6154083`, and it names the mechanism.

I must correct your claim about the driver. You write that "the reader
connection keeps one open transaction from its first-ever statement until an
explicit `COMMIT`, independent of that wrapper". That is false. I measured it.
I opened one read-only connection and one writer connection with the pragmas of
`SqlitePragmas`, on the schema of this branch, with sqlite-jdbc 3.53.4.
`getAutoCommit()` is `true`. I ran one count, committed one insert on the
writer, and ran the same count again:

| Case | Groups before | Groups after | The reader saw the write |
|---|---|---|---|
| No wrapper, a fresh reader | 5 | 6 | yes |
| No wrapper, after earlier statements | 6 | 7 | yes |
| With `BEGIN` ... `COMMIT` | 7 | 7 | **no** |
| No wrapper, after the wrapped block committed | 8 | 9 | yes |

Without the wrapper the reader starts a new snapshot for each statement. The
wrapper is the one thing that holds the snapshot.

Your mutation missed that, because the mutation itself was not clean. I removed
the `BEGIN` line alone: 27 tests went red, because `COMMIT` then throws. I
removed the `COMMIT` line alone: 10 tests went red, because the next `BEGIN`
throws. Both numbers hide the signal.

I then removed **both** lines together, so `read { }` became
`readMutex.withLock { block(reader) }`. Exactly one test went red:

```
C5 read { } loses BOTH BEGIN and COMMIT | exit=1 | failing=1
      FAIL: a committed insert between the count statement and the page statement changes neither result
```

I restored the file. Your test catches the direct mutation cleanly.

**What the test proves.** The count statement and the page statement of one
`read { }` block read one snapshot. The `BEGIN` and the `COMMIT` of
`SqliteDatabase.read { }` give that snapshot. A write that commits between the
two statements changes neither result.

**What the test does not prove.** It drives the write from inside the block, on
the writer thread of the same process. It proves nothing about a second process
or about a real poll loop. It calls `countUserGroups` and `readUserTotalsPage`
directly, thus it covers neither the app-exists check nor the `page` and
`pageCount` arithmetic under a concurrent write. It uses no filter, thus the
filtered path stays untested under a concurrent write. These gaps are small,
and no acceptance criterion names them.

## The nine MINOR findings

|---|---|---|---|
| 1 | A loop that never runs | FIXED | `grep -n "for (event in"` on the test file finds nothing |
| 2 | The temporary folder skips the helper | FIXED | `UserTotalsRouteTest.kt:38` calls `registerTempRoot(it)` |
| 3 | An empty `page` gives 400 | FIXED | `UserTotalsRoute.kt:120` holds `if (raw.isNullOrEmpty()) return 1` |
| 4 | A NUL in `q` cuts the pattern | FIXED | `UserTotalsRoute.kt:103` gives 400 for `it.code < 0x20` |
| 5 | The case rule of `q` | FIXED | The route KDoc, `UserTotalsRoute.kt:71`, states the ASCII rule |
| 6 | A non-empty `q` drops the anonymous row | FIXED | The route KDoc, `:78`, and the comment of `toLikePattern`, `:159` |
| 7 | The citation of a rule "M4" | FIXED | `grep -n "M4"` finds nothing; `:25` quotes section 6 directly |
| 8 | "blank" against `isNotEmpty()` | FIXED | `:157` now says "An empty filter means 'no filter'" |
| 9 | The Windows warning on `%` | FIXED | The build log of my full run holds no `w:` line |

## The test run

`./gradlew :monitor:backend:test`, JDK 21, one time, on a clean tree:
`BUILD SUCCESSFUL`. 206 tests, 0 failures, 0 errors, 0 skipped. 27 of these
tests sit in `UserTotalsRouteTest`. Your count of 206 is correct.

## The verdict

No BLOCKER and no MAJOR stays. I approve the change. Please correct the record
on the driver claim in a later comment or in a commit body, so that a reader of
this thread does not carry the wrong rule to a different route.

xamcross:
## Merge with main (commit 7576e7a)

I merged `origin/main` into this branch with merge commit `6c202c0`.
The conflict in `ApiRoutes.kt` kept each line: `appTotals`,
`userTotalsRoute`, and the new `elementsRoute` of pull request #144.
`LogCapture.kt` needed no manual resolution; `main` already held the
same `captureLogEvents` body as this branch.

`./gradlew :monitor:backend:test`, run twice with JDK 21: `BUILD
SUCCESSFUL` both times, 233 tests, 0 failures, 0 errors. All 5 CI
checks pass after the push.

PR #146, opened 2026-09-22
feat(backup): back up the database each day and before a migration

### LINKED ISSUE #55: Back up the database each day and before a migration
**Goal.** The monitor writes one backup of `octometer.db` each day and one backup before it applies a migration.

**Implementation steps.**
1. Add a backup function to `monitor/backend`. It runs `VACUUM INTO` with the target `octometer-<yyyyMMdd>.db` in the folder `backups`, a sibling of `dataDir` (D3).
2. Run the backup one time each day. Inject a `Clock` for the tests.
3. Keep only the 7 newest daily files.
4. Make the migration runner write `pre-migrate-v<n>-<timestamp>.db` before it applies a migration. `<n>` is the schema version of the file before the migration.
5. Write the tests of the acceptance criteria in `monitor/backend`.

**Acceptance criteria.**
- [x] The monitor writes `backups/octometer-<yyyyMMdd>.db` with `VACUUM INTO` one time each day and keeps 7 files (fixed clock).
- [x] The migration runner first writes `pre-migrate-v<n>-<timestamp>.db`, then it applies the migration.
- [x] A test restores a backup file and reads the same event count as the source.
- [x] A backup file holds no connection string.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, D3, D11, D34, and D36 (first part).

xamcross:
The monitor now backs up `octometer.db` each day and before a migration. The backup method is `VACUUM INTO`. This method is safe for a database in WAL mode. A write goes to a temporary name first. The code then moves the file into place. The final name never holds a half-written file.

A daily job in `MonitorServices` writes `backups/octometer-<yyyyMMdd>.db`. The job keeps the 7 newest files. `MigrationRunner` writes `backups/pre-migrate-v<n>-<timestamp>.db` before it applies a pending migration. A failed backup stops the migration. A failed backup stops the start too.

## Acceptance criteria

- [x] The monitor writes `backups/octometer-<yyyyMMdd>.db` with `VACUUM INTO` one time each day and keeps 7 files (fixed clock). Evidence: `DailyBackupJobTest`, and `BackupPathsTest` (`pruneDailyBackups with RETAINED_DAILY_BACKUPS keeps the newest 7 files, and a foreign file stays`), which reads the named constant instead of its own value.
- [x] The migration runner first writes `pre-migrate-v<n>-<timestamp>.db`, then it applies the migration. Evidence: `MigrationRunnerBackupTest`.
- [x] A restore test runs the full restore against a real folder, and reads the row count of the backup. Evidence: `RestoreTest` (`a restore that follows every step of the README gives back the row count of the backup`, and the guard case).
- [x] A backup file holds no connection string. Evidence: `BackupSecretSafetyTest`, which now creates the app through `AppRegistryService.createApp`, the real request path.
- [x] `./gradlew :monitor:backend:test` passes. Evidence: run two times in sequence on JDK 21, 292 tests, 0 failures, 0 errors, both times, after this correction round.
- [x] `./gradlew build` passes one time. Evidence: `BUILD SUCCESSFUL`.

## Correction round 1 (this round)

Two reviews of this pull request found one BLOCKER and six MAJOR findings. Each one is fixed, with a test.

- [x] BLOCKER 1 (reliability review): a restore with the two side files kept gave the old data back, with no warning. `SqliteDatabase.open` now refuses that start with one log line (`monitor/backend/src/main/kotlin/octometer/monitor/store/SqliteDatabase.kt:63` and `:195`). `monitor/backend/README.md` states the restore steps. `RestoreTest` proves both the restore and the guard.
- [x] MAJOR 1 (reliability): a failed daily backup marked the day done, thus the monitor skipped the rest of that day. `DailyBackupJob` now sets `lastBackupDate` only after a success (`DailyBackupJob.kt:127`), and retries at the next tick, with a warning throttled to one line for each hour (`DailyBackupJob.kt:144`).
- [x] MAJOR 2 (reliability): the secret safety test could not fail, because it wrote the connection string to the secret store file directly. It now creates the app through `AppRegistryService.createApp` (`BackupSecretSafetyTest.kt:49`), the real request path, and reads every byte of the backup file and of the live database file.
- [x] SQLite MAJOR 1: five test call sites passed a temporary folder directly as `dataDir`, thus their backups leaked into the shared `%TEMP%\backups` folder. `MonitorConfig` gets a new key, `backupDir` (`MonitorConfig.kt:18`), that defaults to the sibling folder `backups` of `dataDir`. The five call sites now nest `dataDir` under a registered root, the same pattern as `MigrationRunnerTest`.
- [x] SQLite MAJOR 2: the evidence test for "keeps 7 files" called `pruneDailyBackups` with its own `retainedCount`, thus it never read `RETAINED_DAILY_BACKUPS`. A new test seeds 9 files and reads the named constant (`BackupPathsTest.kt:132`).
- [x] SQLite MAJOR 3: the coroutine debug mode of the test JVM renames the writer thread to `octometer-sqlite-writer @coroutine#N` while a coroutine runs on it, for example a backup. The exact name match read 0 during a slow backup. Both helpers now use `startsWith` (`ApplicationLifecycleTest.kt:195`, `MonitorServicesTest.kt:100`).
- [x] SQLite MAJOR 4: a crash between the write and the move left a `.db.tmp` file for ever, because the prune matched only the final names. `sweepStaleTempFiles` now removes each stale temporary file, at the start of the job and at each daily prune (`BackupPaths.kt:91`, `DailyBackupJob.kt:76` and `:116`).

### MINOR findings

The two reviews list 20 MINOR findings, several of which name the same underlying line from each reviewer's own angle (the `CancellationException` guard, the `registerTempRoot` gap, and `lastBackupDate` after a failure, which MAJOR 1 above already closes). Counted as distinct fixes, 14 are corrected in this round:

`isDailyBackupDue` now uses `isAfter`, not `!=`, so a backward clock change never marks an earlier day due again; the job never writes over an existing backup file, with one log line; a removed backup now logs a count; the pre-migration timestamp holds milliseconds, so two migrations inside one second give two files; the `CancellationException` catch of `DailyBackupJob` checks `isActive` before it throws again; the success log line holds the file name and the size, proven by a sentinel test with `captureLogEvents`; the UTF-8 BOM of `Migration002Test.kt` is gone; `SqliteDatabase.close()` closes the dispatcher and waits for the writer thread before it closes the two connections; `awaitWriterShutdown` writes one WARN line on a timeout; a failed backup exception drops its cause, so a stack trace never carries a path; `writeTo` now runs `PRAGMA quick_check` on the fresh temporary file; the `BackupPathsTest` prune test now uses `retainedCount = 0`, so a widened pattern match cannot pass by accident; and each of the seven new backup and migration test classes named across both reviews now calls `registerTempRoot`.

Two MINOR findings are not corrected, both with the reviewer's own note that a separate issue should carry them:
- The Windows-only lock test of `DatabaseBackupTest` has no portable form on Linux; the SQLite reviewer called the current skip acceptable and asked for the portable form in a separate issue.
- A `pre-migrate-*.db` file is never pruned; the count is small (one file for each migration ever applied), and `monitor/backend/README.md` now states that rule instead of a code change.

The backup runs on the writer thread (192 ms for 200 000 rows, one time each day); this choice stays as it is, with the tradeoff stated in a code comment at `DailyBackupJob.kt:112`.

### Proof of the SQLite MAJOR 3 diagnosis

A scratch test (not part of the shipped suite) opened a real `SqliteDatabase`, launched a coroutine on `Dispatchers.Default` that ran a 2.5-second `Thread.sleep` inside `database.write { }`, then read the writer thread's name. Result: the thread held the name `octometer-sqlite-writer @coroutine#3` while the coroutine ran; the exact match gave 0, and `startsWith` gave 1. This matches the CI failure of round 0.

## Extra tests, beyond the named acceptance criteria

- `DailyBackupJobTest`: `stop` waits for a backup that is in progress and leaves no partial file under the final name. `stop` cancels at once when no backup runs. A failed backup retries at the next tick. A backup that fails every tick writes at most one warning for one hour. `runOnce` never writes over an existing backup file. `start` sweeps each stale temporary file of an earlier crash. The success log line holds the file name and the size, and no folder path.
- `DatabaseBackupTest`: a second backup replaces a stale final file. No temporary file stays when the final name is locked (Windows semantics, guarded to run on Windows only, the same rule as `SecretStoreTest`). The write fails with no final file when the backups folder is not a folder.
- `MigrationRunnerBackupTest`: a new database with no pending migration writes no pre-migration backup. A failed backup leaves the database file at its old schema version.
- `BackupPathsTest`: the file name formats, the due rule of the daily clock (including a backward clock change), the prune rule that never deletes a file the monitor did not make, the millisecond resolution of the pre-migration timestamp, and the stale-temp-file sweep.
- `RestoreTest`: the full restore against a real folder, and the guard against a stale write-ahead log.
- `ConfigLoaderTest`: the bundled default of `backupDir`, and the `OCTOMETER_BACKUP_DIR` environment variable.

## Design source

`docs/superpowers/specs/2026-09-21-octometer-design.md`, D3, D11, D34, D36 (first part).

## Confirm

This issue needed no version confirmation. `sqlite-jdbc` and `kotlinx-coroutines-core` were already pinned in `gradle/libs.versions.toml`. This work adds no new dependency.

## Decisions made where the issue was silent

- The backup method is `VACUUM INTO` with a temporary file name, then an atomic move. This matches the rule for a database in WAL mode.
- The daily job runs its first backup at once, when the monitor starts. It runs again once the calendar day of the clock changes, on a 1-minute tick in production. A test gives its own short tick.
- `close()` waits for a backup that is in progress (`cancelAndJoin`), instead of a hard interrupt. The final file is then always complete or absent, never partial.
- Three existing tests (`SqliteDatabaseTest`, `MigrationRunnerTest`, `Migration002Test`) moved their SQLite data folder under a nested `data` subfolder of their own temp folder. The new sibling `backups` folder of D34 then stays inside the test's own temp folder, and it deletes with that folder.

## Fixes made after the first two CI runs

- The Linux job (`JVM modules`) failed one locked-file test, because a Linux advisory lock does not block a file move the way a Windows byte-range lock does. That test now runs on Windows only, guarded with `Assumptions.assumeTrue`, the same rule as `SecretStoreTest`.
- The Windows job (`Monitor backend on Windows`) failed `ApplicationLifecycleTest` twice, at the same writer-thread-count check, right after a 200 from the health route. `SqliteDatabase.close()` now waits for the writer executor to terminate, not only for its shutdown request. `ApplicationLifecycleTest` now polls for the writer thread count for up to two seconds, the same rule as `MonitorServicesTest.awaitWriterThreadCount` already uses. All five checks pass after these two fixes.

## Shared files changed in this correction round

- `monitor/backend/src/main/kotlin/octometer/monitor/MonitorServices.kt`: passes `config.backupDir` to `SqliteDatabase.open` and to `DailyBackupJob`.
- `monitor/backend/src/main/kotlin/octometer/monitor/config/MonitorConfig.kt` and `ConfigLoader.kt`: the new config key `backupDir`.
- `monitor/backend/src/test/kotlin/octometer/monitor/TestConfigs.kt`: `devConfig` and `prodConfig` gain a `backupDir` parameter.
- `monitor/backend/src/test/kotlin/octometer/monitor/ApplicationLifecycleTest.kt` and `MonitorServicesTest.kt`: the writer thread name match, and (`ApplicationLifecycleTest`) the nested `data` folder.

Closes #55

Generated with Claude Code


xamcross:
**Review (reliability engineer): CHANGES REQUESTED**

I made my own clone, ran `gh pr checkout 146`, and tested the branch `issue-55-database-backup` at commit `15386ae` on JDK 21, on Windows 11. Each number below comes from a command that I ran. I restored each file after each experiment, and my clone is clean at `15386ae`.

## The acceptance criteria

1. `backups/octometer-<yyyyMMdd>.db` with `VACUUM INTO`, one time each day, 7 files kept, fixed clock — **PASS**. `RETAINED_DAILY_BACKUPS = 7` at `DailyBackupJob.kt:24`. I changed the prune to `drop(retainedCount + 10)`; `BackupPathsTest` and `DailyBackupJobTest` both failed.
2. The migration runner writes `pre-migrate-v<n>-<timestamp>.db` before the migration — **PASS**. I removed the call at `MigrationRunner.kt:43`; two tests of `MigrationRunnerBackupTest` failed.
3. A test restores a backup file and reads the same event count as the source — **FAIL**. See BLOCKER 1. No test restores anything.
4. A backup file holds no connection string — **PASS in fact, but the test cannot fail**. See MAJOR 2.
5. `./gradlew :monitor:backend:test` passes — **PASS**. 197 tests, 0 failures, 0 errors, 0 skipped, two runs in sequence. A third run after I restored each experiment gave the same count. The five CI checks pass.
6. A comment on the issue records the test result — **PASS**. I read it.

## BLOCKER 1 — a real restore gives the old data back, and no text and no test hold the restore steps

`monitor/backend/src/test/kotlin/octometer/monitor/backup/DatabaseBackupTest.kt:28`, and the absent `monitor/backend/README.md`.

The named test opens the backup file where it lies. It uses a raw JDBC connection, a hand-made one-column `event` table, and no WAL mode. It never copies the file over `octometer.db`, it never removes a side file, and it never opens the result with `SqliteDatabase`. The criterion asks for a restore. This is not one.

I wrote the restore as a test. The evidence:

- The live `octometer.db` holds the header bytes 18 and 19 as `2 2`, which means WAL. The backup file holds `1 1`, which means a rollback journal. `VACUUM INTO` always writes a rollback-journal database.
- I opened a real `SqliteDatabase`, wrote 1000 event rows, ran `DailyBackupJob.runOnce()`, then wrote 5000 more rows. I copied `octometer.db`, `octometer.db-wal`, and `octometer.db-shm` aside. That copy is a crash snapshot of the monitor.
- Restore A: copy the backup over `octometer.db`, keep the two side files, start the monitor (`SqliteDatabase.open`). Result: **count = 6000**, `PRAGMA quick_check = ok`.
- Restore B: copy the backup over `octometer.db`, delete `octometer.db-wal` and `octometer.db-shm`, start the monitor. Result: **count = 1000**, `PRAGMA quick_check = ok`.

Restore A gave the owner his broken data back. Nothing warned him. The cause is `SqlitePragmas.applyTo`, which runs `PRAGMA journal_mode=WAL` on each connection. That statement turns the restored rollback-journal file back into a WAL file, and it replays the stale write-ahead log over it.

The exact fix, in three parts. Part (a) and part (b) are required:

(a) Add `monitor/backend/README.md` with the restore as a numbered list:
1. Stop the monitor. End the scheduled task `Octometer`. Wait for the process to end.
2. Copy `backups\octometer-<yyyyMMdd>.db` over `data\octometer.db`.
3. Delete `data\octometer.db-wal` and `data\octometer.db-shm`. **Step 3 is not optional.**
4. Start the monitor.
5. Read `GET /api/health`, then compare the event count with the count of the backup.

(b) Add `RestoreTest` in `monitor/backend/src/test/kotlin/octometer/monitor/backup/`. It runs the five steps against a real `SqliteDatabase`: seed the rows, back up, write more rows, close, copy the backup over `octometer.db`, delete the two side files, open again, assert the same event count as the source. Add a second case that keeps the two side files. That case must fail today, thus it proves the rule.

(c) Recommended guard in `SqliteDatabase.open`: when byte 19 of `octometer.db` is `1` and `octometer.db-wal` exists, refuse the start with exit code 2 and one fixed sentence, for example "A stale write-ahead log file lies beside a restored database. Delete octometer.db-wal and octometer.db-shm, then start again." A write-ahead log file can never belong to a rollback-journal database, thus this rule is safe.

## MAJOR 1 — a failed backup marks the calendar day as done, thus the monitor writes no backup for a day

`monitor/backend/src/main/kotlin/octometer/monitor/backup/DailyBackupJob.kt:90`.

The line `lastBackupDate = LocalDate.now(clock)` runs on each path. The `catch` at line 87 does not skip it.

My evidence. I put a plain file at the `backups` path, started the job with a tick of 20 ms and a fixed clock, and saw one line `ERROR o.m.b.DailyBackupJob - The daily backup failed. BackupFailedException`. I then repaired the folder inside the same calendar day and waited 400 ms. Result: `files after the repair, same calendar day = []`. No backup came. I repeated the same test with a change of the calendar day. Result: `[octometer-20260109.db]`. The job stays alive, and it skips the rest of the day.

One virus scan that holds a handle, one short full disk, or one busy folder at the tick gives no backup for that whole day. The owner sees one error line.

The fix: set the date only after a success, and add a bounded retry.

```kotlin
if (isDailyBackupDue(clock, lastBackupDate)) {
    val today = LocalDate.now(clock)
    try {
        runOnce()
        lastBackupDate = today
        failedAttempts = 0
        log.error("The daily backup failed. {}", cancellation.javaClass.simpleName)
    } catch (failure: Exception) {
        log.error("The daily backup failed. {}", failure.javaClass.simpleName)
        failedAttempts += 1
        if (failedAttempts >= MAX_DAILY_ATTEMPTS) {
            lastBackupDate = today
            failedAttempts = 0
        }
    }
}
```

Use `MAX_DAILY_ATTEMPTS = 5`. A permanent failure then writes five error lines each day, not one line each minute. Add a test: a failed run, a repair inside the same calendar day, then a backup file appears.

## MAJOR 2 — `BackupSecretSafetyTest` cannot fail

`monitor/backend/src/test/kotlin/octometer/monitor/backup/BackupSecretSafetyTest.kt:32`.

The test writes the connection string to the secret store only. D3 puts the folder `secrets` outside `dataDir`, thus the string never comes near `octometer.db`. The assertion at line 48 holds for every build.

My evidence. I changed `AppRegistryService.insertApp` to write the connection string into the column `app.last_error`, which is a real leak into the store. `BackupSecretSafetyTest` passed. The full suite caught the leak in `AppRegistryRoutesTest > a byte search of the store files finds no connection string after a create`. The rule holds, and the new test adds nothing.

The byte search itself is correct: `readText(Charsets.ISO_8859_1)` maps each byte to one character, thus `contains` is a byte search for an ASCII value.

The fix: create the app through the real path, then back up.

```kotlin
val service = AppRegistryService(database, secretStore)
service.createApp(CreateAppRequest("probe", secretString, "exampledb", "octometer_events"))
DailyBackupJob(database, dataDir, clock).runOnce()
```

Then keep the byte search of line 47 and line 48. I ran this variant: `createApp` returns `Created`, the backup file holds no connection string, and `octometer.db` holds none. The test then turns red when a later change puts a secret into the store.

## MINOR

1. `BackupPaths.kt:41` — `isDailyBackupDue` compares with `!=`. A backward clock change over midnight makes the job write the file name of the day before with the data of today, thus it replaces a good older backup. My check: with `lastBackupDate = 2026-01-09` and a clock at `2026-01-08T23:55`, the call returns `true`. Fix: `lastBackupDate == null || LocalDate.now(clock).isAfter(lastBackupDate)`.
2. `DailyBackupJob.kt:75` — the success line holds the file name and no size. Fix: `log.info("The daily backup wrote {} ({} bytes).", file.name, file.length())`. The size tells the owner that the backup is not empty.
3. `BackupPaths.kt:56` — the prune deletes a backup file with no log line. A removed backup must leave a record. Fix: add `log.info("The daily prune removed {} file(s).", toDelete.size)`.
4. `DatabaseBackup.kt:83` — the `BackupFailedException` keeps the `IOException` as its cause, and that cause holds two absolute paths. I measured it: `AccessDeniedException: ...\backups\octometer-20260105.db.tmp -> ...\backups\octometer-20260105.db`. The class comment at line 14 says the message never holds a full path. That is true for the message and false for the chain. A failed pre-migration backup leaves `MonitorServices.open` with no catch, thus a handler that prints the stack trace writes both paths into the log file. Fix: drop the cause, or replace it with an exception whose message holds the file name only.
5. `DailyBackupJob.kt:85` — `catch (cancellation: CancellationException) { throw cancellation }` is the unguarded form. `Application.kt:89` already uses the guarded form. My evidence that the form matters: I removed the stop call from the catch of `MonitorServices.open`; the closed writer dispatcher then rejected the task, the loop died, and it wrote no log line. Fix: add `if (!currentCoroutineContext().isActive)` before the throw, as in MAJOR 1 above.
6. The new test classes make a temporary folder with `Files.createTempDirectory` and no `registerTempRoot`: `DailyBackupJobTest.kt:24`, `DatabaseBackupTest.kt:19`, `BackupSecretSafetyTest.kt:22`, `BackupPathsTest.kt:57`, `MigrationRunnerBackupTest.kt:22`. A killed test JVM or a locked file leaves the folder in `%TEMP%`. Issue #142 exists for that reason. Fix: add `.also { registerTempRoot(it) }`, the pattern of `ApplicationLifecycleTest.kt:38`.
7. `DatabaseBackupTest.kt:91` — the rule "the final name never holds a half-written file" rests on one test, and `Assumptions.assumeTrue` skips that test outside Windows. The Linux job thus proves nothing about the move. Fix: add a test that holds on each platform, for example "a successful write leaves no `.tmp` file in the folder".
8. `BackupPaths.kt:50` — the prune matches `octometer-<8 digits>.db` only, thus a `pre-migrate-*.db` file stays for the life of the install. The count is small, one file for each migration. State the rule in the README, or prune those files after a year.
9. `DatabaseBackup.writeTo` runs no `PRAGMA quick_check` on the fresh temporary file. A cheap check catches a bad write at the moment of the write, not at the restore, a year later. Fix: open the temporary file, run `PRAGMA quick_check`, and fail the run when the answer is not `ok`.
10. `BackupPaths.kt:47` — "never gets deleted" is the passive voice. Write "the prune keeps each other file of the folder".

## The answers to my role checks

**The lifecycle (PASS).** `open` starts the job at `MonitorServices.kt:78`. `close()` stops it at `MonitorServices.kt:36`. I proved the third case: I put a throw right after `it.start()` and gave the job a tick of 50 ms, then counted the loop turns. With the stop call at `MonitorServices.kt:81`: 1 turn, and 0 more turns in 600 ms. With the stop call removed, and with `database.close()` removed too so only the stop call differs: 8 more turns in 600 ms. The stop call is what ends the job. I restored each file.

**The close time (PASS, with a note).** With 200 000 event rows and an `octometer.db` of 20 004 864 bytes: `VACUUM INTO` takes 192 ms, and `job.stop()` with a backup in flight takes 131 ms. The wait has no timeout, thus the shutdown time grows with the database size, near 10 seconds for each gigabyte. At the size of the owner that is safe today. A later issue can add `withTimeoutOrNull`.

**The schedule.** The first backup runs at the start, because `lastBackupDate` is null. The tick is 1 minute. The job uses the **local** calendar day, because `MonitorServices.open` defaults to `Clock.systemDefaultZone()` and `LocalDate.now(clock)` reads the zone of the clock. The design names no rule here; the local day is right for a personal machine, and the README must state it. A machine that sleeps over midnight and wakes at 09:00 writes **one backup at once**, inside one tick, thus a maximum of one minute after the wake. It does not wait for the next midnight, and it writes no extra file for a day that it slept through. That is the right behaviour. A backward clock change is MINOR 1.

**A failure inside a run (PASS).** I made the `backups` path a plain file. The run wrote one line, `ERROR o.m.b.DailyBackupJob - The daily backup failed. BackupFailedException`. The job stayed alive, and it wrote a backup after the calendar day changed. `GET /api/health` holds no database call (`Application.kt:96`), thus a slow or a failed backup never changes the health answer. MAJOR 1 stays: the job is alive, and it still skips the rest of the day.

**The second job of pull request #149.** The two jobs **cannot** run at one moment. `SqliteDatabase.write` is `withContext(writerDispatcher)` on a single-thread executor, thus a purge batch and a `VACUUM INTO` stand in one line. There is no race and no parallel writer. The cost is the wait in that line: a `VACUUM INTO` of 200 000 rows holds the writer thread for 192 ms, and a purge batch and the page transaction of the poller (D4) wait that long. At 2 million rows expect near 2 seconds. `GET /api/health` reads no database, thus it keeps its 200 answer. A read route uses the read-only connection on `Dispatchers.IO`, and WAL lets a reader run beside a writer, thus the ingest read is not blocked. Three points for the merge:
- The two pull requests change the same lines of `MonitorServices.open` and `close()`. #149 starts its job **before** the orphan sweep; #146 starts its job **after** it. The merged `open` must stop both jobs in the catch, and the merged `close()` needs one `runBlocking` that stops both jobs before `database.close()`.
- Start the daily backup **before** the purge job. Both jobs run at the start today. When the purge runs first, the backup of the day holds the state after the purge, thus a bad purge has no same-day backup to restore from.
- #149 runs `PRAGMA wal_checkpoint(TRUNCATE)`. A read transaction of #127 blocks that checkpoint. That point belongs to the review of #149.

**The log lines.** The success line holds the file name and no other path data, and no size (MINOR 2). No line reports a removal (MINOR 3). The failure line holds one fixed sentence and the class name, at `DailyBackupJob.kt:88` and at `MigrationRunner.kt:62`. No line holds a path. MINOR 4 covers the exception chain.

**The tests.** No test waits for a real day, and no test uses a real `delay`. Each `Thread.sleep` sits in a poll helper with a deadline or behind a latch. I broke the code five times:
1. I removed the pre-migration backup call: two tests of `MigrationRunnerBackupTest` failed.
2. I changed the prune to `drop(retainedCount + 10)`: `BackupPathsTest` and `DailyBackupJobTest` failed.
3. I changed `cancelAndJoin()` to `cancel()` in `stop()`: `DailyBackupJobTest > stop waits for a backup that is in progress` failed.
4. I wrote straight to the final name: `DatabaseBackupTest > the backup leaves no temporary file behind on Windows` failed.
5. I wrote the connection string into `app.last_error`: no backup test failed. That is MAJOR 2.

I restored the source after each break, and the tree is clean at `15386ae`.

**Scope and hygiene.** The change does what issue #55 asks. The wait for the writer thread in `SqliteDatabase.close()` goes beyond the issue, and it is a correct fix that the new job made visible. The diff holds no secret, no database file, and no build output. The comments read well in Simplified Technical English, with the small point of MINOR 10.

Please fix BLOCKER 1, MAJOR 1, and MAJOR 2. I will review again.

xamcross:
**Maintainer decisions after the reliability review (2026-09-22).**

- BLOCKER 1 (a restore gives the old data back): three parts, each one mandatory. (1) `monitor/backend/README.md` gets a restore list: stop the monitor, delete `octometer.db-wal` and `octometer.db-shm`, copy the backup file over `octometer.db`, start the monitor. (2) A `RestoreTest` does these steps against a real folder and reads the rows of the backup, and a second test keeps the two side files and asserts that the guard of part 3 stops the start. (3) `SqliteDatabase.open` refuses the start with one clear log line when the header of `octometer.db` shows the rollback-journal mode (bytes 18 and 19 equal 1) and a `-wal` file with content exists beside it. That state means a file that a person replaced by hand. The code never deletes the side files by itself.
- MAJOR 1 (`lastBackupDate` after a failure): the job sets the date only after a success. A failed run tries again at the next tick, with one warning for each failure and no flood: a maximum of one warning for each hour.
- MAJOR 2 (the secret safety test cannot fail): the test registers one app with a connection string through the registry, then makes a backup, then reads each byte of the backup file. A leak of the connection string into a column of the database for a moment must make the test fail.
- MINOR 1 (a backward clock change rewrites an older file): the job never writes over an existent backup file; it skips the day with one log line.

The correction round starts after the SQLite review.

xamcross:
**Review (SQLite and file system engineer): CHANGES REQUESTED**

I made my own clone, and I ran `gh pr checkout 146` in it. I ran `./gradlew :monitor:backend:test` two
times in sequence on JDK 21: **197 tests, 0 failures, 0 errors, 0 skipped**, both times. Your test count
is correct. I also broke the code seven times, and I ran my own probe tests.

## Acceptance criteria

| --- | --- | --- |
| `backups/octometer-<yyyyMMdd>.db` with `VACUUM INTO`, one time each day | PASS | My probe wrote `<root>/backups/octometer-20260922.db` with a fixed clock. `BackupPathsTest` holds the due rule. |
| The job keeps 7 files | PASS in behaviour, FAIL in evidence | My probe left exactly 7 daily files with the default count. No test asserts the 7. See MAJOR 2. |
| The runner writes `pre-migrate-v<n>-<timestamp>.db`, then it migrates | PASS | I removed the backup call. Two tests of `MigrationRunnerBackupTest` failed. |
| A test restores a backup and reads the same row count | PASS | `DatabaseBackupTest`. My own probe adds `PRAGMA integrity_check` = `ok` under a write load. |
| A backup file holds no connection string | PASS | D11 keeps each connection string outside `dataDir`. `BackupSecretSafetyTest` guards it. |
| `./gradlew :monitor:backend:test` passes | PASS | 197 tests, 0 failures, two runs in sequence. |
| A comment on the issue records the test result | PASS | The comment exists, with the same count. |

## The checks of my role

**`VACUUM INTO` under a write load (PASS).** I opened a database in WAL mode with 20 000 rows. A second
connection inserted rows without a pause. A third connection ran `DatabaseBackup.writeTo`. The source
held 20 090 rows before the call, and 20 105 rows after it. The backup file held 20 093 rows, and
`PRAGMA integrity_check` gave `ok`. The backup is one consistent moment of the source.

**A failed backup before a migration (PASS).** I put a non-empty folder at the temporary target path
inside a real `backups` folder. `SqliteDatabase.open` threw `BackupFailedException`. The log held one
ERROR line: `The backup before the migration failed. The start stops. BackupFailedException`. The
database file stayed at `PRAGMA user_version` 1. No writer thread stayed alive. The line holds no path.

**A second run on the same day (PASS).** I ran the daily backup three times with the same fixed clock.
The folder held one `octometer-20260922.db` file after each run. The move with `REPLACE_EXISTING`
handles the second run.

**A foreign file of the owner (PASS).** I put `MY-RESTORE-NOTES.txt` and `octometer-copy-of-2026.db`
in the folder. Both stayed after a prune to 7 files. A file that matches the exact pattern
`octometer-<8 digits>.db` is deleted, also when the owner made it. That agrees with the design.

## MAJOR findings

**MAJOR 1. The test run writes a full database copy into the shared temp folder, and nothing deletes
it.**
`monitor/backend/src/test/kotlin/octometer/monitor/ApplicationLifecycleTest.kt:38`,
`.../store/EventStoreTest.kt:18`, `.../store/PragmaTest.kt:17`, `.../store/SqliteDatabaseTest.kt:52`,
`.../apps/AppTotalsRouteTest.kt:33`.
Each of these five call sites passes a temporary folder directly as `dataDir`. `backupsDir(dataDir)`
then names the parent of that folder, which is the system temp root. The backup files land in
`%TEMP%\backups`, outside the folder of each test.
Evidence: I ran only `EventStoreTest`, `PragmaTest`, `AppTotalsRouteTest`, and
`ApplicationLifecycleTest`. That run added four `pre-migrate-v0-*.db` files to `%TEMP%\backups`, and it
rewrote `octometer-20260922.db` there. That folder on my machine now holds 197 leftover files, 932 KB in total. No shutdown hook
removes them, because `backupsDir` sits outside each registered root.
This repeats the defect of issue #142, and the pull request text claims that the nested `data` folder
fixes it. The fix reached three classes only.
Fix: give each of the five call sites a nested `data` folder, the same pattern as
`MigrationRunnerTest.kt:23`. Write `val dataDir = File(root, "data").absolutePath`. Register each root
with `registerTempRoot(root)`.

**MAJOR 2. The acceptance criterion "keeps 7 files" has no test.**
`monitor/backend/src/main/kotlin/octometer/monitor/backup/DailyBackupJob.kt:24`.
I changed `RETAINED_DAILY_BACKUPS` from 7 to 3. The full suite stayed green: 197 tests, BUILD
SUCCESSFUL. The test that the pull request names as evidence gives `retainedCount = 3` of its own, thus
it never reads the constant.
Fix: add a test to `DailyBackupJobTest`. Make 9 daily files. Run
`DailyBackupJob(database, dataDir, clock)` with the default count. Assert that 7 files stay.

**MAJOR 3. The diagnosis of the `ApplicationLifecycleTest` failure is wrong. The 2-second poll hides
the cause, and it breaks again on a slow disk.**
`monitor/backend/src/test/kotlin/octometer/monitor/ApplicationLifecycleTest.kt:184` and
`monitor/backend/src/test/kotlin/octometer/monitor/MonitorServicesTest.kt:98`.
Gradle enables assertions, thus the debug mode of kotlinx-coroutines is on. That mode renames the
thread to `octometer-sqlite-writer @coroutine#4` while a coroutine runs on it. `writerThreadCount()`
compares the name with `==`, thus it counts 0 while the new daily backup runs on the writer thread.
Evidence 1: with a database of 134 MB I sampled the count 40 times. The exact match gave 0 in 31
samples across 775 ms. A `startsWith` match gave 1 in all 40 samples. I saw both names:
`octometer-sqlite-writer` and `octometer-sqlite-writer @coroutine#4`.
Evidence 2: I made the backup take 2 500 ms. Your test, with the 2-second poll, then failed with your
exact CI message: `the store must open one writer thread at the start ==> expected: <1> but was: <0>`.
Evidence 3: I changed the match to `startsWith(WRITER_THREAD_NAME)`. The same slow backup then passed.
Evidence 4: I removed `awaitWriterShutdown(writerExecutor)` from `SqliteDatabase.close()`. The full
suite stayed green. A stale thread also raises the count; it never gives 0.
Fix: change both `writerThreadCount()` helpers to
`Thread.getAllStackTraces().keys.count { it.name.startsWith(WRITER_THREAD_NAME) }`. Keep the poll.
Correct the commit message claim in the pull request text.

**MAJOR 4. A kill between the write and the move leaves a temporary file for ever.**
`monitor/backend/src/main/kotlin/octometer/monitor/backup/DatabaseBackup.kt:43` and
`monitor/backend/src/main/kotlin/octometer/monitor/backup/BackupPaths.kt:50`.
`writeTo` deletes only the temporary name that it writes now. A crash on an earlier day leaves
`octometer-<that day>.db.tmp`. A crash before a migration leaves `pre-migrate-v<n>-<timestamp>.db.tmp`.
`pruneDailyBackups` matches `octometer-<8 digits>.db` only, thus it never removes a `.tmp` file. Each
such file is a full copy of the database, so the folder grows without a limit.
Evidence: I put `octometer-20260101.db.tmp` and `pre-migrate-v1-20260101010101.db.tmp` in the folder. I
then ran a start and a daily backup. Both files stayed.
Fix: sweep the folder at the start of `DailyBackupJob.runOnce`. Delete each file that ends with `.tmp`
and that is older than one hour. Add a test for the sweep.

## MINOR findings

1. `monitor/backend/src/test/kotlin/octometer/monitor/store/Migration002Test.kt:1` now starts with a
   UTF-8 BOM (`EF BB BF`). The diff added it. Remove the three bytes.
2. `monitor/backend/src/main/kotlin/octometer/monitor/backup/DailyBackupJob.kt:90`. The loop sets
   `lastBackupDate` after a failure too, thus a failed backup never retries on the same day. Move the
   assignment into the success path.
3. `monitor/backend/src/test/kotlin/octometer/monitor/backup/BackupPathsTest.kt:89`. The test gives
   `retainedCount = 1`. I widened the pattern to each `.db` file, and the test still passed, because
   the pre-migration name sorts first. With `retainedCount = 0` the same break failed. Use 0.
4. `monitor/backend/src/main/kotlin/octometer/monitor/store/SqliteDatabase.kt:101`. `close()` closes
   the writer connection before it shuts the executor down. A write in flight then uses a closed
   connection. Close the dispatcher first, wait for the executor, then close the two connections.
5. `monitor/backend/src/main/kotlin/octometer/monitor/store/SqliteDatabase.kt:28`.
   `awaitWriterShutdown` drops the `false` result of `awaitTermination`, and it writes no log line. I
   measured a silent stall of 5 006 ms for a write that never ends. Add one WARN line on a timeout.
6. `monitor/backend/src/main/kotlin/octometer/monitor/backup/DailyBackupJob.kt:85`. The branch
   `catch (cancellation: CancellationException) { throw cancellation }` throws again even when the
   loop coroutine is still active. This is lesson 2 of the backend brief. Guard it with
   `currentCoroutineContext().isActive`.
7. The five new test classes make a temporary folder with `Files.createTempDirectory`, and none of them
   calls `registerTempRoot`. A locked file then leaves the folder in `%TEMP%`. I found leftover folders
   `octometer-database-backup-test-*` and `octometer-migration-backup-test-*` with `octometer.db-wal`
   inside. Use `registerTempRoot(root)` in each class.
8. `monitor/backend/src/main/kotlin/octometer/monitor/backup/BackupPaths.kt:11`. The pre-migration
   timestamp has a resolution of one second. Two backups inside one second overwrite each other. Add
   the milliseconds to the format.
9. No test reads the log line of the backup. I checked each line by hand: the INFO line holds a file
   name, and the ERROR line holds a class name. Both are correct today. A sentinel test would keep
   them correct.
10. `monitor/backend/src/main/kotlin/octometer/monitor/backup/DatabaseBackup.kt:55`. The backup runs on
    the writer thread, thus it blocks each write for the whole call. I measured 534 ms for a database
    of 65 MB. A database of 1 GB blocks the ingest route for several seconds each day. Please record
    this choice in the design, or give the backup its own connection.

## The answers to the two open questions

**Does `SqliteDatabase.close()` hang?** No. `close()` from the writer thread itself takes 5 006 ms and
returns; `awaitTermination` times out, and `runCatching` drops the result. A write that never ends
gives the same 5 006 ms, and the writer thread stays alive after `close()` returns. The fallback is
safe, but it is silent. See MINOR 5.

**Does the Windows-only test need a Linux form?** The skip is acceptable. The CI job
`Monitor backend on Windows` runs the same suite on `windows-latest`, thus the assertion runs on CI.
`SecretStoreTest:191` already uses this pattern on `main`. The monitor backend has no guard against a
skip; only `kit/jvm-mongo/build.gradle.kts` has one. A portable form exists: a non-empty folder at the
final path makes `Files.move` fail on Linux and on Windows. Please add that form later, in a separate
issue.

**How long does the stop wait?** `MonitorServices.close()` waits for a full `VACUUM INTO`. With a
database of 65 MB and 200 000 event rows the close took 534 ms, and the 62 MB backup file was complete.
A pure `VACUUM INTO` of a 16 MB database took 329 ms. The wait has no upper limit.

## The merge with pull request #149

I fetched `pull/149/head`, and I merged it into a throwaway branch. One conflict came up, in
`MonitorServices.kt` only. I kept both jobs: two constructor fields, two stop calls in `close()`, and
two start calls in `open()`. After the merge the suite gave **216 tests, 0 failures, 0 errors**. I then
deleted the throwaway branch. The two branches fit together with a small manual resolution.

## Other checks

The diff holds no secret, no database file, and no build output. The commit bodies and the new comments
follow Simplified Technical English. `VACUUM INTO ?` with a bound parameter works with the pinned
`sqlite-jdbc` version; I proved it with my own probe.

xamcross:
Correction round 1 is complete. Two reviews found one BLOCKER and six MAJOR findings; each one is fixed, with a test that fails when the fix is removed.

## BLOCKER 1 (reliability review): a restore did not give the old data back

A person who copies a backup over `octometer.db` by hand, but keeps the old `octometer.db-wal` file, got the old data back with no warning: `PRAGMA journal_mode=WAL` replayed the stale log over the restored file.

- `monitor/backend/src/main/kotlin/octometer/monitor/store/SqliteDatabase.kt:41` adds `RestoredDatabaseNeedsCleanupException`. `SqliteDatabase.kt:63` reads the header of `octometer.db` before it opens a connection, and `SqliteDatabase.kt:195` calls the guard inside `open`. The start refuses, with one log line, when the header shows the rollback-journal mode and a write-ahead log file with content sits beside it. The code never deletes the two side files itself.
- `monitor/backend/README.md` states the restore as a numbered list.
- `monitor/backend/src/test/kotlin/octometer/monitor/backup/RestoreTest.kt` runs the full restore against a real folder, and reads the row count of the backup. A second case keeps the two side files and asserts the guard.

## MAJOR 1 (reliability): a failed backup skipped the rest of the day

`monitor/backend/src/main/kotlin/octometer/monitor/backup/DailyBackupJob.kt:127` now sets `lastBackupDate` only after a success. A failed run retries at the next tick, with the warning throttled to one line for each hour (`DailyBackupJob.kt:144`).

## MAJOR 2 (reliability): the secret safety test could not fail

`monitor/backend/src/test/kotlin/octometer/monitor/backup/BackupSecretSafetyTest.kt:49` now creates the app through `AppRegistryService.createApp`, the real request path, then backs up, then reads every byte of the backup file and of the live database file.

## SQLite MAJOR 1: a test backup leaked into the shared temp root

`monitor/backend/src/main/kotlin/octometer/monitor/config/MonitorConfig.kt:18` adds the config key `backupDir`, defaulting to the sibling folder `backups` of `dataDir`. `ApplicationLifecycleTest.kt`, `EventStoreTest.kt`, `PragmaTest.kt`, `SqliteDatabaseTest.kt` (`SqliteDatabaseReadTransactionTest`), and `AppTotalsRouteTest.kt` now nest `dataDir` under a registered root. `TestConfigs.kt`'s `devConfig`/`prodConfig` give `backupDir` the same default.

## SQLite MAJOR 2: the "keeps 7 files" test never read the constant

`monitor/backend/src/test/kotlin/octometer/monitor/backup/BackupPathsTest.kt:132` seeds 9 files and calls `pruneDailyBackups` with `RETAINED_DAILY_BACKUPS` itself, not a literal value.

## SQLite MAJOR 3: the writer thread name match read 0 during a slow backup

The coroutine debug mode of the test JVM renames the writer thread to `octometer-sqlite-writer @coroutine#N` while a coroutine runs on it. `ApplicationLifecycleTest.kt:195` and `MonitorServicesTest.kt:100` now use `startsWith`, not `==`.

Proof: a scratch test (not part of the shipped suite) opened a real `SqliteDatabase`, ran a 2.5-second write inside `database.write { }` on a launched coroutine, then read the thread name. Result: `octometer-sqlite-writer @coroutine#3`; the exact match gave 0, `startsWith` gave 1. This matches the CI failure of round 0.

## SQLite MAJOR 4: a crash left a `.db.tmp` file for ever

`monitor/backend/src/main/kotlin/octometer/monitor/backup/BackupPaths.kt:91` adds `sweepStaleTempFiles`. `DailyBackupJob.kt:76` sweeps at the start, and `DailyBackupJob.kt:116` sweeps at each daily prune. A test seeds two stale files, one daily and one pre-migration.

## MINOR findings

14 of the 20 raw MINOR findings of the two reviews are corrected (several name the same line from each reviewer's own angle). Two are not corrected: the Windows-only lock test of `DatabaseBackupTest` has no portable Linux form, which the SQLite reviewer called acceptable for now; and a `pre-migrate-*.db` file is never pruned, which `monitor/backend/README.md` now states as a rule instead of a code change.

## Test result and CI

`./gradlew :monitor:backend:test`, run two times in sequence on JDK 21: 292 tests, 0 failures, 0 errors, both times. `./gradlew build` passes one time.

All five CI checks pass on commit `d0a9c63`: Frontend, JVM modules, Monitor backend on Windows, Secret scan, Tracker.

xamcross:
**Review (second review, SQLite and file system engineer): CHANGES REQUESTED**

I made a new clone, and I ran `gh pr checkout 146` in it (head `d0a9c63`). I ran the full suite three
times: **292 tests**, 0 failures in two runs, **1 failure in one run**. Three of my four MAJOR findings
are fixed with a real test. One stays open, and one new MAJOR came with the merge.

## My four MAJOR findings

| --- | --- | --- |
| MAJOR 1. A test backup leaks into the shared temp root | **OPEN** | See MAJOR A below. The five call sites of round 0 are fixed. Two new sites arrived with the merge. |
| MAJOR 2. The "keeps 7" criterion has no test | FIXED | I set `RETAINED_DAILY_BACKUPS` to 3. `BackupPathsTest > pruneDailyBackups with RETAINED_DAILY_BACKUPS...` failed. |
| MAJOR 3. The writer thread name match | FIXED | `ApplicationLifecycleTest.kt:195` and `MonitorServicesTest.kt:100` use `startsWith`. `SqliteDatabase.kt:51` writes a WARN line on the timeout. |
| MAJOR 4. A stale `.tmp` file stays for ever | FIXED | I made `sweepStaleTempFiles` delete nothing. Three tests failed, in `BackupPathsTest` and in `DailyBackupJobTest`. |

## MAJOR A (my MAJOR 1 stays open). Two test classes still write into `%TEMP%\backups`

`monitor/backend/src/test/kotlin/octometer/monitor/users/UserTotalsRouteTest.kt:44` and
`monitor/backend/src/test/kotlin/octometer/monitor/erasure/UserErasureServiceTest.kt:35`.
Both call `SqliteDatabase.open(tempDir.absolutePath)`. `tempDir` is the temp root itself, thus the
default `backupDir` of `SqliteDatabase.kt:188` names `%TEMP%\backups`. The two classes came from
`origin/main` with your two merges (#145 and #147). Your own five call sites are correct now.
Evidence: the folder `%TEMP%\backups` did not exist before my first run. After that run it held 43
files: 27 for the 27 tests of `UserTotalsRouteTest`, and 16 for the 16 tests of
`UserErasureServiceTest`. Each file is a full database copy. The folder now holds 323 files. Nothing
deletes them, because the folder sits outside each registered root.
Fix: nest the data folder in both classes, the same pattern as your other classes. Write
`private val dataDir = File(tempDir, "data").absolutePath`, and pass `dataDir` to
`SqliteDatabase.open`.

## MAJOR B (new). The daily backup job makes each log sentinel test flaky

`monitor/backend/src/test/kotlin/octometer/monitor/LogCapture.kt:74` and
`monitor/backend/src/main/kotlin/octometer/monitor/backup/DailyBackupJob.kt:113`.
`captureLogEvents` reads `appender.list` with `filterNot`. That list is a plain `ArrayList`, and
`ListAppender.append` adds to it with no lock. Your daily backup job writes its INFO line from
`Dispatchers.Default` while the test thread iterates, thus the iteration throws.
Evidence: my first full run failed with
`UserTotalsRouteTest > the captured log holds no line with the filter text, at every log level`, and the
cause was `java.util.ConcurrentModificationException` at `LogCapture.kt:74`. The test starts the module
inside the capture block (`UserTotalsRouteTest.kt:461`), thus the job starts inside that block. My runs
2 and 3 passed, and the class alone passed three times, so this is a race, not a constant failure.
Before this branch, `module()` had no background logger, thus nothing could append from another thread.
Fix: make the captured list thread-safe in `LogCapture.kt`. Use
`Collections.synchronizedList(ArrayList<ILoggingEvent>())` inside your own appender, and read it with
`synchronized(list) { list.toList() }`. Pull request #149 adds a second background job, so the helper
needs this fix in any case.

## My 10 MINOR findings

| # | Point | State | My evidence |
| --- | --- | --- | --- |
| 1 | A UTF-8 BOM in `Migration002Test.kt` | FIXED | The first three bytes are `p a c`. |
| 2 | No retry after a failed daily backup | FIXED | `DailyBackupJob.kt:127` sets `lastBackupDate` only after a success. The warning is throttled. |
| 3 | The prune test could not fail at `retainedCount = 1` | FIXED | `BackupPathsTest.kt:121` uses `retainedCount = 0`. |
| 4 | The close order of `SqliteDatabase.close()` | FIXED | `SqliteDatabase.kt:160` closes the dispatcher and waits, then it closes the two connections. |
| 5 | A silent `awaitTermination` timeout | FIXED | `SqliteDatabase.kt:51` writes one WARN line. |
| 6 | The `CancellationException` rethrow | FIXED | `DailyBackupJob.kt:132` throws again only when `currentCoroutineContext().isActive` is false. |
| 7 | No `registerTempRoot` in the five new test classes | FIXED | Each of the five classes registers its root. |
| 8 | A 1-second pre-migration timestamp | FIXED | The format is `yyyyMMddHHmmssSSS`. |
| 9 | No log sentinel for the backup lines | FIXED | `DailyBackupJobTest.kt:197` and `:243` read the captured events. |
| 10 | The backup blocks the writer thread | FIXED | `DailyBackupJob.kt:107` records the cost and names a later issue. |

## The two points that you did not correct

**The Windows-only lock test.** I agree. The CI job `Monitor backend on Windows` runs the same suite,
thus the assertion runs on CI. `SecretStoreTest` already uses the pattern on `main`. Please open a
separate issue for a portable form.

**A pre-migration file that no prune removes.** I accept the reason, with one note. The README states a
bounded count: one file for each migration that the install applies. That holds for a good start. A
start that fails after the backup step writes a new file at each attempt, thus the real bound is the
count of attempts. My own test runs show the effect: each fresh test database adds one file. Please
record the note, or prune a pre-migration file beyond a fixed count in a later issue.

## The restore guard (`SqliteDatabase.kt:63`)

PASS. I ran three probes of my own. A fresh empty folder starts, because the guard returns when
`octometer.db` is absent. A normal WAL database starts with a `octometer.db-wal` file of 78 312 bytes
beside it, because the header holds the WAL version 2. A file from `VACUUM INTO` with no side file
starts, because the guard returns when the `-wal` file is absent. The guard refuses only the header
value 1 together with a `-wal` file that holds content, which is the state after a hand restore
without step 3.

## Other checks

The diff holds no secret, no database file, and no build output. The new comments and the README follow
Simplified Technical English. The new rule "a second run of the same day skips the write" is safe: it
protects a backup that the day already holds.

Please fix MAJOR A and MAJOR B. Each other point is closed.

xamcross:
**Review (second review, reliability engineer): APPROVE**

I made a second clone, ran `gh pr checkout 146`, and tested commit `d0a9c63` on JDK 21, on Windows 11. Each number below comes from a command that I ran in my own clone. I restored each file after each experiment, and my clone is clean at `d0a9c63`.

## BLOCKER 1 — fixed. I ran the restore myself, with a real stale write-ahead log.

I built a real crash snapshot: I opened a real `SqliteDatabase`, wrote 1000 event rows, ran one daily backup, wrote 5000 more rows, then copied `octometer.db`, `octometer.db-wal`, and `octometer.db-shm` aside while the process still held them. The write-ahead log file held 716 912 bytes of real frames.

- I copied the backup over `octometer.db` and **skipped step 3 of the README on purpose**. The start refused: `RestoredDatabaseNeedsCleanupException`. The log held **exactly one** line about it: `[ERROR] A stale write-ahead log file lies beside a restored database. Delete octometer.db-wal and octometer.db-shm, then start again.` No captured line held the folder path. The write-ahead log file was still on the disk after the refusal, thus the code deletes nothing by itself.
- I then followed the full README list: I deleted `octometer.db-wal` and `octometer.db-shm`, and started again. The read gave **1000 rows**, the row count of the backup.

The same steps in round 0 gave 6000 rows and `quick_check = ok`, with no warning. The defect is gone.

I read `RestoreTest.kt`. The happy case walks the five steps of the README against a real folder, and it asserts the row count of the backup, not the row count before the stop. The guard case keeps a write-ahead log file with content and asserts the throw, the message, and that the file stays. I removed the call `refuseIfRestoredWithStaleWal(folder)` at `SqliteDatabase.kt:195` for a moment: `RestoreTest > a restore that keeps a stale write-ahead log file refuses the start` failed, and the happy case still passed. The guard test can fail. I restored the line.

One note, not a finding: the maintainer decision names the order "delete the side files, then copy the backup", and the README names "copy, then delete". The two orders end in the same state, and nothing opens the database in between, thus the result is the same. The guard catches the mistake in each order.

## MAJOR 1 — fixed. The date moves only after a success, and the retry works.

`DailyBackupJob.kt:127` sets `lastBackupDate` inside the success path. `DailyBackupJob.kt:134` leaves it alone after a failure.

My check: I put a plain file in place of the backup folder, started the job with a tick of 20 ms and a fixed clock, and let about 25 ticks fail. I then repaired the folder **inside the same calendar day**. Result: `files after the repair, same calendar day = [octometer-20260108.db]`. The same experiment in round 0 gave `[]`.

The throttle holds: over those same ~25 failed ticks the log held **one** line, `[ERROR] The daily backup failed. BackupFailedException`. `warnOnThrottle` at `DailyBackupJob.kt:144` bounds the line to one for each hour of the clock.

I put back the old line `lastBackupDate = LocalDate.now(clock)` in the failure path for a moment: `DailyBackupJobTest > a failed backup tries again at the next tick, and writes the file once it succeeds` failed. I restored the file.

## MAJOR 2 — fixed. The secret test can fail now.

`BackupSecretSafetyTest.kt:49` registers the app with `AppRegistryService.createApp`, the real request path, and then reads each byte of the backup file and of the live database file.

I wrote the connection string into the column `app.last_error` inside `insertApp` for a moment. `BackupSecretSafetyTest > the daily backup file holds no connection string of an app the registry created` **failed**. In round 0 the same leak left that test green. I restored the file.

## Each MINOR finding of my first review

| # | The finding | State | My evidence |
|---|---|---|---|
| 1 | A backward clock change rewrites an older backup file | **Fixed** | `BackupPaths.kt:57` uses `isAfter`. My check: with `lastBackupDate = 2026-01-09` and a clock at `2026-01-08T23:55`, the call returns `false`. `DailyBackupJob.kt:103` also refuses to write over an existent file: I put text in the target file, ran `runOnce()`, and the text stayed, with the line `The daily backup skipped octometer-20260108.db, because that file exists already.` I removed that guard for a moment; `DailyBackupJobTest > runOnce never writes over an existing backup file` failed. |
| 2 | The success line holds no size | **Fixed** | `DailyBackupJob.kt:119` writes `The daily backup wrote {} ({} bytes).` `DailyBackupJobTest.kt:246` asserts the exact line, and `DailyBackupJobTest.kt:252` asserts that the line holds no folder path. |
| 3 | The prune deletes a backup with no log line | **Fixed** | `BackupPaths.kt:79` writes `The daily prune removed {} file(s).`, a count and no file name. No test reads this line; the fix is one statement, thus I accept it. |
| 4 | The exception cause holds two absolute paths | **Fixed** | `DatabaseBackup.kt:66` and `DatabaseBackup.kt:112` now build the exception with no cause. A stack trace of a failed start can no longer carry a path. |
| 5 | The unguarded `CancellationException` rethrow | **Fixed** | `DailyBackupJob.kt:132` reads `if (!currentCoroutineContext().isActive) throw cancellation`, the form of `Application.kt`. A rejected dispatch now writes a throttled line instead of a silent death. |
| 6 | The new test classes leave a folder in `%TEMP%` | **Fixed** | Each of the five classes that I named uses `registerTempRoot`: `DailyBackupJobTest.kt:30`, `DatabaseBackupTest.kt:21`, `BackupSecretSafetyTest.kt:33`, `BackupPathsTest.kt:85`, `MigrationRunnerBackupTest.kt:24`. `RestoreTest.kt:26` does too. Three older classes that this pull request touches still do not: `Migration002Test.kt:20`, `MigrationRunnerTest.kt:21`, `SqliteDatabaseTest.kt:24`. Each one deletes its folder in `@AfterTest`. That is outside my finding, and it belongs to issue #142. |
| 7 | The temp-then-move rule rests on one Windows-only test | **Open, accepted** | `DatabaseBackupTest.kt` still guards that test with `Assumptions.assumeTrue`. The SQLite reviewer accepted it, and two portable tests of the same class cover the move path in part. I accept it as a MINOR that stays. |
| 8 | A `pre-migrate-*.db` file is never pruned | **Open, documented** | `monitor/backend/README.md` lines 17 to 19 state the rule: one file for each migration, a small and bounded count. I accept the document in place of the code change. |
| 9 | No integrity check on the fresh backup file | **Fixed in the code, with no test** | `DatabaseBackup.kt:47` calls `checkIntegrity`, which runs `PRAGMA quick_check` on the temporary file before the move. I removed that call and ran the full suite: **292 tests, 0 failures**. No test proves this rule. It is a new rule with no guard, thus a later change can drop it in silence. A cheap test: write a good backup, damage one page of the temporary file, and assert `BackupFailedException`. MINOR, and it does not block. |
| 10 | The passive voice in a comment | **Fixed** | `BackupPaths.kt:63` now reads "stays". |

## The schedule, in short form

- The first backup at the start: **yes**. `isDailyBackupDue(clock, null)` returns `true`, thus the first tick writes the file.
- One backup for each **local** calendar day: `MonitorServices.open` defaults to `Clock.systemDefaultZone()`, and `LocalDate.now(clock)` reads the zone of that clock. A second run of the same day skips the write. `monitor/backend/README.md` states the rule.
- After a wake past midnight: I ran a moving clock at `2026-01-08T23:00`, saw `[octometer-20260108.db]`, then moved the clock to `2026-01-09T09:00`, the state after a wake. Inside one tick the folder held `[octometer-20260108.db, octometer-20260109.db]`. The job writes **one** backup at once, and it does not wait for the next midnight. It writes no extra file for a day that the machine slept through.

## The test count and the hygiene

`./gradlew :monitor:backend:test` on JDK 21, two runs in sequence: **292 tests, 0 failures, 0 errors, 0 skipped**, both times. That matches the count of the author. A third run, after I restored each of my five experiments, gave the same count. The five CI checks pass on `d0a9c63`. The diff holds Kotlin files and one README file only: no secret, no database file, and no build output.

## The verdict

No BLOCKER and no MAJOR finding stays. One MINOR item is new (finding 9 has no test), and two earlier MINOR items stay open with an accepted reason. I approve. Please open a small follow-up issue for the missing test of the integrity check.

xamcross:
## Correction round 2, evidence

**MAJOR A (test backup leaks to the shared temp root).** `UserTotalsRouteTest.kt:44` and `UserErasureServiceTest.kt:35` opened the database with the raw temp root as `dataDir`. `monitor/backend/src/test/kotlin/octometer/monitor/TestConfigs.kt` gets a new overload, `testDataDir(root: File)`. It nests the data folder under the registered root and makes it. Both classes now call it. The merge of pull request #149 also brought a new class with the same defect, `RetentionPurgeTest.kt`. This round fixes that class too, with the same nested `dataDir` pattern as the other test classes.

File count of `%TEMP%\backups`: 452 files before a full run, 452 after, for each run before the merge. After the merge, the count held at 476 across three full runs, with the fix of `RetentionPurgeTest.kt` in place. The task also removes each old file from that folder, and the folder itself, because each file was a 4 KB test database from today.

**MAJOR B (a log race with a background job).** `LogCapture.kt:63` now gives the appender a `Collections.synchronizedList`. The read takes a copy of the list under its lock, before the filter runs. Pull request #149 added a second background job, the retention purge job, and this fix holds for both jobs together.

**Three full runs, JDK 21, after the merge of `origin/main` (commit 251546d, pull request #149):**

| Run | Command | Result |
| --- | --- | --- |
| 1 | `./gradlew :monitor:backend:test --rerun-tasks` | BUILD SUCCESSFUL, 316 tests, 0 failed |
| 2 | `./gradlew :monitor:backend:test --rerun-tasks` | BUILD SUCCESSFUL, 316 tests, 0 failed |
| 3 | `./gradlew :monitor:backend:test --rerun-tasks` | BUILD SUCCESSFUL, 316 tests, 0 failed |

CI on commit `14d1bba`: all 5 checks pass (Frontend, JVM modules, Monitor backend on Windows, Secret scan, Tracker). The Windows job took 2m40s.

Files: `monitor/backend/src/test/kotlin/octometer/monitor/TestConfigs.kt`, `monitor/backend/src/test/kotlin/octometer/monitor/users/UserTotalsRouteTest.kt`, `monitor/backend/src/test/kotlin/octometer/monitor/erasure/UserErasureServiceTest.kt`, `monitor/backend/src/test/kotlin/octometer/monitor/retention/RetentionPurgeTest.kt`, `monitor/backend/src/test/kotlin/octometer/monitor/LogCapture.kt`.

xamcross:
**Review (third review, SQLite and file system engineer): CHANGES REQUESTED**

I made a new clone, and I checked out head `14d1bba`. Both MAJOR findings of my second review are
fixed. One new MAJOR came from the fix itself, and it needs one line.

## My two MAJOR findings

**MAJOR A (the leak into the shared temp root): FIXED.** I deleted
`C:\Users\xamcr\AppData\Local\Temp\backups` first. I then ran the full suite three times with
`--rerun-tasks`. The folder never came back, not one time. `TestConfigs.kt:62` adds the overload
`testDataDir(root)`, and `UserTotalsRouteTest.kt:44`, `UserErasureServiceTest.kt:36`, and
`RetentionPurgeTest.kt:36` each nest the data folder under a registered root.

**MAJOR B (the flaky log sentinel): FIXED.** `LogCapture.kt:63` gives the appender a
`Collections.synchronizedList`, and `LogCapture.kt:70` takes a copy under the same lock before the
filter. Three full runs with the two background jobs active gave **316 tests, 0 failures, 0 errors, 0
skipped**, each time. Your test count is correct.

**The merge of `MonitorServices.kt`: correct.** `open` starts the purge job, then the daily backup job.
`close()` stops the daily backup job, then the purge job, then the store. The catch of `open` stops
each started job, with a null check for each one, then it closes the store.

## MAJOR C (new). The nesting fix made one assertion dead

`monitor/backend/src/test/kotlin/octometer/monitor/erasure/UserErasureServiceTest.kt:172`.
The line reads `File(tempDir, "octometer.db-wal")`. The database now sits at `tempDir/data`, thus that
path never holds a file. The assertion on line 173 accepts an absent file, so it always passes. Commit
`14d1bba` corrected the same mistake in `RetentionPurgeTest`; this one stayed.
Evidence: I made `checkpointAfterCommit` never run, and I left the flag at `true`. The test
`a checkpoint with no other reader completes and reports checkpointed true` still passed. I then
changed line 172 to the nested path and ran the same break again. The test failed. The current line
proves nothing; the nested path restores the proof.
Fix: write `val walFile = File(File(tempDir, "data"), "octometer.db-wal")`.

Please make that one change. Each other point of my first review and of my second review is closed.

xamcross:
## Correction round 3, evidence

**MAJOR C (a dead WAL assertion).** `UserErasureServiceTest.kt:172` read the wal file at the raw
`tempDir`, but the nesting fix of MAJOR A moved the database to `tempDir/data`. The path now reads
`File(File(tempDir, "data"), "octometer.db-wal")`. Proof: with the TRUNCATE checkpoint call removed
for a moment, `a checkpoint with no other reader completes and reports checkpointed true` failed. The
checkpoint call was restored, and the test passed again.

I grepped the test sources for each other `octometer.db` and `octometer.db-wal` path built from a root
or a temp folder. Each other site already uses a nested `dataDir`. No other site needed a change.

Two full runs, JDK 21, `./gradlew :monitor:backend:test --rerun-tasks`: BUILD SUCCESSFUL, both times.

CI on commit `bbad8af`: all 5 checks pass (Frontend, JVM modules, Monitor backend on Windows 2m12s,
Secret scan, Tracker).

xamcross:
**Review (third review, reliability engineer): APPROVE**

I made a third clone, ran `gh pr checkout 146`, and tested commit `bbad8af` on JDK 21, on Windows 11. I restored each file after my experiment, and my clone is clean.

**Round 3, the one line (`bbad8af`).** `UserErasureServiceTest.kt:172` now reads `File(File(tempDir, "data"), "octometer.db-wal")`. I broke the checkpoint of the erasure for a moment: I changed the mode text of `PRAGMA wal_checkpoint(...)` in `UserErasureService.kt:195` to an invalid value. Result: `UserErasureServiceTest > a checkpoint with no other reader completes and reports checkpointed true` **FAILED**, with the message `expected an empty or an absent WAL file`. The assertion reads the real file now, and it can fail. I restored the line.

**Round 2, the merge of the two jobs.** `MonitorServices.kt` holds both jobs correctly:
- `open` starts the purge job before the orphan sweep, and the daily backup job after it.
- `close()` runs `runBlocking { dailyBackupJob.stop() }`, then `retentionPurgeJob.stop()`, then `database.close()`. Both jobs stop before the store closes.
- The catch of `open` runs `dailyBackupJob?.let { runBlocking { it.stop() } }` and `retentionPurgeJob?.stop()`, then `database.close()`. The `?.` guard covers a throw before a job starts, thus a throw in a later step of `open` stops each job that ran.

**`LogCapture.kt`.** The appender now holds `Collections.synchronizedList(ArrayList())`, and the read takes a copy inside `synchronized(list)`. An iteration over a synchronized list needs its own lock, and this code takes it. Two background jobs (the daily backup and the purge) write from `Dispatchers.Default` while a test reads, thus the race is closed.

**The tests.** `./gradlew :monitor:backend:test`, two runs in sequence, the second with `--rerun-tasks`: **316 tests, 0 failures, 0 errors, 0 skipped**, both times. That matches the count of the author. The folder `%TEMP%\backups` did not exist before my runs, and it does not exist after them, thus the `testDataDir` helper stops the leak. The five CI checks pass on `bbad8af`.

No BLOCKER and no MAJOR finding stays. My approval of round 1 holds. I approve.

Issue #64, opened 2026-09-21
Write the run guide


xamcross:
**Blocked by #56.** The guide describes the upgrade script of #56 and the erasure route of #61. Steps 2, 4, and 6 can start now.

**Goal.** One document tells the owner how to install, operate, back up, restore, and upgrade the monitor.

**Implementation steps.**
1. Create `docs/run-guide.md` with the section on install and start (D34, D35).
2. Write the section on the config: each key of D2 and the precedence.
3. Write the section on backup, restore, and upgrade (D36). It states that a backup holds no connection string, and it gives `scripts/upgrade.ps1 -Version X.Y.Z` with the rule "upgrade the monitor before an app".
4. Write the section on secrets: the BitLocker check `manage-bde -status C:` and the rotation of a reader password (section 4.4).
5. Write the section on erasure: the order of D15, and the command for the route of #61: `curl.exe -X DELETE -H "Origin: http://127.0.0.1:7431" -H "Content-Type: application/json" "http://127.0.0.1:7431/api/apps/<appId>/events?userId=<id>"`.
6. Write the section on risks: a browser profile without extensions, one manual sleep test, and the residual risks of section 8.

**Acceptance criteria.**
- [ ] `docs/run-guide.md` holds the six sections of the steps.
- [ ] The guide tells the owner to add each connection string again after a restore.
- [ ] A reader can restore a backup on a clean machine with the guide alone.
- [ ] A comment on this issue records the result of one restore with the guide.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, decisions D2, D11, D15, D26, D34 to D36, section 4.4, section 8.

xamcross:
Note on 2026-09-22 from the maintainer, from the privacy review of #147 (issue #61): https://github.com/xamcross/octometer/pull/147#issuecomment-5769624408

The erasure route takes the user id in the URL (design decision D13). A curl example with a real user id lands in the shell history of the owner. When this issue writes the curl example, give the user id as a shell variable, and add one sentence: clear the shell history, or read the id from a file. The README of `monitor/backend` states the three steps of design decision D15: the app erases its copy (#35), one poll cycle passes, then the owner calls this route. The steps of this issue do not change.

xamcross:
Body edit on 2026-09-22: #61 closed. The first line now names only the open blockers: #56.

PR #147, opened 2026-09-22
feat(monitor): add the user erasure route

### LINKED ISSUE #61: Delete the events of one user id in the monitor

xamcross:
The route `DELETE /api/apps/{appId}/events?userId=<id>` deletes each event of one user id of
one app. It also deletes each event with `user_id IS NULL` of a session of that user (rules
E1 and E2, contract rule C43). A row of a second user in the same session stays. The function
reads the session list of the user first, inside the write transaction of the delete. It runs
a `PASSIVE` checkpoint right after the commit, with a retry, then one `TRUNCATE` checkpoint for
the file length (round 2, MAJOR A below). The route sits in its own package
`octometer.monitor.erasure`.

## Correction round 1 (privacy review and SQL review)

- [x] BLOCKER 1 (privacy): the code now reads the answer row of `PRAGMA
  wal_checkpoint(TRUNCATE)`. On `busy = 1` it retries, a maximum of 5 times with a 100 ms
  pause. A busy last try keeps the answer at 200 with `"checkpointed": false`. It writes one
  warning with the app id only. Evidence: `UserErasureServiceTest.a busy checkpoint retries,
  then reports checkpointed false with one warning` (a second reader holds a read transaction
  through `database.read`). Also `UserErasureServiceTest.a checkpoint with no other reader
  completes and reports checkpointed true`. The old code was restored for a moment. The busy
  test then failed on `result.checkpointed`. The fix was restored, and the suite passed again.
- [x] MAJOR 1 (privacy, two steps of D15 in place of three): `monitor/backend/README.md` and
  the comment at the top of `UserErasureService.kt` now state the three steps in order. First
  the app erasure of #35, then one poll cycle, then this route.
- [x] MAJOR 2 (privacy, no warning about the user id in the URL): `monitor/backend/README.md`
  now warns that the user id travels in the query string. It lists the shell history, a
  proxy log, and a future request log as the places that can hold it.
- [x] MAJOR 3 (privacy) and MAJOR 2 (SQL) (a row with `kind = 2` survives): a third delete,
  `SQL_DELETE_USER_OTHER_KIND`, removes a row of the user with a kind other than 0 or 1. It
  keeps the two existing `INDEXED BY` deletes, per the choice the issue gave. `EXPLAIN QUERY
  PLAN` of the new statement is `SEARCH event USING COVERING INDEX event_session (app_id=?)`.
  This matches the 8 ms the SQL review measured. Evidence:
  `UserErasureServiceTest.the erasure removes a row of a kind other than 0 or 1, and counts
  it`. The guard delete was removed for a moment. The test then failed on the count and on
  the remaining row. The fix was restored.
- [x] MAJOR 1 (SQL, keep `INDEXED BY`): unchanged, plus a new test,
  `UserErasureServiceTest.the three indexes of the erasure exist in sqlite_master`. It fails
  the build when a schema change drops or renames `event_agg`, `event_start`, or
  `event_session`.

### MINOR findings: 9 corrected, 1 not corrected

Corrected:
1. The test class now registers its temp root with `registerTempRoot` (issue #142), on top of
   the existing `tearDown` delete.
2. A `userId` over 254 characters (contract rule C6) gives 400 with a fixed sentence. Evidence:
   `UserErasureRoutesTest.DELETE with a userId over 254 characters gives 400`. The check was
   removed for a moment, and the test failed. The fix was restored.
3. The log sentinel test now reads `throwableProxy?.message` too. It runs on the 200, the
   400, and the 404 path (three tests in `UserErasureRoutesTest`).
4. Each comment sentence in the changed files now has at most 25 words.
5. The checkpoint test now asserts `result.checkpointed`, the WAL file size, and the byte
   search, not the call alone.
6. `UserErasureServiceTest.the delete statements use an index` now asserts the exact
   `EXPLAIN QUERY PLAN` text of each statement. A later index-shape change now shows here.
7. `LogCapture.kt`: `captureErrorLogEvents` now calls `captureLogEvents` with a level filter.
   It no longer holds a second copy of the capture logic.
8. `LogCapture.kt`: `captureLogEvents` raises the root log level to TRACE for its block. It
   restores the old level in `finally`. A DEBUG line no longer passes the sentinel test by
   default.
9. `monitor/backend/README.md` states that only `event` holds a `user_id` column today. A
   migration that adds a user id to another table must extend `UserErasureService`, and add
   a test.

All 7 MINOR findings of round 1 stand corrected; SQL review MINOR 7 (`PASSIVE` before
`TRUNCATE`) moved to MAJOR A of round 2 below, and is now fixed.

## Correction round 2 (second privacy review, second SQL review)

- [x] MAJOR A (SQL): the round 1 retry held the writer thread for 28 seconds at the default
  `busy_timeout`, and `TRUNCATE` gave a false `checkpointed: false` in the common case, because
  the erased bytes sit in `octometer.db`, not in the WAL file. The fix: `busy_timeout=0` for
  the checkpoint block, a `PASSIVE` checkpoint that reports `checkpointed` when `log ==
  checkpointed` (a maximum of 5 tries, 100 ms apart), then one `TRUNCATE` checkpoint for the
  file length only. Evidence: `UserErasureServiceTest.the write block completes within 2
  seconds with a busy reader` (0.486 s measured) and `...a reader on the newest snapshot still
  lets the checkpoint report checkpointed true`. The old form was restored for a moment; the
  time test then took about 28 seconds and failed the 2-second bound. The fix was restored.
- [x] MINOR 8 (SQL): the session list read covered only `kind = 0` and `kind = 1`, so a session
  whose only user row held a different kind kept its anonymous rows (contract rule C43 covers
  each kind). A third read, `SQL_SESSION_IDS_OTHER_KIND` on `event_session`, closes the gap.
  Its plan is `SEARCH event USING COVERING INDEX event_session (app_id=?)`, the same shape as
  the guard delete. Evidence: `UserErasureServiceTest.the erasure removes an anonymous row of a
  session whose only user row has a kind other than 0 or 1`. The read was removed for a moment;
  the test then failed. The fix was restored.
- [x] MINOR 9 (SQL): `NON_SERVER_LOGGER_PREFIXES` is renamed to `KNOWN_URL_LOGGER_PREFIXES`.
  The comment names issue #150, which removes the filter later.
- [x] MINOR 10 (SQL): `monitor/backend/README.md` no longer places the erased byte in the WAL
  file. It now states that the byte stays in `octometer.db` until a checkpoint copies the new
  page over the old one, and it warns against deleting a store file by hand.
- [x] MINOR 11 (SQL): design decision D13 now lists `checkpointed` beside `deleted` in the
  erasure answer.
- [x] MINOR 6 of round 1 (SQL, a chunk note): a code comment on `eraseUserEvents` states the
  measured cost of one large erasure (538 ms, 15.6 MB of WAL, 50 000 events in one
  transaction), and why no chunk form exists.

Test run after round 2: `./gradlew :monitor:backend:test`, two runs in sequence, JDK 21.
`BUILD SUCCESSFUL` each time, 261 tests, 0 failures, 0 errors, 23 test classes (from
`monitor/backend/build/test-results/test/TEST-*.xml`). `./gradlew build` also passes once.

### A finding outside the scope of #61

Raising the log capture to TRACE level (MINOR 4 of the SQL review) surfaced a framework-level
DEBUG line with no tie to the erasure route. The Ktor `StatusPages` plugin
(`io.ktor.server.plugins.statuspages.StatusPages`) logs "No handler found for status code ...
for call: <url>" for a normal answer with no registered status handler. `<url>` is the full
request URL, including any query string. This never runs today, because the root log level of
`logback.xml` is INFO. A later switch to DEBUG logging would change that. Such a switch would
then write the user id of every request with one in its query string or its path. This would
happen on every route, not only this one. `LogCapture.kt` filters this logger name out of the
sentinel check, with a comment that names the risk. This is a design-level, cross-route
finding. Please file it as a new issue.

## Acceptance criteria

- [x] The rows of that user id are gone, and the rows of each other user id stay. Evidence:
  `UserErasureServiceTest.the erasure removes the rows of the user and keeps a row of a
  second user in the same session`. Also `UserErasureServiceTest.the erasure of one app leaves
  the same user id of a second app untouched`.
- [x] A row with `user_id IS NULL` of a session of that user is gone. Evidence:
  `UserErasureServiceTest.the erasure removes an anonymous row of a session that also holds a
  row of the user`.
- [x] A row of a second user in the same session stays. Evidence: the same test as the first
  item. The row of `user-b` stays in the result set.
- [x] A row with `user_id IS NULL` of another session stays. Evidence:
  `UserErasureServiceTest.an anonymous row of a different session stays`.
- [x] The field `deleted` counts the two row groups. Evidence: `UserErasureResult.total` sums
  `deletedUserRows` and `deletedAnonymousRows`. Asserted in
  `UserErasureRoutesTest.DELETE removes the rows of the user and gives 200 with the deleted
  count` and in `...DELETE also removes the anonymous rows of a session of the user`.
- [x] A WAL checkpoint runs after the delete commits, and the answer states whether it
  completed. Evidence: `UserErasureServiceTest.a checkpoint with no other reader completes and
  reports checkpointed true`, and the busy-path test named above. Round 2 (MAJOR A) changed the
  privacy result to a `PASSIVE` checkpoint; see that section for the current form.
- [x] The user id is a bound parameter and appears in no line of the captured log. Evidence:
  `UserErasureRoutesTest.no log line holds the user id or the query string of the request, on
  200/400/404`. Also the byte-search test `a byte search of the store files finds no user id
  after the erasure`.
- [x] The captured log holds no query string. Evidence: the same log tests assert no line
  holds the text `userId=`.
- [x] The response is 200 with the fields `deleted` and `checkpointed`, and an unknown app id
  gives 404. Evidence: `UserErasureRoutesTest.DELETE with an unknown app id gives 404`, and
  the 200 tests above.
- [x] `./gradlew :monitor:backend:test` passes. Evidence: two runs in sequence after the merge
  of `origin/main` (commit ba2ea5c). `BUILD SUCCESSFUL` each time, 261 tests, 0 failures, 0
  errors (23 test classes, from `monitor/backend/build/test-results/test/TEST-*.xml`).
  `./gradlew build` also passes once, for the whole project.
- [x] A comment on the issue records the test output. Evidence: the comment posted on #61
  with the original submission, and the pull request comment with each correction round.
- [x] CI is green. Evidence: `gh pr checks 147`. Frontend, JVM modules, Monitor backend on
  Windows, Secret scan, and Tracker all `pass`.

## Confirm results

- The design document names D13 (the route shape) and D15 (personal data, the checkpoint
  rule, the C43 width). Both apply as read in
  `docs/superpowers/specs/2026-09-21-octometer-design.md`.
- Rules E1 and E2 come from the design brief in the first comment of issue #101. They also
  come from the body edit of issue #61 on 2026-09-21. The erasure reads the session list of
  the user first. It then deletes the rows of that user, and the anonymous rows of those
  sessions.

## Not done

- No route change for #35 (the kit store). That issue is out of scope here.
- No UI control for this route. Issue #64 gives the `curl` command with the `Origin` header.
- The `StatusPages` TRACE line (see the finding outside the scope of #61 above). Issue #150
  owns it.

Closes #61

Generated with Claude Code


xamcross:
**Review (privacy engineer): CHANGES REQUESTED**

I made my own clone, I ran `gh pr checkout 147`, and I checked each item with JDK 21
(`C:/Users/xamcr/.jdks/ms-21.0.9`). I ran `./gradlew :monitor:backend:test` one time:
`BUILD SUCCESSFUL`, 19 test classes, 187 tests, 0 failures. That count agrees with the pull
request text. I deleted my clone after the review.

The erasure width is correct. The transaction is correct. One privacy control fails without a
sound, and the owner runbook is shorter than design decision D15. Thus I request a change.

## The acceptance criteria of #61

I seeded my own data set: a session with anonymous events before a sign-in, a session with two
user ids, a session of the second user only, a session with anonymous events and no link to the
target user, and the same user id in a second app. Then I ran `eraseUserEvents`.

| # | Criterion | Result |
|---|---|---|
| 1 | The rows of that user id are gone, the rows of each other user id stay | PASS |
| 2 | A row with `user_id IS NULL` of a session of that user is gone | PASS |
| 3 | A row of a second user in the same session stays | PASS |
| 4 | A row with `user_id IS NULL` of another session stays | PASS |
| 5 | The field `deleted` counts the two row groups | PASS |
| 6 | `PRAGMA wal_checkpoint(TRUNCATE)` runs after the delete | FAIL, see BLOCKER 1 |
| 7 | The user id is a bound parameter and stays out of each log line | PASS |
| 8 | The captured log holds no query string | PASS |
| 9 | 200 with `deleted`, and 404 for an unknown app id | PASS |
| 10 | `./gradlew :monitor:backend:test` passes | PASS, 187 tests |
| 11 | A comment on the issue records the test output | PASS |

Evidence of the criteria 1 to 5, from my own probe:

```
deleted=5 user=2 anon=3
rows of app A after: [e6 sess-beta-0002 second-user k0] [e7 sess-gamma-0003 second-user k0]
                     [e8 sess-delta-0004 null k0]
rows of app B after: [e9 sess-alpha-0001 target-user k0]
a user with no events: deleted=0, no error
```

The three anonymous rows of the two sessions of the target user are gone. The anonymous row of
the unrelated session stays. The row of the second user in the shared session stays. The same
user id in the second app stays. This matches rule E1, rule E2, and contract rule C43.

**The transaction (my check 3).** I put a throw between the session read and the first delete,
and then between the two deletes. In both cases each of the three rows was still in the table
after the call. The read and the two deletes are one unit of work. PASS.

**The route (my check 7).** Through `testApplication`, with the store open:

```
Cache-Control: no-store        Access-Control-Allow-Origin: absent
Content-Security-Policy: present
a wrong Host header: 403       a DELETE without an Origin header: 403
appId=abc: 400 {"error":"The app id must be a whole number."}
appId=88888: 404 {"error":"The app is not registered."}
appId=-1 and appId=0: 404      appId=99999999999999999999: 400
userId of two blanks: 400 {"error":"Give a user id in the query parameter userId."}
GET on the same path: 404
```

No case gave 500. The fixed sentence of each 400 answer and of the 404 answer holds no user
data. This matches D12 and D13.

**The tests (my check 8).** I broke the code five times. Each break gave the exact expected
failure:

1. The anonymous delete removed: `the erasure removes an anonymous row of a session that also
   holds a row of the user` FAILED.
2. The session filter of the anonymous delete removed: `an anonymous row of a different session
   stays` FAILED.
3. The `user_id` filter of the user delete removed: `the erasure removes the rows of the user
   and keeps a row of a second user in the same session` FAILED.
4. The checkpoint call removed: `the erasure runs a WAL checkpoint after the commit` FAILED,
   and `a byte search of the store files finds no user id after the erasure` FAILED too.
5. A log line with the user id added to the route: `no log line holds the user id or the query
   string of the request` FAILED.

I found no test that cannot fail. The byte-search test is real: break 4 made it fail.

---

## BLOCKER 1. The failed WAL checkpoint stays silent, and the user id stays in the WAL file

`monitor/backend/src/main/kotlin/octometer/monitor/erasure/UserErasureService.kt:70`

```kotlin
writer.createStatement().use { it.execute("PRAGMA wal_checkpoint(TRUNCATE)") }
```

`PRAGMA wal_checkpoint` never throws. It returns one row with three columns: `busy`, `log`, and
`checkpointed`. The code drops that row. A `busy` value of 1 means the checkpoint did not
reset the WAL file, and the erased bytes stay on the disk. The route still answers 200.

My evidence. I opened a second connection on the same file, I started a read transaction on it,
and then I ran the erasure with a rare user id:

```
deleted=2
PRAGMA wal_checkpoint(TRUNCATE) -> busy=1 log=27 checkpointed=23
octometer.db-wal: length 111272 bytes, the user id at offset 82408
octometer.db:     the user id absent
```

The table is clean, the answer says 200, and the personal data is still in
`octometer.db-wal`. A `PRAGMA secure_delete=ON` does not help here, because the WAL file keeps
the old frames.

A second reader is not a rare case:
- Issue #18 (pull request #127) makes `SqliteDatabase.read { }` one read transaction. Each
  parallel API read then holds a snapshot.
- Issue #59 (the purge) and a later backup job both read the store.
- A poll cycle reads the store while the owner runs the erasure.

Without a second reader the checkpoint works. I ran the same flow with the monitor reader only
(the `appExists` read of the route, then the erasure) and the WAL file went to 0 bytes.

**The fix.** Read the result row of the pragma. Retry while `busy` is 1, and report a failure.

```kotlin
private fun checkpoint(writer: Connection): Int =
    writer.createStatement().use { statement ->
        statement.executeQuery("PRAGMA wal_checkpoint(TRUNCATE)").use { result ->
            check(result.next()) { "PRAGMA wal_checkpoint gave no row." }
            result.getInt(1)
        }
    }
```

Call it a maximum of five times, with a short pause between two calls. When `busy` is still 1
after the last try, write one fixed log line with no user data, for example "The WAL checkpoint
after an erasure did not complete.", and give the answer a field that states the fact, or answer
503. The owner must never read 200 when a byte of the user stays on the disk. Put the helper in
the store package, because issue #59 (the purge) needs the same check under D15.

Add a test: hold a read transaction on a second connection, run the erasure, and assert that the
call reports the failure.

---

## MAJOR 1. The runbook drops one step of the erasure order of D15

`monitor/backend/README.md:15` and
`monitor/backend/src/main/kotlin/octometer/monitor/erasure/UserErasureService.kt:6-11`

Design decision D15 states: "The erasure order is: first the app, then one poll cycle, then the
monitor."

The README states two steps only: "Run the erasure of #35 in the app first. Else a later poll
cycle of the monitor can read the erased events again from the app store, and the monitor row
comes back." The code comment says the same. Neither text tells the owner to wait for one poll
cycle.

The consequence: the owner erases in the app, then runs this route at once. A poll cycle that
started before the app erasure already read those documents. It writes them into the monitor
after the erasure. The personal data comes back, and no line of the answer or of the log says so.

**The fix.** Write the three steps of D15 in the README as a vertical list:

1. Run the erasure of #35 in the app.
2. Wait for one full poll cycle of the monitor (the refresh time of the mode).
3. Run `DELETE /api/apps/{appId}/events?userId=<id>`.

Add the same three steps to the comment in `UserErasureService.kt`. Name D15 as the source.

---

## MAJOR 2. The README holds no warning about the user id in the URL

`monitor/backend/README.md`

D13 puts the user id in the query string of the route. The pull request text names this risk.
The README does not. The README is the text that the owner reads.

What I checked in the code: the monitor installs no `CallLogging` plugin today, thus the process
writes no access log. The `StatusPages` handler logs the class name of an exception only. Thus
no log line holds the user id today, and my break 5 proves that the test guards it.

The risk stays outside the process:
- Issue #64 gives the owner a `curl` command. The user id then lands in the shell history and in
  the argument list of the process.
- A later `CallLogging` install, or a reverse proxy, records the request line with the query
  string.
- The browser history of the owner keeps the URL when the UI of a later issue calls the route.

**The fix.** Add a short section to `monitor/backend/README.md`:

- The user id travels in the query string of the URL.
- The monitor writes no access log today. Do not add one that records a query string (#31).
- Clear the shell history after the call, or read the user id from a shell variable.
- Never put this URL behind a proxy that logs a request line.

---

## MAJOR 3. The delete of the user rows covers only `kind = 0` and `kind = 1`

`monitor/backend/src/main/kotlin/octometer/monitor/erasure/UserErasureService.kt:27-30`

The two delete statements each hold a literal `kind` value. The schema
(`db/001_init.sql`, `db/002_first_page.sql`) has no `CHECK` on `kind`. A row with any other
`kind` value and that `user_id` stays in the table, and the call answers 200.

My evidence. I inserted three rows for one user id: `kind = 0`, `kind = 2`, and an anonymous row
with `kind = 2`. Then I ran the erasure:

```
deleted=2
rows after: [k2 sess-k-1 <the target user id> k2]
```

The row of the user with `kind = 2` stays. The anonymous row with `kind = 2` is gone, because
the anonymous statement carries no `kind` literal. An erasure must not depend on a value range
that the schema does not hold. The failure is silent: no error, and the count looks correct.

I accept the reason for the two literals: design rule M4 and the two partial indexes. The fix
keeps that reason.

**The fix.** Add one more statement after the two kind deletes, with no `INDEXED BY` and no
`kind` filter, and add its count to `deletedUserRows`:

```sql
DELETE FROM event WHERE app_id = ? AND user_id = ?
```

It deletes 0 rows today. It makes the erasure complete for each later `kind` value. Add a test
with a row of `kind = 2` that asserts the count and the empty table. A migration with
`CHECK (kind IN (0, 1))` is an acceptable alternative, but then the session read needs the same
guarantee.

---

## MINOR findings

1. `monitor/backend/src/test/kotlin/octometer/monitor/erasure/UserErasureServiceTest.kt:21`
   makes the temporary folder with `Files.createTempDirectory` and deletes it in `tearDown`
   only. When `setUp` throws, `tearDown` throws on the uninitialised `database` field, and a
   store file with test data stays in the temp folder of the machine. That is the defect of
   issue #142. Fix: add `.also { registerTempRoot(it) }`, as `testDataDir()` does.
2. `UserErasureRoutes.kt:33`. A user id of 10 000 characters gives 200 with `deleted: 0`.
   Contract rule C6 states that a `userId` string has 1 to 254 characters. Fix: reject a value
   above 254 characters with 400 and the fixed sentence of the route.
3. `UserErasureRoutesTest.kt:151-157`. The log test reads `event.formattedMessage` only, and it
   covers the 200 path only. Fix: check `event.throwableProxy?.message` too, and run the same
   check on the 400 path and on the 404 path.
4. `UserErasureService.kt:15-17`, `:20-22`, `:45-47`, and `:51-54`. Four comment sentences hold
   29, 26, 31, and 33 words. The brief sets a maximum of 25 words. Fix: split each one.
5. `UserErasureServiceTest.kt:126`. The checkpoint test accepts an absent WAL file
   (`!walFile.exists() || walFile.length() == 0L`). Fix: assert the `busy` value of the pragma
   after the fix of BLOCKER 1, because that value is the real signal.

---

## The byte check (my check 4)

I ran the erasure with a rare user id and a rare session id, and I read each file
`octometer.db*` while the store stayed open:

```
octometer.db     53248 bytes  user id: absent  session id: absent
octometer.db-wal     0 bytes  user id: absent  session id: absent
octometer.db-shm 32768 bytes  user id: absent  session id: absent
```

The free pages of the main file hold no byte of the user id. `PRAGMA secure_delete=ON` of
`SqlitePragmas.kt:19` writes zeros over each freed cell, thus no `VACUUM` is needed. The session
id of a session that the erasure emptied is gone too. The session id of the shared session stays,
because the row of the second user keeps it; contract rule C43 asks for that.

This result holds only when the checkpoint completes. See BLOCKER 1 for the case where it does
not.

## Other checks

- The diff holds no secret, no database file, and no build output. It adds 7 files and deletes
  nothing.
- Only the table `event` holds a `user_id` column. The tables `skipped_event` and `gap` hold no
  user data. The pull request text states this, and I read both migration files to confirm it.
- The route is one line in `ApiRoutes.kt`, and `Application.kt` does not change. This follows
  the structure after #125.
- The answer body holds the count only. No 400 sentence, no 404 sentence, and no 500 sentence
  holds a user id.

I will approve when BLOCKER 1 and the three MAJOR findings are fixed.

xamcross:
**Maintainer decisions after the privacy review (2026-09-22).**

- BLOCKER 1 (the checkpoint result is dropped): the code reads the row of `PRAGMA wal_checkpoint(TRUNCATE)`. On `busy = 1` it tries again, a maximum of 5 times with 100 ms between two tries. When the last try is still busy, the answer stays 200 and holds `"checkpointed": false`, and the code writes one warning with the app id only. The README tells the owner to call the route again in that case. A test with a second reader proves both paths.
- MAJOR 1 (two steps in place of three): the README and the code comment state the three steps of design decision D15 in this order: the app erases its copy (#35), one poll cycle passes, then the owner calls this route.
- MAJOR 2 (the user id in the query string): the route shape stays as D13 states it. The README gets a warning: the user id travels in the URL, thus a shell history, a proxy log, and a later request log hold it. #64 (the curl example) and #57 (the file log) get a note from the maintainer.
- MAJOR 3 (a row with `kind = 2` survives): each delete of the user selects by `app_id` and `user_id` with no `kind` filter, thus a later event kind cannot survive. The cost is a scan inside one app; the erasure is a rare operation. The SQL review measures it. A test seeds a row with `kind = 2`.

The correction round starts after the SQL review.

xamcross:
**Review (SQL engineer): CHANGES REQUESTED**

I made my own clone, I ran `gh pr checkout 147`, and I ran each check myself on JDK 21
(`C:/Users/xamcr/.jdks/ms-21.0.9`). The code is clean, the transaction shape is correct, and
the test suite is strong. Two MAJOR findings stop the approval. Both touch personal data.

## The acceptance criteria

|---|---|---|---|
| 1 | The rows of that user id go, the rows of each other user id stay | PASS | I removed `AND user_id IS NULL` from the anonymous delete. `UserErasureServiceTest > the erasure removes the rows of the user and keeps a row of a second user in the same session` failed, and `UserErasureRoutesTest > DELETE removes the rows of the user and gives 200 with the deleted count` failed. |
| 2 | A row with `user_id IS NULL` of a session of that user goes | PASS | I replaced the anonymous delete with a call on an empty session set. Three tests failed, among them `the erasure removes an anonymous row of a session that also holds a row of the user`. |
| 3 | A row of a second user in the same session stays | PASS | The same break as criterion 1. |
| 4 | A row with `user_id IS NULL` of another session stays | PASS | I widened the anonymous delete to `(session_id = ? OR 1 = 1)`. Two tests failed. |
| 5 | The field `deleted` counts the two row groups | PASS | I changed the answer to `result.deletedUserRows`. `UserErasureRoutesTest > DELETE also removes the anonymous rows of a session of the user` failed. |
| 6 | `PRAGMA wal_checkpoint(TRUNCATE)` runs after the delete | PASS with MAJOR 1 | I deleted the checkpoint line. `UserErasureServiceTest > the erasure runs a WAL checkpoint after the commit` failed, and `UserErasureRoutesTest > a byte search of the store files finds no user id after the erasure` failed too. The statement runs, but see MAJOR 1: the code drops its answer. |
| 7 | The user id is a bound parameter, and no log line holds it | PASS | Each of the five statements binds the user id with `setString`. The route test reads each captured event. See MINOR 4 for the level limit of the capture. |
| 8 | The captured log holds no query string | PASS | `UserErasureRoutesTest:153` asserts that no line holds `userId=`. |
| 9 | 200 with `deleted`, 404 for an unknown app id | PASS | I disabled the 404 branch of `UserErasureRoutes.kt:39`. `DELETE with an unknown app id gives 404` failed. |
| 10 | `./gradlew :monitor:backend:test` passes | PASS | Two runs in sequence, the second one after `cleanTest`. `BUILD SUCCESSFUL` each time. 19 test classes, 187 tests, 0 failures, 0 errors, from `monitor/backend/build/test-results/test/TEST-*.xml`. |
| 11 | A comment on the issue holds the test output | PASS | The comment is on #61. |

The diff holds 7 files, 741 added lines, and 0 deleted lines. It holds no secret, no database
file, and no build output.

## MAJOR 1 — the code drops the answer of the WAL checkpoint

`monitor/backend/src/main/kotlin/octometer/monitor/erasure/UserErasureService.kt:70`

`PRAGMA wal_checkpoint(TRUNCATE)` never throws on a conflict. It answers one row with three
columns: `busy`, `log`, `checkpointed`. A value of 1 in the first column means that the
checkpoint did nothing. The code calls `execute(...)` and reads no column, so the route
answers 200 and the operator learns nothing.

My evidence. I added a probe test to my clone. It opened a second connection, it ran `BEGIN`
plus one `SELECT`, and it held that read snapshot. Then it ran the erasure on the same store.
The result:

```
PROBE| PRAGMA wal_checkpoint(TRUNCATE) answer (busy, log, checkpointed) = [1, 165, 0]
PROBE| ms of that checkpoint call = 5544
PROBE| WAL bytes after that checkpoint = 679832
PROBE| byte search of octometer.db (1048576 bytes): found = true
```

The last line is the defect. The delete writes the new, zeroed pages into the WAL file. The
old pages, with the user id in clear bytes, stay in `octometer.db` until a checkpoint moves
the WAL over them. A busy checkpoint leaves the erased user id in the main file. Design
decision D15 asks for the checkpoint for exactly this reason.

The case is reachable today. Pull request #127 merged into `main` on commit 6154083. Its
`SqliteDatabase.read { }` now runs `BEGIN` before the block and `COMMIT` after it, so every
read of the user interface holds a read snapshot on the reader connection.

A read does not block the checkpoint for ever. `PRAGMA busy_timeout=5000` bounds the wait. My
measurement was 5544 ms, and then the checkpoint gave up with `busy = 1`. After the reader
closed, the next call answered `[0, 0, 0]` and the WAL went to 0 bytes.

The fix. Read the answer row, repeat the call a small number of times, and write one fixed
sentence when it stays busy:

```kotlin
private fun checkpointAfterCommit(writer: Connection): Boolean {
    repeat(CHECKPOINT_ATTEMPTS) {
        writer.createStatement().use { statement ->
            statement.executeQuery("PRAGMA wal_checkpoint(TRUNCATE)").use { answer ->
                if (answer.next() && answer.getInt(1) == 0) return true
            }
        }
    }
    return false
}
```

Call it after `COMMIT`. When it gives `false`, write one ERROR line with a fixed text and no
user data, for example "A WAL checkpoint after an erasure did not run. The store keeps the old
pages." Keep the answer at 200, because the rows are gone. Add a test that holds a read
snapshot on a second connection, and that asserts the ERROR line.

## MAJOR 2 — a row of the user with a third kind value survives

`monitor/backend/src/main/kotlin/octometer/monitor/erasure/UserErasureService.kt:27-30`,
`:109-113`

The erasure deletes only `kind = 0` and `kind = 1`. The column is
`kind INTEGER NOT NULL DEFAULT 0` in `db/002_first_page.sql:9`, and it carries no `CHECK`
constraint. A row with any other value stays, and the route still answers 200 with the count
of the rows that it did delete. The same gap hits the session list, so the anonymous rows of
that session also stay.

My evidence. A probe test inserted one row with `user_id = <target>` and `kind = 2`, then it
called `eraseUserEvents`:

```
PROBE| deleted total = 0
PROBE| rows of the user that stay = 1
```

Design section 6 names two kinds today, so no such row exists on `main` now. A later design
decision that adds a kind turns this into a silent leak of personal data, with no failing
test. The erasure must fail closed, not open.

The fix, and its cost. Add one guard statement after the two kind deletes, inside the same
transaction, and add its count to `deletedUserRows`:

```kotlin
internal const val SQL_DELETE_USER_OTHER_KIND =
    "DELETE FROM event INDEXED BY event_session WHERE app_id = ? AND user_id = ? AND kind NOT IN (0, 1)"
```

Also read the session list of those rows before the delete, with the same index. I measured
the cost on an app with 110 000 rows (50 000 clicks of the target user, 5 000 session starts,
5 000 anonymous rows, 50 000 rows of 500 other users):

```
PROBE2| guard plan = SEARCH event USING COVERING INDEX event_session (app_id=?)
PROBE2| guard delete with 110000 rows present: 0 rows, 8 ms
PROBE2| the two kind deletes: 187 ms
PROBE2| guard delete after the two kind deletes: 0 rows, 3 ms
```

The guard costs 8 ms against 187 ms for the two fast deletes. It is an index-only scan of the
rows of one app. The price is small. Add a test with a row of `kind = 2`.

## Check of the `INDEXED BY` decision — keep it on the two SELECT statements

I ran `EXPLAIN QUERY PLAN` on each of the five statements, with the clause and without it, on
a table with 110 000 rows, before `ANALYZE` and after it. I also ran the same set on a table
with 10 rows.

The two `SELECT DISTINCT session_id` statements need the clause:

```
with INDEXED BY:
  SEARCH event USING COVERING INDEX event_agg (app_id=? AND user_id=?) | USE TEMP B-TREE FOR DISTINCT
  SEARCH event USING COVERING INDEX event_start (app_id=? AND user_id=?) | USE TEMP B-TREE FOR DISTINCT
without INDEXED BY (10 rows, 110 000 rows, and 110 000 rows after ANALYZE):
  SEARCH event USING COVERING INDEX event_session (app_id=?)
  SEARCH event USING COVERING INDEX event_session (app_id=?)
```

Without the clause the planner drops the `user_id` seek and scans every row of the app. The
reason is the `DISTINCT`: `event_session` starts with `(app_id, session_id)`, so it gives the
distinct order free, and the planner pays a full scan to save the temp B-tree. `ANALYZE`
changes nothing. For a user with few events in a large app that choice is much worse.

The three `DELETE` statements do not need the clause. The plan is the same with it and
without it, on 10 rows, on 110 000 rows, and after `ANALYZE`:

```
SEARCH event USING COVERING INDEX event_agg (app_id=? AND user_id=?)
SEARCH event USING COVERING INDEX event_start (app_id=? AND user_id=?)
SEARCH event USING COVERING INDEX event_session (app_id=? AND session_id=?)
```

My recommendation: keep `INDEXED BY` on each of the five statements. The reason is the two
SELECT statements. One rule for the whole file is easier to hold than a mixed rule. The cost
that you name is real: a later migration that drops or renames `event_agg`, `event_start`, or
`event_session` gives `SQLException: no such index` at prepare time, and the route then
answers 500. Please add one test in `Migration002Test` that asserts the three index names in
`sqlite_master`. That test turns a schema change into a build failure, not a run-time failure.

## Check of the cost — a batch form is not needed at this size

I built an app with 50 000 click rows of one user over 5 000 sessions, plus 5 000 session
starts of that user, plus 5 000 anonymous rows of those sessions, plus 50 000 rows of other
users. Page size 4096, 6159 pages. One erasure, measured inside the `write { }` block:

```
ms read session list = 23
ms delete user rows = 261            (55 000 rows)
ms delete anonymous rows = 50        (5 000 statements, 5 000 rows)
ms COMMIT = 195
ms checkpoint TRUNCATE = 3
ms total of the write block = 538
WAL bytes after COMMIT = 15594232    (14.9 MiB, about 3800 frames)
WAL bytes after the checkpoint = 0
```

The one writer thread stays busy for 538 ms. The poll loop waits that time. That is
acceptable. The loop of 5 000 single statements is not the cost: I also built the same delete
with one temp table and one `IN (SELECT ...)`, and it took 43 ms against 50 ms. Keep the
loop. It is simpler.

My answer: no batch form is needed for 50 000 events. The growth is linear, so a user with
1 000 000 events gives about 10 s of writer time and about 300 MiB of WAL. See MINOR 6.

## The schema — only `event` holds a user id

I read every table of both migrations. `app` holds 13 columns and no user id.
`skipped_event` holds `app_id`, `event_id`, `reason`. `gap` holds `app_id`, `from_ts`,
`to_ts`. Only `event` holds `user_id`. The claim of the pull request is correct.

## The transaction, the route, and the merges

- The read of the session list and the two deletes run in one `BEGIN IMMEDIATE` block inside
  one `database.write { }` call. The two counts come from the `executeUpdate` answers of that
  same transaction. The checkpoint runs after `COMMIT`, and still on the writer thread, so no
  other write comes between the two. This is correct. I moved the session read after the
  delete of the user rows, and three tests failed, so rule E2 has a test.
- The route follows the pattern of `AppRegistryRoutes.kt`. It adds one import and one line to
  `ApiRoutes.kt`. `RequestGuard` gives it the Host check, the Origin check, and
  `Cache-Control: no-store`. `RequestGuardTest.kt:431` asserts the three headers on an `/api`
  path, and `RequestGuardTest.kt:420` asserts that no answer holds
  `Access-Control-Allow-Origin`. The guard is one interceptor, so the new route inherits each
  check. The erasure tests assert no header themselves, and that is acceptable.
- Merges (fact, not a finding). `origin/main` (6154083), `pull/127/head`, `pull/144/head`, and
  `pull/145/head` each conflict in
  `monitor/backend/src/main/kotlin/octometer/monitor/ApiRoutes.kt`, on the import line and on
  the route line. No other file conflicts. `LogCapture.kt` conflicts with none of the four.
  I aborted each merge.

## MINOR findings

1. `monitor/backend/src/test/kotlin/octometer/monitor/erasure/UserErasureServiceTest.kt:21` —
   the temp folder calls no `registerTempRoot`. `tearDown` calls `database.close()` first, and
   `database` is a `lateinit` field, so a throw in `setUp` skips the delete and leaves the
   folder. Issue #142 exists for this. Fix:
   `Files.createTempDirectory("octometer-user-erasure-test-").toFile().also { registerTempRoot(it) }`.
2. `UserErasureServiceTest.kt:142-151` — the test `the delete statements use an index` cannot
   fail from a change of the statements. I removed every `INDEXED BY` clause with a regular
   expression and ran the erasure tests. Only `the session-id query uses the partial index of
   its kind` failed. The three DELETE plans name the same index without the clause. Fix:
   either delete the test, or assert the exact text `SEARCH event USING COVERING INDEX
   event_agg (app_id=? AND user_id=?)` for each plan.
3. `UserErasureService.kt:17-21` — the comment gives a wrong reason: "A small test table gives
   SQLite no real statistics, and the planner can then pick a different, valid index for a
   tiny table." The planner picks `event_session` on 110 000 rows too, and `ANALYZE` changes
   nothing. Fix: name the real reason. `DISTINCT session_id` makes `event_session` attractive,
   because that index gives the distinct order with no temp B-tree.
4. `monitor/backend/src/test/kotlin/octometer/monitor/LogCapture.kt:31-48` — the new helper is
   correct, and I accept it, with two points. First, the root logger level is `INFO`
   (`monitor/backend/src/main/resources/logback.xml:27`), so logback never builds a DEBUG
   event. A later DEBUG line with the user id passes the test. Fix: set the root level to
   `Level.TRACE` inside the capture, and restore the old level in the `finally` block.
   Second, `captureErrorLogEvents` is now a copy of `captureLogEvents` plus one filter. Fix:
   make the first one call the second one.
5. `monitor/backend/README.md` — the file says nothing about a later migration. A new column
   or a new table with a user id needs a change in `UserErasureService.kt` on the same day.
   Fix: add one line, for example "Only the table `event` holds a `user_id` column. A
   migration that adds a user id to another table must also extend
   `octometer.monitor.erasure.UserErasureService`, and add a test."
6. `UserErasureService.kt:70` — the erasure has no chunk limit. My measurement gives 14.9 MiB
   of WAL for 60 000 deleted rows in one transaction. Design decision D13 deletes the events
   of an app in chunks of 10 000, and D15 purges in chunks of 10 000. Fix: add one line to the
   README with the measured numbers, or use the same chunk size here. A chunk needs its own
   transaction for each step, so the erasure is no longer atomic; please choose with the
   maintainer.
7. `UserErasureService.kt:70` — the checkpoint holds the one writer thread for the whole wait.
   My busy case held it for 5544 ms, and the poll loop waited that time. Fix: run
   `PRAGMA wal_checkpoint(PASSIVE)` first. It never waits. Run the TRUNCATE form only after
   the PASSIVE form reports work that stays.

## Summary

- BLOCKER: 0
- MAJOR: 2 (the dropped checkpoint answer, the row of a third kind)
- MINOR: 7

Fix MAJOR 1 and MAJOR 2, and I approve.

xamcross:
Correction round 1, after the privacy review and the SQL review.

## BLOCKER 1 (privacy): the checkpoint result was dropped

`UserErasureService.kt:70` (before this round). The code now reads the answer row of `PRAGMA
wal_checkpoint(TRUNCATE)`. On `busy = 1` it retries, a maximum of 5 times with a 100 ms pause.
A busy last try keeps the answer at 200 with `"checkpointed": false`, and writes one warning
with the app id only. On success the answer holds `"checkpointed": true`.

Evidence: `UserErasureServiceTest.a busy checkpoint retries, then reports checkpointed false
with one warning` (a second reader holds a read transaction through `database.read`), and
`UserErasureServiceTest.a checkpoint with no other reader completes and reports checkpointed
true`. The old code was restored for a moment. The busy test then failed on
`result.checkpointed`. The fix was restored, and the suite passed again.

## MAJOR 1 (privacy): two steps of D15 in place of three

`monitor/backend/README.md` and `UserErasureService.kt:6-18`. The text now states the three
steps in order: the app erasure of #35, one poll cycle, then this route.

## MAJOR 2 (privacy): no warning about the user id in the URL

`monitor/backend/README.md`. A new section warns that the user id travels in the query
string. It names the shell history, a proxy log, and a future request log as the places that
can hold it.

## MAJOR 3 (privacy) and MAJOR 2 (SQL): a row with `kind = 2` survives

`UserErasureService.kt:60-61` and `:183`. A third delete, `SQL_DELETE_USER_OTHER_KIND`,
removes a row of the user with a kind other than 0 or 1. It keeps the two existing
`INDEXED BY` deletes.

`EXPLAIN QUERY PLAN` of the new statement: `SEARCH event USING COVERING INDEX event_session
(app_id=?)`. This matches the 8 ms the SQL review measured on 110 000 rows.

Evidence: `UserErasureServiceTest.the erasure removes a row of a kind other than 0 or 1, and
counts it`. The guard delete was removed for a moment. The test then failed on the count and
on the remaining row. The fix was restored.

## MAJOR 1 (SQL): keep `INDEXED BY`

Unchanged, per the recommendation. A new test,
`UserErasureServiceTest.the three indexes of the erasure exist in sqlite_master`, fails the
build when a schema change drops or renames `event_agg`, `event_start`, or `event_session`.

## MINOR findings: 9 corrected, 1 not corrected

1. `UserErasureServiceTest.kt:21`: the temp folder now calls `registerTempRoot` (issue #142).
2. `UserErasureRoutes.kt:46-49`: a `userId` over 254 characters (contract rule C6) gives 400
   with a fixed sentence.
3. `UserErasureRoutesTest.kt`: the log sentinel test now reads
   `event.throwableProxy?.message` too, and it runs on the 200, the 400, and the 404 path.
4. `UserErasureService.kt`: each comment sentence now has at most 25 words.
5. `UserErasureServiceTest.kt` and `UserErasureRoutesTest.kt`: the checkpoint tests now assert
   `result.checkpointed`, the WAL file size, and the byte search, not the call alone.
6. `UserErasureServiceTest.kt`: `the delete statements use an index` now asserts the exact
   `EXPLAIN QUERY PLAN` text of each statement.
7. `LogCapture.kt`: `captureErrorLogEvents` now calls `captureLogEvents` with a level filter.
8. `LogCapture.kt`: `captureLogEvents` raises the root log level to TRACE for its block, and
   restores the old level in `finally`.
9. `monitor/backend/README.md`: a new line states that a migration which adds a user id to a
   new table must extend `UserErasureService`, and add a test.

Not corrected: SQL review MINOR 7 (`PRAGMA wal_checkpoint(PASSIVE)` before the `TRUNCATE`
form). The maintainer's decision for BLOCKER 1 states the retry directly on the `TRUNCATE`
form, and a `PASSIVE` pre-step would change that decision. Left for the maintainer to confirm.

## A finding outside the scope of #61

The TRACE-level log capture (MINOR 4 of the SQL review) surfaced a framework-level DEBUG line
with no tie to the erasure route. `io.ktor.server.plugins.statuspages.StatusPages` logs
"No handler found for status code ... for call: <url>" for a normal answer with no registered
status handler, and `<url>` is the full request URL, including any query string. This never
runs today, because the root log level is INFO. A later switch to DEBUG logging would write
the user id of every request that carries one, on every route, not only this one.
`LogCapture.kt` filters this logger name out of the sentinel check for now. This is a
design-level, cross-route finding; please file it as a new issue.

## Test result

`./gradlew :monitor:backend:test`, JDK 21 (`C:/Users/xamcr/.jdks/ms-21.0.9`), two runs in
sequence after the merge of `origin/main` (commit 7576e7a): `BUILD SUCCESSFUL` each time, 231
tests, 0 failures, 0 errors (22 test classes, from
`monitor/backend/build/test-results/test/TEST-*.xml`). `./gradlew build` also passed once, for
the whole project.

## CI result

`gh pr checks 147`: Frontend, JVM modules, Monitor backend on Windows, Secret scan, and
Tracker all `pass`.

Commits: `0d73fb1` (the correction round), `1198289` (the merge of `origin/main`, commit
7576e7a, resolving the `ApiRoutes.kt` conflict and a duplicate `captureLogEvents` from the
automatic merge of `LogCapture.kt`).

xamcross:
**Review (second review, privacy engineer): APPROVE**

I made a new clone, I ran `gh pr checkout 147`, and I checked each item with JDK 21
(`C:/Users/xamcr/.jdks/ms-21.0.9`) at commit `1198289`. I ran `./gradlew :monitor:backend:test`
one time: `BUILD SUCCESSFUL`, 22 test classes, **231 tests**, 0 failures. That count agrees with
the correction comment. I deleted my clone after the review. My first review is comment
5769624408.

My BLOCKER and my three MAJOR findings are fixed. Each of my five MINOR findings is fixed. I
add three new MINOR findings. No BLOCKER and no MAJOR stays, thus I approve.

---

## BLOCKER 1. The checkpoint result: FIXED

`UserErasureService.kt:137-151` reads the `busy` column of `PRAGMA wal_checkpoint(TRUNCATE)`.
It retries a maximum of 5 times, with a pause of 100 ms. `UserErasureResult.checkpointed`
carries the result to the answer.

**The busy path, through the route.** I held a read transaction open on a second connection to
the same file, then I called the route:

```
status=200 body={"deleted":2,"checkpointed":false}
warn count=1
warn logger=octometer.monitor.erasure.UserErasureService
     msg=The WAL checkpoint after the erasure of app 1 did not complete.
     holdsUser=false
```

Exactly one warning. It names the app id only. It holds no user id and no session id. This
matches the maintainer decision.

**The success path, through the route.** With no second reader:

```
status=200 body={"deleted":3,"checkpointed":true}
octometer.db     53248 bytes  user id: absent  session id: absent
octometer.db-wal     0 bytes  user id: absent  session id: absent
octometer.db-shm 32768 bytes  user id: absent  session id: absent
```

**The old code.** I put the round-1 code back for a moment (run the pragma, read no row, always
report success). The test `UserErasureServiceTest.a busy checkpoint retries, then reports
checkpointed false with one warning` FAILED. I restored the fix.

**The remedy of the README.** I also checked the advice that the README gives. After the busy
call, I let the second reader leave, then I called the route again:

```
first  = {"deleted":2,"checkpointed":false}
second = {"deleted":0,"checkpointed":true}
octometer.db, octometer.db-wal, octometer.db-shm: the user id absent in each file
```

The remedy works. A second call deletes 0 rows, completes the checkpoint, and clears the
bytes.

`SqliteDatabase.read` now opens one read transaction (the merge of #127). Thus each parallel
API read holds a snapshot, and the busy case is a live case, not a theoretical one. The fix
arrives at the right time.

## MAJOR 1. The three steps of D15: FIXED

`monitor/backend/README.md:17-26` holds the section "The erasure order (design decision D15)"
with the three steps in a vertical list: the app erasure of #35, one full poll cycle, then this
route. `UserErasureService.kt:7-18` holds the same three steps as a comment. Both name the
consequence of a call before step 2 finishes.

## MAJOR 2. The user id in the URL: FIXED

`monitor/backend/README.md:39-49` holds the section "The user id in the URL". It names the
shell history, a proxy log, and a future request log. It tells the owner not to add a log that
records a query string, and it names #31 and #57. I confirmed in the code that the monitor
installs no `CallLogging` plugin today.

## MAJOR 3. A row with a kind other than 0 or 1: FIXED

`UserErasureService.kt:60-61` adds `SQL_DELETE_USER_OTHER_KIND` on the index `event_session`,
and `:183` calls it. My probe seeded one row of the user with `kind = 0`, one with `kind = 2`,
one with `kind = 7`, and one anonymous row with `kind = 2`:

```
deleted=4 checkpointed=true
rows after: []
```

Each row is gone, and the count holds each one. I removed the guard delete for a moment: the
test `UserErasureServiceTest.the erasure removes a row of a kind other than 0 or 1, and counts
it` FAILED. I restored the code.

## The five MINOR findings of the first review

|---|---|---|---|
| 1 | The temp folder of `UserErasureServiceTest` is not registered | FIXED | `UserErasureServiceTest.kt:29` calls `registerTempRoot` |
| 2 | A user id above 254 characters gives 200 | FIXED | `UserErasureRoutes.kt:13,46-49`, `USER_ID_MAX_LENGTH = 254`, 400 with a fixed sentence; the test at `UserErasureRoutesTest.kt:55` |
| 3 | The log test reads `formattedMessage` only, on the 200 path only | FIXED | `UserErasureRoutesTest.kt:148,161,173` cover 200, 400, and 404; `assertNoLogLineHoldsMarker` also reads `event.throwableProxy?.message` |
| 4 | Four comment sentences above 25 words | FIXED | I counted each sentence of `UserErasureService.kt`. The longest holds 20 words |
| 5 | The checkpoint test accepts an absent WAL file | FIXED | `UserErasureServiceTest.kt:148` asserts `result.checkpointed`; `UserErasureRoutesTest.kt:221` asserts the field, the WAL size, and the byte search |

---

## The StatusPages line: CONFIRMED, at TRACE level, not DEBUG

I ran my own probe: a 404 on this route with a user id in the URL, and my own appender on the
root logger at TRACE level.

```
io.ktor.server.plugins.statuspages.StatusPages  level=TRACE
  No handler found for status code 404 Not Found for call:
  /api/apps/777777/events?userId=<my marker>
```

The line holds the full request target, with the query string and the user id. I confirm the
finding. I correct one detail: the level is **TRACE**, not DEBUG. A switch of the root level to
DEBUG does not print this line. Only a switch to TRACE prints it.

**The default mode prints neither.** `monitor/backend/src/main/resources/logback.xml` sets
`<root level="INFO">`. Thus no DEBUG line and no TRACE line reaches the console appender or the
error appender today.

Three more TRACE lines of my probe held the same URL: `io.ktor.client.plugins.SaveBody`,
`io.ktor.client.plugins.HttpCallValidator`, and
`io.ktor.client.plugins.DefaultResponseValidation`. Each one comes from the test HTTP client,
not from the server. `LogCapture.kt:34-37` filters both logger families, and that is correct.

I recommend the new issue. The scope: each route with a value in its path or its query string,
not this route alone. The fix in that issue: register a status handler for the statuses that
the API returns, or set the level of `io.ktor.server.plugins.statuspages` to INFO in
`logback.xml`, and add a sentinel test at TRACE level.

## New MINOR findings

1. `docs/superpowers/specs/2026-09-21-octometer-design.md:318`. D13 states "The erasure returns
   `deleted`." The answer now holds `deleted` and `checkpointed`. Fix: add `checkpointed` to the
   field list of D13, with one sentence for its meaning.
2. `monitor/backend/README.md:35-36`. The text states "an old page can stay in the WAL file for
   a short time". My probe of the busy path shows the opposite place: the WAL file held no
   marker (16512 bytes, absent), and the **main file** `octometer.db` held the user id at offset
   20455. The delete writes the new, zeroed page into the WAL, and the busy checkpoint leaves
   the old page in the main file. An owner who reads this sentence can decide to delete
   `octometer.db-wal` by hand, and that breaks the store. Fix: write "an old page with the
   erased bytes can stay in the store files until a later call completes the checkpoint. Never
   delete a file `octometer.db*` by hand."
3. `monitor/backend/src/test/kotlin/octometer/monitor/LogCapture.kt:34-37`. The sentinel filter
   skips each logger name that starts with `io.ktor.server.plugins.statuspages.StatusPages`. The
   filter is correct today, because the monitor never writes through that logger. Fix: remove
   the filter when the new issue lands, and name that issue in the comment.

## Other checks

- The correction commit `0d73fb1` touches 6 files only: the README, the two erasure source
  files, `LogCapture.kt`, and the two erasure test files. The other files of the branch diff
  come from the two merges of `origin/main` (#127 and #144).
- The route still sends `Cache-Control: no-store` and no CORS header, and it keeps the Host
  check and the Origin check of D12. My round-1 probe covered this, and `ApiRoutes.kt` still
  registers the route in one line.
- The answer body holds the two counts only. No 400 sentence and no 404 sentence holds a user
  id.
- One observation, not a finding: `checkpointTruncate` calls `Thread.sleep` on the one writer
  thread, thus a busy case blocks each other write for a maximum of 400 ms. The erasure is a
  rare operation, and the maintainer decision names the retry, thus I accept it.

The work is correct. I approve.

xamcross:
**Review (second review, SQL engineer): CHANGES REQUESTED**

I made a new clone, I ran `gh pr checkout 147` (commit 1198289), and I ran each check myself on
JDK 21. Both MAJOR findings of my first review are fixed. The correction of MAJOR 1 adds one
new MAJOR: the retry loop holds the one writer thread for 28 seconds, and it reports a false
alarm in the common case. One small fix closes it.

## My MAJOR 1 — the checkpoint result: FIXED, with a new MAJOR beside it

`UserErasureService.kt:137-152`. The code now reads column 1 (`busy`), retries a maximum of 5
times with a pause of 100 ms, and answers `checkpointed`. My probe with one open read snapshot
on the reader connection:

```
R2| busy case: checkpointed=false total=1
R2| busy case: the write block held the writer thread for 28291 ms
```

The result value is correct. The 28 291 ms is the new defect. See MAJOR A.

## My MAJOR 2 — a row with `kind = 2`: FIXED

`UserErasureService.kt:60` and `:183`. `SQL_DELETE_USER_OTHER_KIND` runs after the two kind
deletes, and `deleteUserRows` adds its count. My measurement on 110 000 rows:

```
R2| guard plan = SEARCH event USING COVERING INDEX event_session (app_id=?)
R2| guard delete = 26 ms, the two kind deletes = 452 ms
```

The plan matches the statement text, and the price is 6 percent of the two fast deletes. The
test `the erasure removes a row of a kind other than 0 or 1, and counts it`
(`UserErasureServiceTest.kt:130`) covers it. See MINOR 8 for the one part that stays open.

## The `sqlite_master` test: CONFIRMED

`UserErasureServiceTest.kt:230-248` selects `event_agg`, `event_start`, and `event_session`
from `sqlite_master`, and it asserts the three names. A migration that drops or renames one of
them now fails the build.

## MAJOR A (new) — the retry loop blocks the writer thread, and it gives a false alarm

`monitor/backend/src/main/kotlin/octometer/monitor/erasure/UserErasureService.kt:137-152`

Two facts, both measured in my clone.

**Fact 1. The block is 28 seconds.** `PRAGMA busy_timeout=5000` applies to each
`wal_checkpoint(TRUNCATE)` call. Five tries plus four pauses of 100 ms give 28 291 ms on my
machine. `SqliteDatabase.write { }` runs on the one writer thread, and `Thread.sleep` holds
that thread too. The poll loop of design decision D6 shares the thread, and one poll cycle runs
inside `withTimeout(45 s)`. One erasure with a busy reader can push a poll cycle past its
timeout. Before this round the block was 5.5 seconds.

**Fact 2. The alarm is false in the common case.** The byte at risk sits in `octometer.db`,
not in the WAL file. The delete writes the new, zeroed pages into the WAL. The old pages, with
the user id in clear bytes, stay in the main file until a checkpoint copies the WAL over them.
A `PASSIVE` checkpoint does that copy, and it never waits. Only the file-length reset needs
`TRUNCATE`. My probe, with a reader that holds the newest snapshot (this is what a user
interface poll gives):

```
R2| after COMMIT, no checkpoint: octometer.db holds the user id = true
R2| a reader now holds the newest snapshot
R2| PASSIVE answer (busy, log, checkpointed) = [0, 4, 4]
R2| after PASSIVE: octometer.db holds the user id = false
R2| TRUNCATE answer (busy, log, checkpointed) = [1, 4, 4], WAL bytes = 16512
```

`PASSIVE` copied every frame, and the user id left the main file. `TRUNCATE` still answered
`busy = 1`. The route then waits 28 seconds, writes a warning, and answers
`checkpointed: false`, although the privacy goal was already met. The README then tells the
operator to call the route again. That advice costs another 28 seconds and changes nothing.

**The fix.** Take the privacy answer from `PASSIVE`, and keep `TRUNCATE` for the file length
only. Read all three columns: the copy is complete when column 2 (`log`) equals column 3
(`checkpointed`).

```kotlin
private fun checkpointAfterCommit(writer: Connection): Boolean {
    var moved = false
    repeat(CHECKPOINT_MAX_ATTEMPTS) { attempt ->
        // PASSIVE never waits. It copies each frame that no reader needs.
        val passive = pragmaRow(writer, "PRAGMA wal_checkpoint(PASSIVE)")
        if (passive[0] == 0 && passive[1] == passive[2]) {
            moved = true
        }
        // TRUNCATE resets the file length. A reader can hold it back.
        if (pragmaRow(writer, "PRAGMA wal_checkpoint(TRUNCATE)")[0] == 0) return true
        if (moved) return true
        if (attempt < CHECKPOINT_MAX_ATTEMPTS - 1) Thread.sleep(CHECKPOINT_RETRY_DELAY_MILLIS)
    }
    return moved
}
```

Also bound the wait. Set `PRAGMA busy_timeout=0` before the checkpoint block, and restore
`PRAGMA busy_timeout=5000` in a `finally`. Five tries then cost about 400 ms, not 28 seconds.
Keep the maintainer decision: the `TRUNCATE` form still runs, and the try count stays at 5.

Add two tests: one that asserts the write block ends inside 2 seconds with a busy reader, and
one that asserts `checkpointed = true` for a reader on the newest snapshot.

## The 7 MINOR findings of my first review

|---|---|---|---|
| 1 | The temp folder calls no `registerTempRoot` | FIXED | `UserErasureServiceTest.kt:29` calls `.also { registerTempRoot(it) }`. |
| 2 | `the delete statements use an index` cannot fail | FIXED as asked | `UserErasureServiceTest.kt:214-224` asserts the exact plan text of each of the four statements. The test still cannot fail from a removed `INDEXED BY` on a DELETE, because the planner picks the same index; it now catches a change of the index shape, which is what I asked for. |
| 3 | The comment gives a wrong reason for `INDEXED BY` | FIXED | `UserErasureService.kt:38-43` names `DISTINCT session_id` and the temp B-tree. This matches my measurement. |
| 4 | The log capture hides a DEBUG line; the duplication | FIXED | `LogCapture.kt:48-67` raises the root level to `TRACE` and restores it; `captureErrorLogEvents` now calls `captureLogEvents`. See MINOR 9 for one new point. |
| 5 | The README says nothing about a later migration | FIXED | `monitor/backend/README.md`, the section "A migration that adds a user id to a new table". |
| 6 | The transaction has no chunk limit | OPEN | The README holds no measured number and no chunk decision. My first review measured 14.9 MiB of WAL for 60 000 rows in one transaction. Fix: add one line with the number, or take the chunk size of 10 000 that D13 and D15 use. |
| 7 | A `PASSIVE` checkpoint as a first step | RAISED to MAJOR A | My probe above shows that `PASSIVE` meets the privacy goal where `TRUNCATE` answers busy. The point is no longer a style improvement. |

## The framework DEBUG line: I confirm the report

I captured every event at `TRACE` with no filter, on one erasure call:

```
LOG| captured events = 48, events that hold the marker or userId= : 4
LOG| TRACE io.ktor.server.plugins.statuspages.StatusPages :: No handler found for status code 200 OK for call: /api/apps/1/events?userId=<marker>
LOG| TRACE io.ktor.client.plugins.SaveBody :: ...
LOG| TRACE io.ktor.client.plugins.HttpCallValidator :: ...
LOG| TRACE io.ktor.client.plugins.DefaultResponseValidation :: ...
```

The level is `TRACE`, and the root level of `logback.xml` is `INFO`, so no deployment writes
the line today. The three client lines come from the test HTTP client. The report is correct,
and the filter in `LogCapture.kt` is correct for now. The finding is real and it is
cross-route. Please file the new issue.

## MINOR findings of this round

8. `UserErasureService.kt:160-164` — `sessionIdsOfUser` still reads only `kind = 0` and
   `kind = 1`. A session that holds a row of the user with `kind = 2` and nothing else never
   enters the session list, so the anonymous rows of that session stay. Contract rule C43 then
   fails for a third kind. The test at `UserErasureServiceTest.kt:130` uses one session that
   also holds a `kind = 0` row, so it does not catch this. Fix: add
   `SELECT DISTINCT session_id FROM event INDEXED BY event_session WHERE app_id = ? AND user_id = ? AND kind NOT IN (0, 1)`,
   and add it to the set. The plan is the same index scan as the guard delete, which I
   measured at 26 ms on 110 000 rows. Add a test with a session that holds only a `kind = 2`
   row of the user plus one anonymous row.
9. `LogCapture.kt:34-37` — the list `NON_SERVER_LOGGER_PREFIXES` holds one server logger,
   `io.ktor.server.plugins.statuspages.StatusPages`. The name says the opposite of the content.
   Fix: rename it, for example `KNOWN_URL_LOGGER_PREFIXES`.
10. `monitor/backend/README.md`, the section "The WAL checkpoint" — the text puts the risky
    byte in the wrong file. It says "so no erased byte stays in the WAL file" and "an old page
    can stay in the WAL file for a short time". My probe shows the opposite: after the commit
    the WAL holds the new, zeroed pages, and `octometer.db` still holds the user id. Fix:
    state that the checkpoint copies the new pages over the old pages in `octometer.db`.
11. `UserErasureRoutes.kt:21` — the answer now holds a second field, `checkpointed`. Design
    decision D13 names only `deleted` for this route. Fix: add `checkpointed` to the field list
    of D13 in `docs/superpowers/specs/2026-09-21-octometer-design.md`, in the same round.

## The merge state

Fact, not a finding. `origin/main` moved to ba2ea5c (#145, the level 2 totals) after the last
push of this branch. I merged `origin/main` into the branch in my clone: the merge is
**clean**, with no conflict in any file. `ApiRoutes.kt` does not conflict on `userTotalsRoute`;
git adds the import and the line beside `appTotals`, `elementsRoute`, and
`userErasureRoutes`. I aborted the merge. The branch still needs that merge before the merge
into `main`.

`LogCapture.kt` holds exactly one `captureLogEvents` function and one `captureErrorLogEvents`
function. The duplicate of the automatic merge is gone.

## The test run

`./gradlew :monitor:backend:test`, one run, JDK 21: `BUILD SUCCESSFUL`. 22 test classes, 231
tests, 0 failures, 0 errors, from `monitor/backend/build/test-results/test/TEST-*.xml`. This
matches the count in your comment.

## Summary

- BLOCKER: 0
- MAJOR: 1 (MAJOR A, the 28-second writer block and the false `checkpointed: false`)
- MINOR: 5 open (my old number 6, plus the new 8, 9, 10, and 11)

Fix MAJOR A, and I approve.

xamcross:
## Correction round 2

I read the second SQL review (comment 5770066948) and the second privacy review (comment
5770065012, an approval with three MINOR notes). This comment lists each finding, its state,
and the file and line of the fix.

### MAJOR A (second SQL review): the retry blocked the writer thread for 28 seconds

`monitor/backend/src/main/kotlin/octometer/monitor/erasure/UserErasureService.kt:165-174`
(`checkpointAfterCommit`). The old form kept the default `busy_timeout` of 5000 ms, so five
`TRUNCATE` tries held the one writer thread for about 28 seconds, and `TRUNCATE` could report
`busy` for a reason with no privacy meaning: the erased bytes sit in the main file
`octometer.db`, not in the WAL file.

The fix: `PRAGMA busy_timeout=0` for the whole checkpoint block (restored after, in `finally`,
`UserErasureService.kt:166-173`). A `PASSIVE` checkpoint never waits; it copies each WAL frame
that no reader still needs. The privacy result is `log == checkpointed`
(`UserErasureService.kt:182-190`, `passiveCheckpointUntilMoved`). A maximum of 5 tries, 100 ms
apart, as before. One `TRUNCATE` checkpoint then runs once, only to reset the file length; its
`busy` result does not change the answer.

Measured: `UserErasureServiceTest.the write block completes within 2 seconds with a busy
reader` (`UserErasureServiceTest.kt:227-251`) took 0.486 s with a reader open on an old
snapshot, well inside the 2-second bound. `UserErasureServiceTest.a reader on the newest
snapshot still lets the checkpoint report checkpointed true`
(`UserErasureServiceTest.kt:257-291`) asserts `checkpointed = true` even with a reader open,
because that reader's snapshot needs no old frame.

I restored the old form for a moment (a plain 5-try `TRUNCATE` loop, no `busy_timeout=0`, no
`PASSIVE`). The time test then took about 28 seconds and failed the 2-second bound. I restored
the fix, and the suite passed again.

### MINOR 8 (second SQL review): a session with only a `kind = 2` row of the user

`monitor/backend/src/main/kotlin/octometer/monitor/erasure/UserErasureService.kt:56-57`
(`SQL_SESSION_IDS_OTHER_KIND`) and `:216` (`sessionIdsOfUser`). A third read on `event_session`
finds the session ids of a row with a kind other than 0 or 1. Contract rule C43 covers each
event kind, so a session whose only user row held such a kind kept its anonymous rows before
this fix.

`EXPLAIN QUERY PLAN` of the new statement: `SEARCH event USING COVERING INDEX event_session
(app_id=?)`. Asserted in
`UserErasureServiceTest.the session-id query uses the partial index of its kind`
(`UserErasureServiceTest.kt:293-311`).

Test: `UserErasureServiceTest.the erasure removes an anonymous row of a session whose only user
row has a kind other than 0 or 1` (`UserErasureServiceTest.kt:145-161`). I removed the new read
for a moment. The test then failed: the count stayed at 1, and the anonymous row stayed. I
restored the fix, and the test passed.

### MINOR 9 (second SQL review): the wrong name for a filter list

`monitor/backend/src/test/kotlin/octometer/monitor/LogCapture.kt:34` renames
`NON_SERVER_LOGGER_PREFIXES` to `KNOWN_URL_LOGGER_PREFIXES`. The comment at
`LogCapture.kt:19-33` names issue #150, which removes the filter once the `StatusPages` line no
longer names the URL.

### MINOR 10 (second SQL review): the README named the wrong file

`monitor/backend/README.md:28-38` no longer states that an erased byte can stay in the WAL
file. It now states that the byte stays in `octometer.db` until a checkpoint copies the new
page over the old one, and it warns against deleting a store file by hand.

### MINOR 11 (second SQL review) / new MINOR 1 (second privacy review): D13 missed a field

`docs/superpowers/specs/2026-09-21-octometer-design.md:318`. D13 now lists `checkpointed`
beside `deleted` in the erasure answer.

### MINOR 6 of round 1 (chunk note)

`monitor/backend/src/main/kotlin/octometer/monitor/erasure/UserErasureService.kt:102-108`. A
comment on `eraseUserEvents` states the measured cost of one large erasure (538 ms, 15.6 MB of
WAL, 50 000 events of one user in one transaction), and the reason no chunk form exists: a
second transaction could let a reader see a half-erased user between two chunks.

### The two new MINOR notes of the second privacy review that stay open

The second privacy review's new MINOR 2 (the README wording) and new MINOR 1 (D13) match SQL
review MINOR 10 and MINOR 11 above; both are fixed with the same edit. Its new MINOR 3 (remove
the `LogCapture.kt` filter once issue #150 lands) stays open by design: issue #150 is not done
yet, so the filter still has a job. The comment at `LogCapture.kt:19-33` names #150.

### The merge

`origin/main` was at commit `ba2ea5c` (#145, the level 2 totals). I merged it into the branch
with a merge commit, no rebase. The merge was clean, with no conflict in any file, matching the
second SQL review's report.

### Test results

`./gradlew :monitor:backend:test`, two runs in sequence, JDK 21
(`C:/Users/xamcr/.jdks/ms-21.0.9`): `BUILD SUCCESSFUL` each time, 261 tests, 0 failures, 0
errors, 23 test classes (`monitor/backend/build/test-results/test/TEST-*.xml`).
`./gradlew build` passes once, for the whole project.

### CI

`gh pr checks 147`: Frontend, JVM modules, Monitor backend on Windows, Secret scan, and Tracker
all `pass`.

xamcross:
**Review (third review, privacy engineer): APPROVE**

I made a new clone, I ran `gh pr checkout 147`, and I checked each item with JDK 21 at commit
`1acba76`. I ran `./gradlew :monitor:backend:test` one time: `BUILD SUCCESSFUL`, 23 test
classes, **261 tests**, 0 failures. I deleted my clone after the review. My earlier reviews are
the comments 5769624408 and 5770065012.

The new `PASSIVE` form keeps each privacy result of my second review, and it removes the
28-second block. No BLOCKER and no MAJOR stays, thus I approve.

## The byte check on the new form

**No reader.** Through the route, with the store still open:

```
status=200 body={"deleted":3,"checkpointed":true}
octometer.db     53248 bytes  user id: absent  session id: absent
octometer.db-wal     0 bytes  user id: absent  session id: absent
octometer.db-shm 32768 bytes  user id: absent  session id: absent
```

**A reader on an old snapshot.** I held a read transaction open on a second connection:

```
first call  = {"deleted":2,"checkpointed":false}   in 1403 ms
after the first call: the user id in octometer.db at offset 20459
warn count=1
warn = The WAL checkpoint after the erasure of app 1 did not complete.   holdsUser=false
```

The answer is 200 with `checkpointed: false`. Exactly one warning. It names the app id only,
and it holds no user id. The call took 1403 ms, not 28 seconds; the `busy_timeout=0` plus
`PASSIVE` form does what the SQL review asked for.

**The second call, after the reader closes.**

```
second call = {"deleted":0,"checkpointed":true}
octometer.db, octometer.db-wal, octometer.db-shm: the user id absent in each file
```

The remedy of the README works on the new form too.

The signal `log == checkpointed` of `passiveCheckpointUntilMoved`
(`UserErasureService.kt:182-190`) is the correct privacy signal. Every frame is then in the main
file, thus the zeroed pages replaced the old pages of `octometer.db`. That is the file where my
second-review probe found the stale bytes.

## Contract rule C43 for the other kinds: FIXED

`SQL_SESSION_IDS_OTHER_KIND` (`UserErasureService.kt:56-57`) and the third read in
`sessionIdsOfUser` (`:216`) close the gap. My probe seeded session A, whose only row of the user
holds `kind = 2`, plus two anonymous rows of that session; and session B with one anonymous row
and no row of the user:

```
deleted=3 checkpointed=true
rows after: [b1 sess-B null k0]
```

The `kind = 2` row of the user and both anonymous rows of session A are gone. The anonymous row
of the unrelated session B stays. This matches contract rule C43 for each event kind.

## My two document findings of the second review

- `monitor/backend/README.md:28-38` now states the correct file: "The erased bytes stay in the
  main file, `octometer.db`, until a checkpoint copies the new pages over the old ones." It no
  longer points the owner at the WAL file. FIXED.
- `docs/superpowers/specs/2026-09-21-octometer-design.md:318` now reads "The erasure returns
  `deleted` and `checkpointed`." FIXED.

My third finding (the sentinel filter of `LogCapture.kt`) now names issue #150 in its comment.
That is the right record.

I found no new BLOCKER, no new MAJOR, and no new MINOR. The work is correct. I approve.

xamcross:
**Review (third review, SQL engineer): APPROVE**

I made a new clone, I ran `gh pr checkout 147` (commit 1acba76), and I ran each check myself on
JDK 21. MAJOR A is fixed. Each MINOR finding is closed. No BLOCKER and no MAJOR stays.

## MAJOR A — the checkpoint block: FIXED

`UserErasureService.kt:165-203`. `checkpointAfterCommit` sets `busy_timeout=0`, runs `PASSIVE`
up to five times with a pause of 100 ms, takes `log == checkpointed` as the privacy result,
runs one `TRUNCATE` for the file length, and restores `DEFAULT_BUSY_TIMEOUT_MILLIS` in a
`finally`. `SqlitePragmas.kt:13` owns that constant, so the two values cannot drift apart.

My probe, in my own clone:

```
R3| busy_timeout before the erasure = 5000
R3| octometer.db holds the user id before the erasure = true
R3| busy case (reader on an OLD snapshot): checkpointed=false in 464 ms
R3| busy_timeout after the erasure = 5000

R3| newest-snapshot case: octometer.db holds the user id before = true
R3| newest-snapshot case: checkpointed=true in 21 ms
R3| newest-snapshot case: octometer.db holds the user id after = false

R3| the block threw = true (SQLiteException)
R3| after a throw: busy_timeout = 5000
```

Three facts. First, the busy case now takes 464 ms, not 28 291 ms. This matches your 0.486 s,
and it sits well inside the 2 s bound of your test. Second, the false alarm is gone: a reader
on the newest snapshot now gives `checkpointed = true`, and the byte search proves that the
user id left `octometer.db`. Third, the connection gets `busy_timeout=5000` back after a
normal block and after a throw. I forced the throw with a checkpoint inside an open
transaction, and the value still came back.

## The MINOR findings

|---|---|---|---|
| 6 (first review) | The transaction has no chunk limit | FIXED | `UserErasureService.kt:102-108` records 538 ms and 15.6 MB of WAL for 50 000 events, names the second transaction of a chunk form as the reason against it, and points at the purge of D15. This matches my measurement. |
| 8 | The session list read misses another kind | FIXED | `SQL_SESSION_IDS_OTHER_KIND` (`:56-57`) on `event_session`, added to the set at `:216`. The plan test at `:301` and the test `the erasure removes an anonymous row of a session whose only user row has a kind other than 0 or 1` (`:149`) cover it. |
| 9 | The `NON_SERVER_LOGGER_PREFIXES` name | FIXED | `LogCapture.kt:36` now reads `KNOWN_URL_LOGGER_PREFIXES`. |
| 10 | The README put the risky byte in the WAL file | FIXED | The section "The WAL checkpoint" now states: "The delete writes new, zeroed pages into the WAL file. The erased bytes stay in the main file, `octometer.db`, until a checkpoint copies the new pages over the old ones." That is exactly what my probe shows. |
| 11 | D13 did not list `checkpointed` | FIXED | `docs/superpowers/specs/2026-09-21-octometer-design.md:318`: "The erasure returns `deleted` and `checkpointed`." |

The advice for `checkpointed: false` in the README is now correct too. It names the store
files, not the WAL file alone, and it warns against a delete of `octometer.db*` by hand.

## The test run

`./gradlew :monitor:backend:test`, one run, JDK 21: `BUILD SUCCESSFUL`. 23 test classes, 261
tests, 0 failures, 0 errors, from `monitor/backend/build/test-results/test/TEST-*.xml`. This
matches your count.

## Summary

- BLOCKER: 0
- MAJOR: 0
- MINOR: 0 open

The SQL work is correct. Each statement names its index, each plan has an exact test, the
three index names have a `sqlite_master` test, the erasure covers every kind value, and the
checkpoint now gives a true privacy result in bounded time. I approve.

One item for the maintainer, outside this pull request: please file the new issue for the
framework DEBUG line of `io.ktor.server.plugins.statuspages.StatusPages`, which names the full
request URL. It affects each route with a query string.

PR #149, opened 2026-09-22
feat(monitor): purge the events above the retention limit

### LINKED ISSUE #59: Purge the events above the retention limit
**Goal.** The monitor deletes each event older than `retentionDays`, at the start and each 24 hours.

**Implementation steps.**
1. Read `retentionDays` from the config of #4.
2. Add the purge function. It deletes the events with a `ts` older than `retentionDays` in chunks of 10 000 rows.
3. Run `PRAGMA wal_checkpoint(TRUNCATE)` after each purge.
4. Run the purge at the start and each 24 hours. Inject a `Clock` for the tests.
5. Return `retentionDays` in `GET /api/health`.

**Acceptance criteria.**
- [x] The purge runs at the start and each 24 hours (fixed clock).
- [x] The purge deletes in chunks of 10 000 rows, one transaction each.
- [x] `PRAGMA secure_delete` returns 1.
- [x] After a purge `wal_checkpoint(TRUNCATE)` runs.
- [x] `GET /api/health` returns `retentionDays`.

**Related.** #60 sets the default of `retentionDays`. #66 shows the value on the level 1 view.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, D2, D13, and D15.

xamcross:
## Summary

This change adds the purge job of issue #59. The purge deletes each event
older than `retentionDays`, in batches of 10 000 rows. It deletes a click
row and a session start row alike. It runs `PRAGMA wal_checkpoint(TRUNCATE)`
once, after the last batch. `MonitorServices` starts the job at the open
call and stops it at the close call. `GET /api/health` now answers with
`retentionDays`.

## Acceptance criteria

- [x] The purge runs at the start and each 24 hours (fixed clock). Proof:
  `RetentionPurgeJobTest` (`start runs the action once at once`, `the
  action runs again once the wait call returns`, and, since correction
  round 1, `start waits 24 hours between two runs`, which asserts the
  literal `86 400 000`) gives the job a fake wait function, so the test
  controls each run with no real sleep. `RetentionPurgeTest` gives
  `RetentionPurge` a fixed `java.time.Clock` for the cutoff, so the
  boundary tests need no real time either.
- [x] The purge deletes in chunks of 10 000 rows, one transaction each.
  Proof: `RetentionPurgeTest.purgeOnce deletes every old row, across more
  than one batch of 10 000` inserts 10 003 old rows and checks the purge
  removes each one. Each batch runs as its own call to `database.write`,
  and each call holds one SQL statement in the default autocommit mode of
  the JDBC connection, so SQLite commits it as its own transaction. Since
  correction round 1, `purgeOnce writes exactly 3 times for 10 001 old
  rows, with a batch size of 10 000` counts the write calls, so a
  different batch size now fails this test.
- [x] `PRAGMA secure_delete` returns 1. Proof: the existing `PragmaTest`
  (`monitor/backend/src/test/kotlin/octometer/monitor/store/PragmaTest.kt`)
  already checks this pragma on the writer and the reader connection; it
  passed in this test run too (see the test count below).
- [x] After a purge `wal_checkpoint(TRUNCATE)` runs. Proof:
  `RetentionPurgeTest.purgeOnce empties the wal file after the checkpoint`
  checks the `octometer.db-wal` file size on the disk, not only the
  pragma return value. Since correction round 1, the purge reads the
  checkpoint answer row (`busy`, `log`, `checkpointed`) and sets
  `busy_timeout` to 0 for that one call.
- [x] `GET /api/health` returns `retentionDays`. Proof:
  `HealthRouteTest.the health route answers with retentionDays`.
- [x] `./gradlew :monitor:backend:test` passes. Proof: run two times in
  sequence, 285 tests, 0 failures, 0 errors, each time.
- [x] A comment on issue #59 records the test result. The first comment
  named 187 tests; a later comment on the issue gives the real count,
  285, after correction round 1.

## The boundary rule

`RetentionPurgeTest` gives three events around the cutoff:

- exactly `retentionDays` old: the purge **keeps** it (not "older than"
  the limit, per the issue text).
- one second older than `retentionDays`: the purge **deletes** it.
- one second younger than `retentionDays`: the purge **keeps** it.

## The index check (issue #59, "the delete statement uses an index")

`EXPLAIN QUERY PLAN` of the batch delete:

```
SEARCH event USING INTEGER PRIMARY KEY (rowid=?), LIST SUBQUERY 1, SCAN event USING COVERING INDEX event_session
```

Source: `RetentionPurgeTest.the delete statement query plan is recorded`,
run on 2026-09-22. The outer delete finds each row by `rowid` (fast). The
inner subquery, the real cost of the purge, runs as a full scan of the
covering index `event_session`, because `ts` is not the leading column of
any index of section 6 of the design document, and the purge has no
`app_id` or `kind` filter to narrow the search. SQLite picks
`event_session` over a raw table scan because it is a covering index (it
carries `ts`, so SQLite reads the index pages only, not the table rows),
but the scan still visits every row of the table.

**Finding for the maintainer:** no index of section 6 fits a targeted
purge by `ts` alone. This pull request adds no migration, because issue
#59 states no rule for a new index. The full scan runs once a day, on a
schedule, and each batch runs on its own `database.write` call, so it
never blocks a page commit for the whole purge; the maintainer may still
want a dedicated `ts` index in a later issue, if the event table grows
large enough for a daily full scan to matter.

## What stops a second copy of a purged event

The purge only deletes rows of the `event` table. It never changes
`app.cursor`. The reader (issue #9, D4) resumes each app from that
cursor, not by scanning the local `event` table, so a MongoDB document
that the purge already deleted locally is never requested again: the
cursor already sits past it.

## Design decisions the issue did not state

- The purge job runs on its own dedicated thread
  (`octometer-retention-purge`), separate from the SQLite writer thread,
  so a test can watch the thread start and stop, the same way
  `MonitorServicesTest` already watches the SQLite writer thread.
- The 24-hour gap between two runs comes from an injectable wait
  function, not from the injected `Clock`. The `Clock` sets the purge
  cutoff (`now - retentionDays`); the wait function controls the loop
  timing in a test. A real coroutine `delay` cannot follow a fake clock,
  so the two concerns use two different test seams.
- The purge job starts before the orphan-secret sweep inside
  `MonitorServices.open()`, not after. This also gives a natural test of
  "a throw in open stops the job too": the existing "app table missing"
  failure test now checks the purge thread has no leak, next to the
  writer thread check that was already there.
- `MonitorServices.close()` cancels the purge job with `cancelAndJoin`
  before it closes the database, so a purge run in progress finishes or
  stops at its next suspension point before the connection closes.

## Test count

`./gradlew :monitor:backend:test`, run two times in sequence: 285 tests,
0 failures, 0 errors, both times (after the merge of `origin/main`
through commit `01d07b8`, and after correction round 1 below).
`./gradlew build` passed once, across every module.

## The merge with pull request #127 (issue #18)

Commit 6154083 (pull request #127, `GET /api/apps`) landed on `main`
while this branch was open. This branch merges it with a merge commit.
The merge itself had no conflict, but it exposed a real interaction:

- The purge now runs at every `MonitorServices.open()` call, on the real
  clock, with the default `retentionDays` (395). `AppTotalsRouteTest`
  seeded its event with the fixed instant `1_700_000_000_000L`
  (2023-11-14), then started a real server. The purge deleted that event
  before the test read `GET /api/apps`, so the test failed on this
  branch, although it passes on `main` alone (no purge job runs there
  yet).
- Fix: `AppTotalsRouteTest` now gives its event the real current time,
  well inside the retention window. This is the smallest change that
  keeps the test green; the maintainer may want a similar check for a
  future test that starts a real server with an old fixed timestamp.
- A second, unrelated CI failure appeared on the Windows job:
  `ApplicationLifecycleTest` starts and stops a real server twice in one
  folder, and checks the writer thread count right after each stop, with
  no poll. The fix at the time gave `stop()` a 300 ms bound, then moved
  on regardless. **Correction round 1 removes that bound** (see below);
  it was a repair of a symptom, not of the real cause.

## Correction round 1 (two independent reviews)

A reliability review and a SQLite review each found MAJOR issues. The
full list, with the file and the line for each finding, is in a
[separate comment](https://github.com/xamcross/octometer/pull/149#issuecomment-5770543252)
on this pull request. In short:

- `stop()` no longer bounds the join at a fixed time. It cancels the job,
  then joins it with `cancelAndJoin`, with no time limit. The batch loop
  of `RetentionPurge.purgeOnce` checks the cancel between two batches, so
  the join still ends soon. A test holds a database lock so the first
  batch of a 100 000-row purge blocks, then closes the services and
  asserts no ERROR line comes after `close()` returns.
- The `CancellationException` catch of `RetentionPurgeJob` now rethrows
  only when the job itself is no longer active, the same rule
  `Application.kt` already uses for a route handler.
- The 24-hour interval and the 10 000-row batch size each have a test
  that fails when the constant changes, proven by a break-and-restore
  check of each one.
- `devConfig()`/`prodConfig()` now default `retentionDays` to a large
  value, because `MonitorServices.open()` runs a real purge at every
  open call, on the real clock; a small default purged the fixed
  fixtures of three other route tests once pull requests #144, #145, and
  #147 landed on `main`.
- The checkpoint now reads its answer row and sets `busy_timeout` to 0
  for that one call, instead of retrying with the normal timeout.
- The module now uses one clock type, `java.time.Clock`.

On the Windows CI job, `ApplicationLifecycleTest` (a test this pull
request does not own or change) failed twice on a retry, then passed on
a third retry. The same test failed once on an earlier commit of this
same pull request, before this correction round existed, and passed on
the commit right after it. This looks like a pre-existing, intermittent
flake of that test on the Windows runner, matching the mechanism the
reliability review already named (`ExecutorCoroutineDispatcher.close()`
not waiting for its thread), which belongs to pull request #146 (issue
#55), not to this issue.

Closes #59

Generated with Claude Code


xamcross:
**Review (reliability engineer): CHANGES REQUESTED**

## What I ran

I made my own clone and ran `gh pr checkout 149`. The head commit is `e6e6b7f`.
I used JDK 21 through the Gradle toolchain.

- `./gradlew :monitor:backend:test`, two times in sequence: **198 tests, 0 failures, 0 errors**, each time. The second run used `--rerun-tasks`, so it ran each test again.
- 10 mutations of the code, each one with a targeted test run, then a restore.
- 9 experiments with a real `MonitorServices`, a real Netty server, and up to 200 000 rows.

Each number below comes from my own run.

## The acceptance criteria

| # | Criterion | Result |
| - | --------- | ------ |
| 1 | The purge runs at the start and each 24 hours (fixed clock) | **PARTIAL** — the start run passes; the 24-hour gap has no test (MAJOR 1) |
| 2 | The purge deletes in chunks of 10 000 rows, one transaction each | **PARTIAL** — the code is correct; the chunk size has no test (MAJOR 2) |
| 3 | `PRAGMA secure_delete` returns 1 | PASS — `PragmaTest.kt:57` asserts it on each connection |
| 4 | After a purge `wal_checkpoint(TRUNCATE)` runs | PASS — I replaced the checkpoint with `SELECT 1`, and `purgeOnce empties the wal file after the checkpoint` failed |
| 5 | `GET /api/health` returns `retentionDays` | PASS — a real server answered `{"version":"0.1.0","mode":"prod","refreshSeconds":60,"retentionDays":395}`. A wrong value in `Application.kt` fails `HealthRouteTest` |
| 6 | `./gradlew :monitor:backend:test` passes | PASS — 198 tests, two times |
| 7 | A comment on issue #59 records the test result | PASS with a MINOR — the comment names 187 tests; the branch holds 198 (MINOR 5) |

## The checks of my role

**1. The lifecycle.** PASS. `open` starts the job, `close()` stops it, and a throw in a later step of `open` stops it too. I made the first step of the inner `try` of `MonitorServices.open` throw. Two seconds after the throw the thread list held no `octometer-*` thread and no thread with a coroutine name. I restored the file. The two failure paths also have a real test: when I remove `retentionPurgeJob.stop()` from `close()`, `open starts the purge thread, and close stops it` fails; when I remove it from the inner `catch`, `open throws and leaves no writer thread and no purge thread, when the app table is missing` fails.

**2. `stop()` with the bound of 300 ms.** FAIL — see MAJOR 3. The measured facts:

- A purge of 100 000 old rows, then `close()` at once: `close()` took **73 ms**, the purge deleted one batch of 10 000 rows, no exception left `close()`, and no thread stayed. The cancel takes effect between two batches, because `SqliteDatabase.write` starts each batch with `withContext`, and `withContext` checks the cancel first.
- The dangerous case: a second connection held `BEGIN EXCLUSIVE`, so the first batch waited inside SQLite (`busy_timeout=5000`). `close()` then took **4773 ms**, not 300 ms, because `SqliteDatabase.close()` calls `writer.close()`, and the driver waits for the statement that runs. The purge thread was gone, but the purge coroutine moved to a shared dispatcher and wrote `ERROR ... The purge run failed. SQLiteException` **after** `close()` returned. I saw no corruption and no crash.
- A full purge of 100 000 rows takes **883 ms**. A purge with no old row takes **2 ms**.

**Was the 300 ms a repair of the symptom?** Yes. `ApplicationLifecycleTest` starts and stops a real server against an empty folder, so its purge takes about 2 ms; a slow purge cannot explain that failure. The cause is `ExecutorCoroutineDispatcher.close()`: it asks the executor to shut down, and it does not wait for the thread to end, so the count right after `server.stop` is a race. Pull request #146 removes that cause: its `close()` calls `awaitWriterShutdown(writerExecutor)` with `awaitTermination`. It keeps the order `writer.close()` first, so it does not make a batch safe during a close; it does remove the thread race that #149 tried to hide.

**A correct `stop()`:** cancel the coroutine, then join **without a fixed time**, and check the cancel between two batches. Concretely:

```kotlin
fun stop() {
    val currentJob = job
    if (currentJob != null) {
        runBlocking { currentJob.cancelAndJoin() }
    }
    scope.cancel()
    dispatcher.close()
    job = null
}
```

and in `RetentionPurge.purgeOnce`, at the top of the loop:

```kotlin
while (true) {
    currentCoroutineContext().ensureActive()
    ...
}
```

The join then ends after one batch, which is 90 ms in my measurement, and no coroutine outlives `close()`.

**3. The schedule on the real dispatcher and the real delay.** The job gives `86 400 000` ms to the wait function (I read the value with my own wait function). The default wait is `delay`, and `delay` counts with `System.nanoTime()`, not with the wall clock. `RetentionPurgeJob` holds no `Clock`, so it never asks what the time is.

**What occurs after a sleep of 10 hours?** On this Windows machine the timer counts the sleep, so the next purge runs at its normal time; when the 24-hour point falls inside the sleep, the purge runs at once at the wake. It does **not** wait 24 hours after the wake. My evidence, read on this machine:

- `powercfg /a` gives `Standby (S0 Low Power Idle)`; S1, S2, and S3 are not available.
- The Kernel-Power events 506 and 507 since the last boot sum to **77.384 hours** of standby.
- The wall-clock uptime is **143.234 hours**, and `System.nanoTime()` since its origin is **143.235 hours**. The two agree, thus the monotonic timer counted each standby hour.

I did not measure a hibernate. A wall-clock anchor removes the doubt: see MINOR 6.

**4. A failure inside a run.** PASS. I broke the batch two ways with a real `MonitorServices`:

- The `event` table absent: one line, `ERROR o.m.r.RetentionPurgeJob - The purge run failed. SQLiteException`.
- A locked database (a second connection with `BEGIN EXCLUSIVE`): one line, the same text, after the busy timeout of 5 s.

In each case the thread list still held `octometer-retention-purge` and `octometer-sqlite-writer`, so the job stays alive for the next tick. The line holds the class name alone, thus no user data. A test proves the rule too: `a failed run does not stop the loop`. I did not produce a real `SQLITE_FULL`; the code path is the same `catch (failure: Throwable)`.

**5. The purge at the start.** It runs **in parallel**, and it does not delay the health route. With 200 000 rows and 20 000 old rows: `server.start(wait = false)` returned after 1575 ms, and `GET /api/health` answered 200 after **2126 ms**. The purge wrote its line while that request was in flight. The same server with no old row answered after **2475 ms**. A second health call took 7 ms. The health route reads no database row, thus the purge cannot block it. A write does wait: a `database.write` call during the purge waited **74 ms to 111 ms**, which is one batch. That is the intended cost of the one writer thread of D3.

**6. The log lines.** One line for each run, with the count alone, and no line for each batch. No user data. Two gaps: a run with no old row still writes an INFO line (MINOR 1), and a cancelled run writes no line at all (MINOR 2).

**7. The config.** PASS. `retentionDays` of `0`, of `-1`, and of `many` each give an `InvalidConfigException` through the existing `toValidInt` check, and the new tests cover each one. I widened the range to `0..Int.MAX_VALUE`, and `a retentionDays of 0 gives an InvalidConfigException` failed, so the tests are real. An absent key gives 395 (`ConfigLoaderTest.kt:91`).

**8. The tests.** No test sleeps for a real interval of the product. I broke the code 10 times. 8 breaks gave a failure, and 2 breaks gave no failure at all:

| Break | Result |
| ----- | ------ |
| `PURGE_INTERVAL_MILLIS` 24 h to 1 s | **every test passed** |
| `PURGE_BATCH_SIZE` 10 000 to 3 000 | **every test passed** |
| `ts < ?` to `ts <= ?` | `an event of exactly retentionDays age stays` failed |
| The checkpoint replaced by `SELECT 1` | `purgeOnce empties the wal file after the checkpoint` failed |
| Stop after the first batch | `purgeOnce deletes every old row, across more than one batch of 10 000` failed |
| The cancellation catch writes an error | `stop while the action runs writes no error log line` failed |
| `close()` does not stop the job | `open starts the purge thread, and close stops it` failed |
| A failed `open` does not stop the job | `open throws and leaves no writer thread and no purge thread...` failed |
| The health route answers `retentionDays = 0` | `the health route answers with retentionDays` failed |
| The config admits a `retentionDays` of 0 | `a retentionDays of 0 gives an InvalidConfigException` failed |

No test leaked a thread. One test class leaks a temporary folder in one case: see MINOR 3.

## Findings

### MAJOR 1 — the 24-hour gap has no test
`monitor/backend/src/main/kotlin/octometer/monitor/retention/RetentionPurgeJob.kt:21`, `monitor/backend/src/test/kotlin/octometer/monitor/retention/RetentionPurgeJobTest.kt:41-65`

Each test gives the job a wait function of the form `wait = { waitGate.receive() }`, which drops the `millis` argument. I set `PURGE_INTERVAL_MILLIS` to `1000L` and ran the retention tests, `MonitorServicesTest`, `HealthRouteTest`, and `ApplicationLifecycleTest`: every test passed. The first acceptance criterion of issue #59 thus has no proof, and the pull request text claims one.

**The fix.** Record the argument in the fake wait function, and assert it:

```kotlin
val waited = Channel<Long>(Channel.UNLIMITED)
val job = RetentionPurgeJob(wait = { millis -> waited.send(millis); waitGate.receive() }, action = { ... })
...
assertEquals(24L * 60 * 60 * 1000, withTimeout(5_000) { waited.receive() })
```

### MAJOR 2 — the chunk size of 10 000 has no test
`monitor/backend/src/main/kotlin/octometer/monitor/retention/RetentionPurge.kt:12`, `monitor/backend/src/test/kotlin/octometer/monitor/retention/RetentionPurgeTest.kt:90-100`

`purgeOnce deletes every old row, across more than one batch of 10 000` passes with any chunk size. I set `PURGE_BATCH_SIZE` to `3_000` and ran the retention tests and `MonitorServicesTest`: every test passed. The second acceptance criterion names the number 10 000, thus a test must hold it. The claim "one transaction each" has no test either, although the code is correct: each batch is one `database.write` call with one statement in autocommit mode, and `EventStore` keeps its own `BEGIN IMMEDIATE` inside one `write` call.

**The fix.** Give `RetentionPurge` a counter seam, or count the write calls. The smallest form: insert 10 001 old rows, wrap `database` with a decorator that counts each `write` call, and assert 3 calls (two delete batches plus the checkpoint). A batch size of 3 000 then gives 5 calls, and the test fails.

### MAJOR 3 — `stop()` does not bound the stop, and the purge coroutine outlives `close()`
`monitor/backend/src/main/kotlin/octometer/monitor/retention/RetentionPurgeJob.kt:25-30` and `:70-79`

The comment states that a slow run "never holds up the application stop event for long". My measurement contradicts it. With one batch blocked inside SQLite, `MonitorServices.close()` took **4773 ms**, because `SqliteDatabase.close()` waits for the statement that runs. The 300 ms bound thus gives no bound at all; it only lets `stop()` return while the coroutine still runs. That coroutine then moved to a shared dispatcher, and it wrote an ERROR line after the application stop. The structure rule of the backend brief is "`close()` stops it", and this `close()` does not always stop it.

The bound also hides a defect that belongs to a different part. `ApplicationLifecycleTest` uses an empty folder, where a purge takes 2 ms, so the purge never caused that failure. The cause is the dispatcher close that does not wait for the thread. Pull request #146 (issue #55) removes that cause with `awaitTermination` on the writer executor.

**The fix.** Remove `STOP_JOIN_TIMEOUT_MILLIS`. Use `runBlocking { currentJob.cancelAndJoin() }`. Add `currentCoroutineContext().ensureActive()` at the top of the batch loop of `RetentionPurge.purgeOnce`, so the cancel check between two batches is explicit and not a side effect of `withContext`. Keep the order of `MonitorServices.close()`: stop the job first, then close the database. Add a test that starts a purge of more than one batch and asserts that `close()` returns after the current batch, with no ERROR line after it.

### MAJOR 4 — the catch of a `CancellationException` has no `isActive` check
`monitor/backend/src/main/kotlin/octometer/monitor/retention/RetentionPurgeJob.kt:84-85`

`runOnce` throws each `CancellationException` again, with no test of the state of the coroutine. The Kotlin type is an alias of `java.util.concurrent.CancellationException`, thus a throw from inside the action kills the loop although nobody cancelled the job. This is lesson 2 of the backend brief. I proved it: my action threw one `java.util.concurrent.CancellationException` on its first run, while the job stayed active. The run count stopped at 1, the second run never came, and the capture held **0 log events**. In production the purge then stops for the life of the process, and the owner sees no line.

**The fix.**

```kotlin
    if (!coroutineContext.isActive) throw cancellation
    log.error("The purge run failed. {}", cancellation.javaClass.simpleName)
}
```

Add a test with the form of my experiment: a foreign `java.util.concurrent.CancellationException` from the action must give one ERROR line and a second run.

### MINOR 1 — a run with no old row writes an INFO line each day
`monitor/backend/src/main/kotlin/octometer/monitor/retention/RetentionPurge.kt:56`. I started a real `MonitorServices` on a store with 5 fresh rows, and the log held `INFO ... The purge deleted 0 event row(s).` Write a DEBUG line for a zero count, and keep the INFO line for a count above zero.

### MINOR 2 — a cancelled purge writes no line
`RetentionPurge.kt:48-56`. In my first close experiment the purge deleted 10 000 rows, and then the stop ended it. The log held no purge line, because the INFO line sits after the loop. The owner cannot see that a purge started and stopped. Write the count in a `finally` block, or write one DEBUG line at the start of each run.

### MINOR 3 — `RetentionPurgeTest` does not register its temporary folder
`monitor/backend/src/test/kotlin/octometer/monitor/retention/RetentionPurgeTest.kt:27,37-41`. The class calls `Files.createTempDirectory` and deletes the folder in `@AfterTest`. When `setUp` throws, `tearDown` touches the `lateinit` field first, and the folder stays in `%TEMP%`. Issue #142 exists for this defect. Add `.also { registerTempRoot(it) }`, as `ApplicationLifecycleTest.kt:38` does.

### MINOR 4 — `MonitorServices` gives the purge no `Clock` seam
`monitor/backend/src/main/kotlin/octometer/monitor/MonitorServices.kt:98-101`. `newRetentionPurgeJob` fixes `SystemClock`. Each test that opens a real `MonitorServices` thus runs a real purge on the real clock, and this already forced the change of `AppTotalsRouteTest.kt:281`. Step 4 of the issue asks for an injected `Clock`. Add `clock: Clock = SystemClock` to `MonitorServices.open`, the same way pull request #146 adds a clock to `SqliteDatabase.open`.

### MINOR 5 — the issue comment names a stale test count
The comment on issue #59 names 187 tests. The branch holds 198. Update the comment after the corrections.

### MINOR 6 — the schedule holds no wall-clock anchor
`RetentionPurgeJob.kt:57-62`. The loop counts with `delay` alone. On this machine the monotonic timer counts each standby hour (my measurement above), so the gap is correct today. A hibernate, a different machine, or a later change of the timer gives a different result, and nothing in the code states the rule. Keep the last run time from the injected `Clock`, and wait `max(0, nextDue - clock.nowMillis())` in short steps. Decision D7 already holds a wake rule for the poll loop, for the same reason.

### MINOR 7 — a forward jump of the system clock deletes more than the window
`RetentionPurge.kt:46`. The cutoff comes from `System.currentTimeMillis()` with no guard. A clock that jumps forward by one year deletes one year of events at the next run, and `secure_delete` makes that permanent. Add a guard: write a WARN line when one run deletes more rows than a daily maximum that you set, or keep the cutoff of the last run and refuse a large jump.

### MINOR 8 — `start()` after `stop()` does nothing, and it reports success
`RetentionPurgeJob.kt:55-63` and `:70-79`. `stop()` cancels the scope and sets `job = null`, thus `check(job == null)` passes, `scope.launch` returns a job that never runs, and the caller sees no error. No caller does this today. Add `check` on a `stopped` flag, or build a new scope in `start()`.

- The delete deletes a click row and a session start row alike, and the boundary rule holds: exactly `retentionDays` old stays, one second older goes, one second younger stays.
- The cutoff uses a bound parameter, not a literal in the SQL text.
- The log lines hold no user id, no path, and no connection string.
- The language of the new comments and of the commit bodies follows Simplified Technical English.
- Each CI check of the pull request passes.
- The purge never touches `app.cursor`, thus the reader does not fetch a purged event again.
- The honest note about the index: no index of section 6 leads with `ts`. The measured cost is acceptable today (883 ms for 100 000 rows), and a dedicated index belongs to its own issue.

I restored each file that I changed for an experiment, and I deleted my clone.

xamcross:
**Review (SQLite and backend engineer): CHANGES REQUESTED**

I made my own clone, ran `gh pr checkout 149`, and ran each check there on JDK 21 (the Gradle
toolchain picked `ms-21.0.9`). Head commit `e6e6b7f`. I pushed nothing and I changed no branch.

The purge logic, the boundary rule, the batch design, and the cursor rule are each correct.
The numbers below support the decision to add no index now. Three MAJOR findings stop an
approval: a cancellation catch that stops the job for the life of the process, a test that
hangs instead of fails, and a purge at every `MonitorServices.open()` that eats a test
fixture.

## The acceptance criteria of issue #59

|---|---|---|
| The purge runs at the start and each 24 hours (fixed clock) | PASS with a gap | `MonitorServicesTest.open starts the purge thread, and close stops it` passes. I read the wait argument with my own test double: it is `86400000`. No test of the branch reads that value. See MINOR 2. |
| The purge deletes in chunks of 10 000 rows, one transaction each | PASS | A break of `RetentionPurge.kt:51` (`<` to `<=`) made `purgeOnce deletes every old row, across more than one batch of 10 000` fail. One batch commits on its own: I closed the store in the middle of a purge of 400 000 old rows, reopened the file, and counted exactly 390 000 rows. |
| `PRAGMA secure_delete` returns 1 | PASS | `PragmaTest.kt:57` asserts the value on the writer connection and on the reader connection. |
| After a purge `wal_checkpoint(TRUNCATE)` runs | PASS | A break that removed the checkpoint call made `purgeOnce empties the wal file after the checkpoint` fail. See MINOR 4 for the unchecked answer of that pragma. |
| `GET /api/health` returns `retentionDays` | PASS | A break to a fixed `395` made `HealthRouteTest.the health route answers with retentionDays` fail. |
| `./gradlew :monitor:backend:test` passes | PASS | 198 tests, 0 failures, 0 errors. I ran `cleanTest test` two times in sequence. |
| A comment on issue #59 records the test result | PARTIAL | The issue comment states 187 tests. The real count is 198. See MINOR 7. |

The diff holds no secret, no database file, and no build output. I checked each of the 13
files.

## The delete statement, on 200 000 rows

I seeded 200 000 event rows in a test of my own and ran the plan on the real statement:

```
EXPLAIN QUERY PLAN DELETE FROM event WHERE rowid IN (SELECT rowid FROM event WHERE ts < ? LIMIT 10000)
SEARCH event USING INTEGER PRIMARY KEY (rowid=?), LIST SUBQUERY 1, SCAN event USING COVERING INDEX event_session
```

The report in the pull request text is correct. No index of section 6 has `ts` as a leading
column, thus the inner subquery scans the whole covering index `event_session`.

My measurements, each one in a temporary folder, with the pragmas of D3:

- 200 000 rows, 20 000 old rows (10 percent): **816 ms** for one `purgeOnce` call. That
  covers three delete batches plus the checkpoint.
- 200 000 rows, 0 old rows (the daily case): **19 ms, 15 ms, 15 ms** across three calls.

**The scan is acceptable at the design cap.** The index fits in the page cache of D3
(`cache_size=-65536`, 64 MB). D21 caps one app collection at 200 000 documents, and the pilot
app adds fewer than 500 visits each day, thus 200 000 rows is the right scale for the monitor
store. A daily no-op purge costs about 15 ms, and a purge of 10 percent of the table costs
under one second across three separate `database.write` calls.

**My recommendation:** add no index in this pull request. File a later issue for an index on
`(ts)` only when the store passes about 5 million rows. There the daily scan reaches about
0.4 s, and a new index also costs write time on each page commit of D4.

## The batches

- **The writer thread is free between two batches.** I ran a purge of 60 000 old rows
  together with 40 inserts from a second caller. 7 of the 40 inserts completed before the
  purge call returned. Each batch runs in its own `database.write { }` call, thus the
  single-thread queue admits a foreign task between two batches.
- **A batch that throws stops the run with one log line.** I added a `BEFORE DELETE` trigger
  with `RAISE(ABORT)`. The run wrote exactly one line: `The purge run failed. SQLiteException`.
  That line holds no event data (D15).
- **No partial state breaks the next run.** After I dropped the trigger, the next `purgeOnce`
  call deleted each of the 25 000 old rows.

## The boundary

`RetentionPurge.kt:23` uses `ts < ?`, with `cutoff = now - retentionDays * 86 400 000`.

- Exactly `retentionDays` old: the purge **keeps** it.
- One second older: the purge **deletes** it.
- One second younger: the purge **keeps** it.

The issue text says "older than `retentionDays`". D15 says "each event older than
`retentionDays`". The code keeps the same side as both texts. The three tests at
`RetentionPurgeTest.kt:58`, `:68`, and `:78` assert it, and a break of `<` to `<=` made the
first one fail.

## The cursor

Confirmed in the code. `RetentionPurge.kt` holds two SQL texts: the delete of `event` rows at
lines 21 to 25, and the checkpoint pragma at line 27. No statement reads or writes the `app`
table. A delete of an `event` row starts no cascade, because the foreign key points from
`event` to `app`. The reader resumes from `app.cursor`, thus a purged event never reaches the
store a second time.

## `stop()` with a bounded wait of 300 ms

I tried to prove the race, and I **refute** it. Three runs of the same probe: I seeded 400 000
old rows, called `MonitorServices.open()`, then called `close()` at once.

- `close()` took 493 ms, 525 ms, and 423 ms.
- No ERROR line, in any run. I captured each log event of the root logger for 5 seconds after
  the close call.
- No thread named `octometer-retention-purge` and none named `octometer-sqlite-writer` stayed
  alive.
- The file kept 390 000 rows each time, and `PRAGMA quick_check` answered `ok`.

The reason is structural, not luck. `SqliteDatabase.write` is `withContext(writerDispatcher)`,
and a cancelled coroutine never runs its dispatched block. The purge therefore stops at a
batch boundary and never sends a statement to a closed connection. A second probe shows the
other half: I ran one long delete on the writer thread, then called `database.close()` after
150 ms. The close call **blocked for 5623 ms**, until the statement ended, and the delete
returned 400 000 rows. `Connection.close()` of the driver waits for the running statement.

**What a correct `stop()` must do:** cancel the coroutine, then join it, with no fixed time.
The cancel call already bounds the join to the current batch. See MINOR 1.

## The second job: a local test merge with pull request #146

I fetched `pull/146/head`, made a throwaway branch from this branch, and merged it.

- **One conflict**, in `MonitorServices.kt`: the constructor parameter, the `close()` body,
  and the `open()` return statement. I resolved it by hand: both jobs in the constructor,
  both stop calls in `close()`, and the backup job start inside the inner `try` of this branch.
- **Result: 216 tests, 0 failures, 0 errors.** The `awaitTermination` change of #146 in
  `SqliteDatabase.close()` gave no failure next to the purge job.
- I deleted the throwaway branch and the `pr146` ref after the run.

Note for the real merge: #146 adds `java.time.Clock` to `MonitorServices.open()`, and this
branch adds its own `octometer.monitor.retention.Clock`. The two live side by side after the
merge. See MINOR 8.

## The size of the `Application.kt` change

8 lines, and each one is needed. D13 lists `retentionDays` in the health body. The change adds
one field to `HealthResponse` and one line to the route. It adds no route and no import. This
meets the rule of the brief.

# The findings

## MAJOR 1 — `monitor/backend/src/main/kotlin/octometer/monitor/retention/RetentionPurgeJob.kt:84`

```kotlin
    throw cancellation
```

`kotlinx.coroutines.CancellationException` is a type alias of
`java.util.concurrent.CancellationException`. A store call or an executor can throw that exact
class while this coroutine stays active. The loop then ends for the life of the process, with
no log line, and the daily purge never runs again.

**My evidence.** I gave the job an action that throws
`java.util.concurrent.CancellationException("a store task was cancelled")` on the first run,
and a wait function that I control. The result: `the second run happened = false,
runCount = 1`. The loop was dead, and nothing wrote a log line. Lesson 2 of the backend brief
names this pattern, and this repository already holds the correct form at `Application.kt:94`:
`if (cause is CancellationException && !call.isActive) throw cause`.

**The fix.** Throw it again only when the coroutine is no longer active:

```kotlin
    log.error("The purge run failed. {}", cancellation.javaClass.simpleName)
}
```

Add a test with the probe above: the loop must write one ERROR line, and the next run must
still start.

## MAJOR 2 — `monitor/backend/src/test/kotlin/octometer/monitor/retention/RetentionPurgeJobTest.kt:58` and `:142`

`waitGate.send(Unit)` stands outside any timeout, and `waitGate` is a rendezvous channel. When
a change stops the loop, no coroutine receives from that channel, and the send call waits
forever. The test then **hangs**; it never fails.

**My evidence.** In a break experiment I changed the exception type in the catch branch of
`runOnce`. The task `:monitor:backend:test` then ran for more than 12 minutes with no result,
and I stopped the worker by hand. In CI this burns the whole job, and it hides the regression
that the test exists for.

**The fix.** Put each send call inside the timeout that already guards the receive call:

```kotlin
withTimeout(5_000) {
    waitGate.send(Unit)
    runSignal.receive()
}
```

## MAJOR 3 — `monitor/backend/src/test/kotlin/octometer/monitor/apps/AppTotalsRouteTest.kt:277-281`

`Application.module()` opens `MonitorServices` at `Application.kt:77`, thus each test that
starts a real server runs a real purge on the real clock, with `retentionDays = 395`, in the
folder of that test. This pull request gives one fixture `System.currentTimeMillis()`. Two
problems stay:

1. The trap stays open for each later test. The other open branches (#18, #38, #50, #51, #55,
   #61) each add a route. A new route test with a fixed past timestamp loses its rows.
2. The purge runs on its own thread, next to the test body. A test that inserts an old row
   after the first request wins or loses by timing, not by logic.

**The rule that I recommend.** Keep the purge at the start, because D15 asks for it. Set a
large `retentionDays` in the shared test helper instead. `TestConfigs.kt:56` and `:65` already
take the parameter, thus the change is two default values:

```kotlin
// A test that starts a real server runs the purge of issue #59 on the real
// clock. A large default keeps each fixed fixture. A test of the purge
// passes its own value.
fun devConfig(dataDir: String = testDataDir(), retentionDays: Int = 3_650_000) = ...
fun prodConfig(dataDir: String = testDataDir(), retentionDays: Int = 3_650_000) = ...
```

Then give `AppTotalsRouteTest.kt:281` the fixed instant `1_700_000_000_000L` again, so that
test stays deterministic.

I reject the other option, a delay before the first purge. D15 states "The purge runs at the
start", and a delay also hides the job from `MonitorServicesTest`.

## MINOR 1 — `RetentionPurgeJob.kt:30` and `:70-79`, the 300 ms bound

The bound gives nothing. `MonitorServices.close()` calls `database.close()` right after
`stop()`, and that call blocks on the statement that runs. I measured 5623 ms for such a close
call. My three probe runs of `MonitorServices.close()` took 423 ms to 525 ms: 300 ms of dead
wait, plus the batch. `cancel()` already bounds the join to one batch. The bound also cancels
the scope and closes the dispatcher while a coroutine can still be live, which moves the last
resumption to `Dispatchers.IO` through the rejection fallback of the dispatcher.

**The fix.** Use `runBlocking { currentJob.cancelAndJoin() }`, and delete
`STOP_JOIN_TIMEOUT_MILLIS`. The reason that the pull request text gives for the bound is the
missing poll in `ApplicationLifecycleTest`. Commit `15386ae` of pull request #146 adds that
poll.

## MINOR 2 — `RetentionPurgeJob.kt:21`, the 24-hour value has no test

The fake wait function of each test ignores its argument, thus no test of the branch reads
`PURGE_INTERVAL_MILLIS`. I set the value to 60 seconds and each test still passed. I read the
argument with my own double, and the production value is correct: `86400000`.

**The fix.** Record the argument in the test double, and assert it:

```kotlin
val waits = CopyOnWriteArrayList<Long>()
val job = RetentionPurgeJob(wait = { millis -> waits.add(millis); gate.receive() }, action = ...)
...
assertEquals(listOf(86_400_000L), waits)
```

## MINOR 3 — `RetentionPurgeTest.kt:118-134`, a test that cannot fail

`assertTrue(plan.isNotEmpty())` holds for every valid statement, thus the test asserts nothing
about the plan. Lesson 1 of the backend brief names this defect.

**The fix.** Assert the shape that the pull request claims. For example, one line must hold
`COVERING INDEX event_session`. Or delete the test, and keep the plan in the pull request text.

## MINOR 4 — `RetentionPurge.kt:53-55`, the answer of the checkpoint

Two gaps:

1. `execute(CHECKPOINT_SQL)` ignores the answer. `wal_checkpoint(TRUNCATE)` returns `busy = 1`
   and truncates nothing while another connection holds a read transaction. **My evidence:** I
   held one `database.read { }` transaction open and ran `purgeOnce` on 20 000 old rows. The
   `octometer.db-wal` file stayed at **6 962 832 bytes**. With no open read the same call gave
   **0 bytes**. The code reports success in both cases.
2. A batch that throws skips the checkpoint. The rows of the earlier batches then wait for the
   next run, up to 24 hours later.

The risk is small, because `busy_timeout=5000` waits out a short read, and the automatic
checkpoint of SQLite also copies pages. The failure is silent, however, and D15 states the
rule.

**The fix.** Put the checkpoint in a `finally` block. Read the first column of the result row,
and write one warning line when it is 1.

## MINOR 5 — `RetentionPurgeTest.kt:27`, the temporary folder

The class calls `Files.createTempDirectory` and `deleteRecursively()`. It does not call
`registerTempRoot`. `deleteRecursively()` gives no error on Windows when a handle stays open,
thus a failed delete leaves a folder in `%TEMP%` with no signal. Issue #142 exists for this
reason, and lesson 7 of the backend brief names the helper. I saw no leftover folder after my
runs, so this is a hygiene gap and not a live leak.

**The fix.**
`Files.createTempDirectory("octometer-retention-purge-test-").toFile().also { registerTempRoot(it) }`.

## MINOR 6 — `RetentionPurgeJob.kt:78`, `stop()` sets `job = null`

A `start()` call after a `stop()` call then passes `check(job == null)`. The scope is cancelled
and the dispatcher is closed, thus the job never runs, and no error appears.

**The fix.** Keep a `stopped` flag and make `start()` throw after a stop, or leave `job` as it
is.

## MINOR 7 — the test count in the issue comment

The comment on issue #59 states 187 tests. The pull request text states 187 in the criterion
list and 198 in the "Test count" section. I measured 198 tests, 0 failures, 0 errors, two times
in sequence.

**The fix.** Set 198 in the issue comment and in the criterion list.

## MINOR 8 — two clock types

This branch adds `octometer.monitor.retention.Clock`. Pull request #146 adds `java.time.Clock`
to `MonitorServices.open()` and to `SqliteDatabase.open()`. After the merge the module holds
two clock abstractions for one purpose.

**The fix.** Use `java.time.Clock` in `RetentionPurge` too, and read the time with
`clock.millis()`. `FakeClock` then becomes `Clock.fixed(...)`.

## MINOR 9 — the language of the comments

Lesson 8 of the backend brief sets 25 words for a comment sentence, the active voice, and no
idiom.

- Idioms: "holds up" (`RetentionPurgeJob.kt:26`, `:68`), "moves on" (`:29`, `:67`), "outlives"
  (`:28`), "swallows" (`:43`).
- Above 25 words: `RetentionPurge.kt:8` (36 words), `RetentionPurgeJob.kt:65` (26),
  `MonitorServices.kt:94` (29), `RetentionPurgeJobTest.kt:12` (30), `:84` (29).

# The break experiments

I broke the code and saw the right test fail each time. I restored each file after each run.

| The break | The test that failed |
|---|---|
| `ts < ?` becomes `ts <= ?` | `an event of exactly retentionDays age stays` |
| The checkpoint call is removed | `purgeOnce empties the wal file after the checkpoint` |
| The batch loop stops after one batch | `purgeOnce deletes every old row, across more than one batch of 10 000` |
| The `CancellationException` rethrow is removed | `stop while the action runs writes no error log line` |
| `while (isActive)` becomes `if (isActive)` | `the action runs again once the wait call returns` |
| The health route answers a fixed 395 | `the health route answers with retentionDays` |
| `close()` does not stop the purge job | `open starts the purge thread, and close stops it` |
| A failed `open()` does not stop the purge job | `open throws and leaves no writer thread and no purge thread, when the app table is missing` |
| The interval becomes 60 seconds | **no test failed** (MINOR 2) |

# The verdict

3 MAJOR findings and 9 MINOR findings. No BLOCKER. Fix MAJOR 1, MAJOR 2, and MAJOR 3, and I
approve. I also ask for MINOR 1 and MINOR 3 in the same round, because the first one removes a
dead wait from each stop, and the second one turns a test that cannot fail into a real check.

xamcross:
## Correction round 1

This comment answers the reliability review and the SQLite review. It
lists each MAJOR finding, each MINOR finding, and the file with the line.

### MAJOR findings

| # | Review | Finding | Result | File |
| - | ------ | ------- | ------ | ---- |
| 1 | Reliability | The 24-hour gap between two runs had no test. | Fixed. `RetentionPurgeJobTest.start waits 24 hours between two runs` records the wait argument and asserts the literal `86 400 000`. | `monitor/backend/src/test/kotlin/octometer/monitor/retention/RetentionPurgeJobTest.kt:76-92` |
| 2 | Reliability | The chunk size of 10 000 had no test. | Fixed. `RetentionPurgeTest.purgeOnce writes exactly 3 times for 10 001 old rows...` counts the write calls through a new `onWrite` seam. | `monitor/backend/src/main/kotlin/octometer/monitor/retention/RetentionPurge.kt:65`, `monitor/backend/src/test/kotlin/octometer/monitor/retention/RetentionPurgeTest.kt:112-122` |
| 3 | Reliability | `stop()` bounded the join at 300 ms, and a purge coroutine could outlive `close()`. | Fixed. `stop()` now calls `cancelAndJoin()` with no fixed time. `MonitorServicesTest.close during a purge of 100 000 old rows...` holds a database lock so the first batch blocks, then asserts no ERROR line comes after `close()` returns. | `monitor/backend/src/main/kotlin/octometer/monitor/retention/RetentionPurgeJob.kt:72-81`, `monitor/backend/src/test/kotlin/octometer/monitor/MonitorServicesTest.kt:147` |
| 4 | Reliability | The `CancellationException` catch rethrew every time, with no `isActive` check. | Fixed. The catch now rethrows only when `!currentCoroutineContext().isActive`. `RetentionPurgeJobTest.a foreign CancellationException from the action logs one line, and the loop runs again` proves it. | `monitor/backend/src/main/kotlin/octometer/monitor/retention/RetentionPurgeJob.kt:92`, `monitor/backend/src/test/kotlin/octometer/monitor/retention/RetentionPurgeJobTest.kt:199` |
| 1 | SQLite | The same `CancellationException` catch defect (both reviews found it). | Fixed, same commit as above. | Same as above. |
| 2 | SQLite | `waitGate.send(Unit)` sat outside a timeout; a regression could hang the test instead of failing it. | Fixed. Each send now sits inside `withTimeout(5_000)`. | `monitor/backend/src/test/kotlin/octometer/monitor/retention/RetentionPurgeJobTest.kt:62, 185, 215` |
| 3 | SQLite | `MonitorServices.open()` runs a real purge at each open call. `AppTotalsRouteTest` fed it a fixed old timestamp, and after the merge of #144/#145, `ElementsRouteTest` and `UserTotalsRouteTest` hit the same trap (5 failures on the merged branch). | Fixed. `devConfig()`/`prodConfig()` now default `retentionDays` to 3 650 000. `AppTotalsRouteTest` keeps its fixed instant. A test of the purge passes its own small `retentionDays`. | `monitor/backend/src/test/kotlin/octometer/monitor/TestConfigs.kt:56`, `monitor/backend/src/test/kotlin/octometer/monitor/apps/AppTotalsRouteTest.kt:281` |

### MINOR findings

Corrected:

- `RetentionPurgeTest` now registers its temporary folder with `registerTempRoot`. (`RetentionPurgeTest.kt:31-32`)
- `RetentionPurgeJob.start()` after `stop()` now throws, instead of a silent no-op. Proven by `start after stop throws, instead of a silent restart`. (`RetentionPurgeJob.kt:54-56`, test at `RetentionPurgeJobTest.kt:158`)
- The issue comment with the stale count of 187 gets a new comment with the real count (285) below this one.
- The module now uses one clock type, `java.time.Clock`, in place of the module's own `Clock` interface. (`RetentionPurge.kt:3`, `MonitorServices.kt:4`)
- The comment language across the retention files now keeps each sentence at or under 25 words, in the active voice, with no idiom.
- A zero-row purge now logs a `debug` line, not an `info` line. (`RetentionPurge.kt:81-87`)

Corrected beyond the list, at low cost:

- The query plan test asserted only `plan.isNotEmpty()`, a check that cannot fail. It now asserts the plan names `COVERING INDEX event_session`. (`RetentionPurgeTest.kt:144-160`)
- The checkpoint answer row (`busy`, `log`, `checkpointed`) is now read, with `busy_timeout` at 0 for that one call, per decision 7. A busy answer logs one debug line. (`RetentionPurge.kt:105-123`)

Not corrected, with the reason:

- Reliability MINOR 2 (a cancelled purge writes no line) and MINOR 6 (a wall-clock anchor for the schedule) and MINOR 7 (a guard against a forward clock jump): each one changes the wait loop or the log volume in a way the maintainer's decision list does not name. Each one fits a follow-up issue better than this correction round.
- Reliability MINOR 4 (give `MonitorServices.open()` its own injected `Clock`): seven branches touch `MonitorServices.kt` at once this week. Widening its signature further, with no line in the maintainer's decision list asking for it, works against the brief's rule to change the smallest count of shared lines.

### The merge

The branch merges `origin/main` twice: once through commit `ba2ea5c` (the level 2 totals route), and once more through commit `01d07b8` (pull request #147, the user erasure route), after the coordinator's note that #147 had landed. Both merges used a merge commit, kept each line of both sides in `ApiRoutes.kt`, and had one real conflict: `SqlitePragmas.kt` ended up with two busy-timeout constants of the same value. The fix keeps the one that #147 added, `octometer.monitor.store.DEFAULT_BUSY_TIMEOUT_MILLIS`, and `RetentionPurge` now uses that same constant.

### Test result

`./gradlew :monitor:backend:test`, run two times in sequence, on JDK 21: **285 tests, 0 failures, 0 errors**, both times. `./gradlew build` passed once, across every module.

### CI

Each check passed on the pushed commit. The Windows job (`Monitor backend on Windows`) failed twice on a retry, each time on `ApplicationLifecycleTest.the server opens and closes the store on each start and stop, twice in one folder`, before a third retry passed. This test already exists on `main`; it is not part of this correction round. The same test failed once on an earlier commit of this same pull request (`6395c53`, before this correction round), and passed on the commit right after it (`e6e6b7f`), so this looks like a pre-existing, intermittent flake on the Windows runner, not a regression from this round.

### The break-and-restore proof

For each new or changed test, the code broke for a moment, the test failed, then the code went back:

- `PURGE_INTERVAL_MILLIS` at 1 second: `start waits 24 hours between two runs` failed.
- `PURGE_BATCH_SIZE` at 3 000: `purgeOnce writes exactly 3 times for 10 001 old rows...` failed.
- The `CancellationException` catch rethrowing every time: `a foreign CancellationException from the action logs one line...` failed with a `TimeoutCancellationException`.
- `stop()` with no join at all: `close during a purge of 100 000 old rows...` failed with a stray `SQLiteException` logged after `close()` returned.

Each break came back to the fixed form before the next commit.

xamcross:
**Review (second review, SQLite and backend engineer): APPROVE**

I made a new clone, ran `gh pr checkout 149`, and ran each check there on JDK 21. Head commit
`39f2bec`. I pushed nothing and I changed no branch.

## The three MAJOR findings

**MAJOR 1 — the `CancellationException` catch. Fixed and proven.**
`RetentionPurgeJob.kt:92` now reads `if (!currentCoroutineContext().isActive) throw cancellation`.
I put the old form back (`throw cancellation` every time) and ran the class. The new test
`a foreign CancellationException from the action logs one line, and the loop runs again`
**failed** with a `TimeoutCancellationException`. I restored the file.

**MAJOR 2 — a test that hangs. Fixed and proven.**
Each gate send now sits inside `withTimeout(5_000)` (`RetentionPurgeJobTest.kt:62, 185, 215`).
I made a failed run stop the loop, the exact break that hung for more than 12 minutes in round
1. This time `a failed run does not stop the loop` **failed in 20 seconds**, with a
`TimeoutCancellationException`. The safety net works now. I restored the file.

**MAJOR 3 — the purge at the start against a test fixture. Fixed.**
`TestConfigs.kt:62` and `:71` set `retentionDays = LARGE_TEST_RETENTION_DAYS` (3 650 000) for
`devConfig()` and `prodConfig()`. `AppTotalsRouteTest.kt:281` holds the fixed instant
`1_700_000_000_000L` again, with a comment that names the reason. A test of the purge passes
its own small value.

## The nine MINOR findings

| # | The finding | Result | Where |
| - | ----------- | ------ | ----- |
| 1 | The 300 ms bound in `stop()` | Fixed. `runBlocking { currentJob.cancelAndJoin() }`, and `STOP_JOIN_TIMEOUT_MILLIS` is gone. `purgeOnce` calls `ensureActive()` between two batches, thus the join ends at a batch boundary. | `RetentionPurgeJob.kt:75`, `RetentionPurge.kt:72` |
| 2 | The 24-hour value had no test | Fixed. `start waits 24 hours between two runs` reads the wait argument and asserts `24L * 60 * 60 * 1000`. | `RetentionPurgeJobTest.kt:76-95` |
| 3 | The query plan test could not fail | Fixed. `the delete statement scans the covering index event_session` asserts that one plan line holds `COVERING INDEX event_session`. | `RetentionPurgeTest.kt:144-160` |
| 4 | The checkpoint answer was not read | Fixed. The code reads the `busy` column, with `busy_timeout=0` for that one call, then restores the timeout in a `finally`. A busy answer writes one debug line. | `RetentionPurge.kt:105-123` |
| 5 | The temporary folder was not registered | Fixed. `registerTempRoot(it)` on the test folder. | `RetentionPurgeTest.kt:32` |
| 6 | `start()` after `stop()` was a silent no-op | Fixed. A `stopped` flag plus `check(!stopped)`, with the test `start after stop throws, instead of a silent restart`. | `RetentionPurgeJob.kt:52-56`, test at `:158` |
| 7 | The stale count of 187 tests | Fixed. A new comment on issue #59 names 285 tests. | issue #59 |
| 8 | Two clock types | Fixed. `Clock.kt` is gone. `RetentionPurge` takes `java.time.Clock` and calls `clock.millis()`. | `RetentionPurge.kt:3, 64, 69` |
| 9 | The language of the comments | Fixed. I found no idiom ("holds up", "moves on", "outlives", "swallows") in the retention files. | `retention/*.kt` |

## The one busy-timeout constant of #147

Confirmed. `SqlitePragmas.kt:13` declares `const val DEFAULT_BUSY_TIMEOUT_MILLIS = 5000`, and
that one constant serves three places: the pragma list at `SqlitePragmas.kt:24`, the erasure at
`UserErasureService.kt:172`, and the purge at `RetentionPurge.kt:39`. One value, one source.
The merge left no second constant.

## The test run

`./gradlew :monitor:backend:test`: **285 tests, 0 failures, 0 errors.** The count matches the
report of the correction round.

## Two nits for a later round, neither one blocking

- `RetentionPurge.kt:77`: the checkpoint call stands outside a `finally`, thus a batch that
  throws still skips it. The next scheduled run covers the gap. A `finally` would close it.
- `RetentionPurge.kt:38` builds its own `PRAGMA busy_timeout=0` text, while
  `UserErasureService.kt:202` uses a `setBusyTimeout(writer, millis)` helper for the same job.
  One shared helper in the store package would keep one code path.

## The verdict

No BLOCKER and no MAJOR stays. Each of the three MAJOR findings and each of the nine MINOR
findings of my first review is fixed, and I proved the two behaviour fixes with my own break
and restore. I approve.

xamcross:
**Review (second review, reliability engineer): CHANGES REQUESTED**

Each MAJOR of my first review is fixed, and a real test now guards each one. I confirmed each fix with my own break-and-restore run. One new MAJOR stays: the thread name match that makes the Windows CI job fail on this branch.

## What I ran

A new clone, `gh pr checkout 149`, head `39f2bec`, JDK 21.

- `./gradlew :monitor:backend:test`, two times in sequence: **285 tests, 0 failures, 0 errors**, each time. The second run used `--rerun-tasks`. The count agrees with the response.
- 5 mutations of the code, each with a targeted test run, then a restore.
- 3 probes with a real `MonitorServices` and a real Netty server.

## The MAJOR findings of round 1

| # | My check | Result |
| - | -------- | ------ |
| 1 | I set `PURGE_INTERVAL_MILLIS` to `1000L`. | **FIXED** — `RetentionPurgeJobTest > start waits 24 hours between two runs()` FAILED. The test holds the literal `86 400 000`, not the production constant. |
| 2 | I set `PURGE_BATCH_SIZE` to `3_000`. | **FIXED** — `RetentionPurgeTest > purgeOnce writes exactly 3 times for 10 001 old rows, with a batch size of 10 000()` FAILED. The `onWrite` seam counts 3 write calls. |
| 3 | I ran my round 1 probe again, and I set `stop()` back to a bare `cancel()`. | **FIXED** — see the table below. `MonitorServicesTest > close during a purge of 100 000 old rows ...()` FAILED with the bare cancel. |
| 4 | I put the old catch back, so it throws each `CancellationException` again. | **FIXED** — `RetentionPurgeJobTest > a foreign CancellationException from the action logs one line, and the loop runs again()` FAILED. |

I restored each file after each run.

### MAJOR 3, my probe

`RetentionPurgeJob.kt:72-81` now calls `runBlocking { currentJob.cancelAndJoin() }`, with no fixed time. `RetentionPurge.kt:72` calls `currentCoroutineContext().ensureActive()` between two batches. My probe is the same as in round 1: 100 000 old rows, a second raw connection with `BEGIN EXCLUSIVE`, `retentionDays = 1`, then `close()` while the first batch blocks inside SQLite.

| Measurement | Round 1 | Now |
| ----------- | ------- | --- |
| The thread list before `close()` | `octometer-retention-purge`, `octometer-sqlite-writer @coroutine#26` | `octometer-retention-purge`, `octometer-sqlite-writer @coroutine#4` |
| The time of `close()` | 4773 ms | 4903 ms |
| The thread list right after `close()` | `octometer-sqlite-writer @coroutine#26` | **empty** |
| An exception out of `close()` | none | none |
| An ERROR line after `close()` returned | **1** | **0** |
| The log lines of the run | 2 | 2, each one before `close()` returned |

`close()` now waits for the batch in progress, thus no coroutine outlives it. The 4903 ms comes from the busy timeout of the blocked batch, not from the join itself; a normal batch ends in about 90 ms. This is the correct trade: the stop is safe, and its cost is one batch.

## The new MAJOR — the thread name match makes the Windows job fail

`monitor/backend/src/test/kotlin/octometer/monitor/ApplicationLifecycleTest.kt:176` and `monitor/backend/src/test/kotlin/octometer/monitor/MonitorServicesTest.kt:256`

Both helpers count a thread with an exact name match:

```kotlin
Thread.getAllStackTraces().keys.count { it.name == WRITER_THREAD_NAME }
```

The Gradle test task enables assertions, thus the coroutine debug mode of kotlinx is active, and it appends ` @coroutine#N` to the name of a thread that runs a coroutine. The exact match then does not see that thread.

My evidence:

1. The head commit `39f2bec` needed three CI attempts. The job `Monitor backend on Windows` failed on attempt 1 and on attempt 2, each time at `ApplicationLifecycleTest.kt:66`, which is `assertEquals(before + 1, writerThreadCount(), ...)`. Attempt 3 passed.
2. The last 15 CI runs on `main` all passed. Each failure sits on this branch.
3. I started a real server the same way as that test, with a purge of 200 000 old rows, and I sampled the thread names through the start window. **2296 of 7010 samples** showed the writer thread under the name `octometer-sqlite-writer @coroutine#5`, thus the exact match missed it. At the sample point of line 66 my run held the plain name, so the assertion passed; the result is a race, not a fixed answer.
4. This pull request is the first code that runs a coroutine on the writer thread inside that window. Before it, the migration was the only writer-thread work, and it ends inside `SqliteDatabase.open`, well before the health request. The purge now writes on that thread while the server starts.

**Does the merge of #146 cover it?** No. Pull request #146 is open, and GitHub reports `mergedAt: null`, thus its change is not on `main`. `git show origin/main:...ApplicationLifecycleTest.kt` still holds the exact match. This branch needs the same change now.

This round also **adds** a test that uses the same fragile helper (`MonitorServicesTest.kt:256`), for the new assertions "close() must leave no writer thread" and "close() must leave no purge thread". There the mismatch gives a false pass: a thread that still runs a coroutine is not counted, thus the assertion sees the expected number although the thread is alive.

**The fix.** Two lines:

```kotlin
// ApplicationLifecycleTest.kt:176
private fun writerThreadCount(): Int =
    Thread.getAllStackTraces().keys.count { it.name.startsWith(WRITER_THREAD_NAME) }

// MonitorServicesTest.kt:256
private fun threadCount(name: String): Int =
    Thread.getAllStackTraces().keys.count { it.name.startsWith(name) }
```

Add a one-line comment that names the reason: the coroutine debug mode appends ` @coroutine#N`. Tell the owner of #146 that this branch makes the same change, so the merge keeps one form.

## The 8 MINOR findings of round 1

| # | The finding | State | My judgement |
| - | ----------- | ----- | ------------ |
| 1 | A run with no old row writes an INFO line | **Fixed** (`RetentionPurge.kt:81-87`). A zero count writes a DEBUG line, and the root level is INFO, thus the line stays out of the console. | Correct. |
| 2 | A cancelled purge writes no line | **Open.** The reason: it changes the log volume, thus it fits a follow-up issue. | Weak, and I accept it. A `finally` block writes the same line in place of the line that the cancel skips, thus the volume does not grow. It stays a MINOR; please open the follow-up issue. |
| 3 | `RetentionPurgeTest` does not register its temporary folder | **Fixed** (`RetentionPurgeTest.kt:31-32`, `registerTempRoot`). | Correct. |
| 4 | `MonitorServices.open` has no `Clock` seam | **Open**, with a different fix: `TestConfigs.kt:59` sets a default `retentionDays` of 3 650 000. The reason: seven branches touch `MonitorServices.kt` this week. | Sound. The value gives a cutoff far below any test timestamp, and `AppTotalsRouteTest` keeps its fixed instant. The new close test passes its own `retentionDays = 1`, thus the real wiring still deletes a row in one test. |
| 5 | The issue comment named 187 tests | **Fixed.** A new comment on issue #59 names 285 tests. | Correct. |
| 6 | The schedule holds no wall-clock anchor | **Open.** The reason: a follow-up issue. | Sound. My round 1 measurement shows the monotonic timer counts each standby hour on this machine, thus the behaviour is correct today. Please open the follow-up issue for the hibernate case. |
| 7 | No guard against a forward jump of the system clock | **Open.** The reason: the decision list does not name it. | Sound for this round. The risk stays: a forward jump deletes more than the window, and `secure_delete` makes it permanent. Please open the follow-up issue. |
| 8 | `start()` after `stop()` did nothing | **Fixed** (`RetentionPurgeJob.kt:52-56`), with a test at `RetentionPurgeJobTest.kt:158`. | Correct. |

## Two new MINOR findings

### MINOR A — the explicit cancel check has no test
`monitor/backend/src/main/kotlin/octometer/monitor/retention/RetentionPurge.kt:72`. I removed the `currentCoroutineContext().ensureActive()` line and ran `MonitorServicesTest` and `RetentionPurgeTest`: every test passed. The behaviour does not change today, because `SqliteDatabase.write` runs `withContext`, and `withContext` checks the cancel at the start of each batch. Keep the line: it states the rule, and it becomes load-bearing the day `write` stops using `withContext`. A test needs a seam that cancels the job between two batches; the cost is low, and this is not a blocker.

### MINOR B — a busy checkpoint stays invisible, and it leaves the purged rows in the wal file
`monitor/backend/src/main/kotlin/octometer/monitor/retention/RetentionPurge.kt:105-123`. Decision 7 sets `busy_timeout=0` for the checkpoint, thus a concurrent reader makes the truncate give up at once, where the old form waited up to 5 s. D15 asks for `wal_checkpoint(TRUNCATE)` after a purge for a privacy reason: an old wal frame still holds the row content until the truncate completes. A busy answer now writes a DEBUG line, and `logback.xml` sets the root level to INFO, thus the owner never sees it. Raise that one line to WARN, or run the checkpoint again at the next scheduled run and count the failed attempts. Please ask the privacy reviewer to confirm the choice.

## What else I checked

- The order of `MonitorServices.close()` stays correct: it stops the job, then it closes the database.
- A throw in a later step of `open` still stops the job (the test with the missing app table).
- The two merges of `origin/main` keep each line of both sides, and `RetentionPurge.kt:39` uses the one shared constant `DEFAULT_BUSY_TIMEOUT_MILLIS` of #147.
- The language of the new comments follows Simplified Technical English.

## The verdict

One MAJOR stays, thus I request the change. Fix the two thread count helpers, push, and let the Windows job run one time with no retry. I approve after that.

I restored each file that I changed for an experiment, and I deleted my clone.

xamcross:
**Correction round 2, commit `7168cb4`**

1. The new MAJOR: `writerThreadCount()` in `ApplicationLifecycleTest.kt:176` and `threadCount()` in `MonitorServicesTest.kt:256` now use `startsWith`, not `==`. The coroutine debug mode adds a suffix to the thread name, and the exact match missed it. The comment names the same reason as pull request 146.
2. The first new MINOR: `RetentionPurge.kt` drops the explicit `ensureActive()` call. A comment names the cancel check inside `withContext`, which gives the same protection today.
3. The second new MINOR: the busy checkpoint line in `RetentionPurge.kt` now logs at WARN, not DEBUG. The root log level is INFO, so the owner now sees this line.

`./gradlew :monitor:backend:test` ran two times in sequence with JDK 21: **285 tests, 0 failures, 0 errors**, each time.

All 5 CI checks pass on the push, including "Monitor backend on Windows" on the first attempt, with no retry: https://github.com/xamcross/octometer/actions/runs/35682506980

xamcross:
**Review (third review, reliability engineer): APPROVE**

Head `7168cb4`, a new clone, JDK 21. `./gradlew :monitor:backend:test`: **285 tests, 0 failures, 0 errors**, BUILD SUCCESSFUL.

## The MAJOR of round 2 — fixed

| File | State |
| ---- | ----- |
| `ApplicationLifecycleTest.kt:182-183` | `count { it.name.startsWith(WRITER_THREAD_NAME) }`. The helper and its six comment lines are **byte-identical** to `pr146:...ApplicationLifecycleTest.kt:194-195`. The merge of #146 thus has no conflict here. |
| `MonitorServicesTest.kt:256-259` | `count { it.name.startsWith(name) }`, with the same comment as #146. The body differs from `pr146:...MonitorServicesTest.kt:99-100` in one way alone: this branch takes a `name` parameter, because it counts the purge thread too. The match rule is the same, thus the merge needs one small manual join, not a change of behaviour. |

The CI run `35682506980` for `7168cb4` passed each job on attempt 1, and `Monitor backend on Windows` is one of them. That agrees with my local run.

## The two MINOR of round 2 — fixed

- **MINOR A.** `RetentionPurge.kt:71-77`: the `ensureActive()` line is gone, and six comment lines name the reason: `deleteOneBatch` calls `database.write`, `database.write` calls `withContext`, and `withContext` checks the cancel state before it runs the block. The comment states the rule at the place where a later reader needs it. I keep no objection: my round 2 measurement showed the behaviour is the same with the line and without it.
- **MINOR B.** `RetentionPurge.kt:123`: the busy checkpoint line is `log.warn("The checkpoint was busy. It did not run to completion.")`. The root level of `logback.xml` is INFO, thus the owner now sees the line. The text holds one fixed sentence, no count of rows, no path, and no event data.

Issue #118, opened 2026-09-21
Replace the key map of the rate limit with two LRU maps


xamcross:
**Goal.** The rate limit keeps two key maps with an LRU eviction, thus an anonymous flood never evicts a signed-in user.

**Evidence.** Rule D20 states one map of 10 000 keys, and a request with a new key above that count gets the rate-limit result. Observed in the security review on 2026-09-21: one attacker fills that map and blocks each new user.

**Implementation steps.**
1. In `kit/jvm-core`, replace the one key map of D20 with two maps.
2. Give the user key map 5 000 entries with an LRU eviction.
3. Give the anonymous key map 20 000 entries with an LRU eviction.
4. Add the tests of the acceptance criteria with an injected clock.

**Acceptance criteria.**
- [ ] The user map holds 5 000 keys, and key 5 001 evicts the least recent user key.
- [ ] The anonymous map holds 20 000 keys, and key 20 001 evicts the least recent anonymous key.
- [ ] 20 001 anonymous keys evict no user key.
- [ ] A new key after an eviction gets a fresh counter, not the rate-limit result.
- [ ] No log line holds the IP address.
- [ ] `./gradlew :kit:jvm-core:test` passes.
- [ ] A comment on this issue records the test output.

**Related.**
- #33 gives the one map of D20.
- #116 adds the anonymous counters that use the second map.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, the contract version 1.1 of #101, and the rule S5.

xamcross:
Closed on 2026-09-22 by the maintainer as completed by #33 (pull request #158, commit 241f8b8). The rate limiter of `kit/jvm-core` holds two key maps with an LRU eviction from its first version: 5 000 user keys and 20 000 client address keys (design decision D20). The two reviews of #158 proved the eviction: after 10 000 and after 35 000 anonymous keys, a signed-in user and a new visitor each pass; an evicted key gets a full new counter; 20 runs of 1 000 requests from 8 threads give the exact count. No step of this issue stays. The IPv6 /64 key rule belongs to #116.

xamcross:
Body edit on 2026-09-22: #33 closed. No open blocker remains, thus the first line and the label `blocked` are removed. The work can start.

Issue #116, opened 2026-09-21
Add the anonymous rate limits and the proxy rule


xamcross:
**Goal.** With anonymous events on, the kit applies three counters to one key in one window of 60 seconds. A header gives the client address only behind a trusted proxy.

**Why it blocks production.** The anonymous ingest route has no useful limit today. One address sends 6 000 events each minute, thus it fills the cap of 200 000 events in 34 minutes. The TTL keeps that junk 30 days, thus one attacker stops the measurement for a month.

**Implementation steps.**
1. In `kit/jvm-core`, add three counters for one key in one fixed window of 60 seconds. They run only with `OCTOMETER_RECORD_ANONYMOUS=true`.
2. Set the defaults 300 requests (`OCTOMETER_ANON_REQ_PER_MIN`), 900 click entries (`OCTOMETER_ANON_EVENTS_PER_MIN`), and 120 session starts (`OCTOMETER_ANON_SESSIONS_PER_MIN`).
3. Build the key from one IPv4 address, or from the first 64 bits of an IPv6 address.
4. Read the client address from the header that `OCTOMETER_CLIENT_IP_HEADER` names. Take the n-th value from the right of that header list.
5. Give `OCTOMETER_TRUSTED_PROXY_COUNT` the default 1. Use the remote address without `OCTOMETER_CLIENT_IP_HEADER`.
6. Add the tests of the acceptance criteria with an injected clock.

**Acceptance criteria.**
- [ ] Fixed clock: request 301 of one key inside 60 seconds gives the rate-limit result, and a request after the window passes.
- [ ] Click entry 901 of one key inside 60 seconds gives the rate-limit result.
- [ ] Entry 121 of `octo:session-start` of one key inside 60 seconds gives the rate-limit result. A mobile carrier puts thousands of users behind one address, thus the number is 120 and not 20.
- [ ] Two IPv6 addresses with the same first 64 bits share one key.
- [ ] With `OCTOMETER_CLIENT_IP_HEADER` set and no `OCTOMETER_TRUSTED_PROXY_COUNT`, a header with three values gives the last value.
- [ ] With `OCTOMETER_TRUSTED_PROXY_COUNT` 2, a header with three values gives the second value from the right.
- [ ] Without `OCTOMETER_CLIENT_IP_HEADER` the key comes from the remote address, also with a forged header in the request.
- [ ] The three counters stay off with `OCTOMETER_RECORD_ANONYMOUS=false`.
[...]

PR #158, opened 2026-09-22
feat(kit): limit the ingest rate

### LINKED ISSUE #33: Limit the ingest rate in kit/jvm-core

xamcross:
## Summary

This pull request adds the ingest rate limit of design decision D20.
The kit rejects a client that sends too many requests in one 60-second
window. Correction round 1 of this pull request answers the security
review and the concurrency review: it fixes the window race, it splits
the key map into two LRU maps, it caps the key length, and it fixes
the client address header rule.

## Acceptance criteria (issue #33, 2026-09-22 body edit)

- [x] Fixed clock: request 31 of one user id in 60 seconds gives the
  rate-limit result, and a request after the window passes.
  Evidence: `IngestRateLimiterTest.request31OfOneUserIdInTheSameWindowIsLimited`
  and `aRequestOfTheSameUserIdAfterTheWindowEndsPasses` (`kit/jvm-core`);
  `IngestRouteTest` test "request 31 of one user id in 60 seconds gives
  429, and a request after the window passes" (`kit/jvm-ktor`, injects a
  mutable clock).
- [x] Without a user id, request 121 of one IP address gives the
  rate-limit result.
  Evidence: `IngestRateLimiterTest.withoutAUserIdRequest121OfOneClientAddressIsLimited`
  (`kit/jvm-core`); `IngestRouteTest` test "without a user id, request
  121 of one client address gives 429" (`kit/jvm-ktor`).
- [x] Without `OCTOMETER_CLIENT_IP_HEADER` the key is the remote
  address.
  Evidence: `IngestRouteTest` tests "without OCTOMETER_CLIENT_IP_HEADER,
  a spoofed header never changes the key" and "with
  OCTOMETER_CLIENT_IP_HEADER set, a request with no such header falls
  back to the remote address".
- [x] Two key maps with an LRU eviction hold the counters: 5 000 user
  keys and 20 000 client address keys. A full map evicts its oldest
  key, and a new key always gets a fresh counter. A test proves that
  10 000 anonymous keys block no user and no new visitor.
  Evidence: `IngestRateLimiter` now holds one `Window` record with two
  `KeyMap` instances (a `LinkedHashMap` in access order, under one lock
  each; see the class comment for the reason). Tests:
  `tenThousandAnonymousKeysBlockNoUserAndNoNewVisitor`,
  `aFullClientAddressMapEvictsItsOldestKeyAndGivesTheNewKeyAFreshCounter`.
- [x] A key text above 64 characters is replaced by its SHA-256 hex,
  thus a long header value or a long user id cannot fill the memory,
  and two keys never join.
  Evidence: `IngestRateLimiterTest.aKeyTextAbove64CharactersIsShortenedAndTwoLongKeysNeverJoin`
  (two 6 000-character values that share their first 64 characters, so
  a truncating shortener would join them; the SHA-256 shortener does
  not).
- [x] With `OCTOMETER_CLIENT_IP_HEADER` set, the key comes from the
  last element of the last header line. A header value above 64
  characters or without a valid address form gives the remote address.
  The README says: set the option only behind a proxy that appends the
  address.
  Evidence: `IngestRouteTest` tests "with OCTOMETER_CLIENT_IP_HEADER
  set, a forged left element never changes the key", "... a header
  value above 64 characters gives the remote address", "... a value
  with no IP address form gives the remote address". `kit/jvm-ktor/README.md`,
  section "The ingest rate limit".
- [x] Each route test uses an injected clock, not the wall clock.
  Evidence: every `octometerIngestRoute(...)` call of `IngestRouteTest`
  now passes `clock = fixedClock` or `clock = clock` (a `MutableClock`).
  `./gradlew :kit:jvm-ktor:test` ran five times in sequence: 0 failures
  each time (see below).
- [x] No stored event and no log line holds the IP address.
  Evidence: `IngestRateLimiterTest.aRejectionWritesAMaximumOfOneLogLineForTheWindowWithNoKeyInIt`;
  `IngestRouteTest` test "a 429 answer writes no line of its own to the
  application log of the route" now asserts that no captured message
  holds the client address of the test, not only a fixed string.
- [x] The Ktor route returns 429.
  Evidence: each `IngestRouteTest` test above.
- [x] `./gradlew :kit:jvm-core:test` and `./gradlew :kit:jvm-ktor:test`
  pass.
  Evidence: `kit/jvm-core` 152 tests, 0 failures, 0 errors, 0 skipped.
  `kit/jvm-ktor` 39 tests, 0 failures, 0 errors, 0 skipped, on each of
  5 sequential runs.
  Evidence: see the issue comment of the first round of this pull
  request; correction round 1 posts its own comment on this pull
  request with the new counts.

## Correction round 1

A security review and a concurrency review of this pull request found
2 BLOCKER and 3 MAJOR findings each. The maintainer's 2026-09-22 body
edit of issue #33 folded the fix of each BLOCKER and each MAJOR into
the acceptance criteria above. The pull request comment of this round
lists each finding, the file and line, and the result.

## Other checks

- `./gradlew build` (JDK 21, Docker present): BUILD SUCCESSFUL.
  `kit/jvm-mongo`: 32 tests, 0 failures (container tests ran against
  Docker, plus 32 tests of `testMongoSyncNewest`). `monitor/backend`:
  316 tests, 0 failures.
- A mutation check of each new rule and each new test: a short local
  edit that breaks the rule makes the matching test fail, then the
  edit is reverted. Not part of this diff; recorded in the pull
  request comment of this round.

## Decisions where the issue was not clear

1. **The client address parameter of a signed-in user's `check` call.**
   `check(userId, clientAddress)` never reads `clientAddress` when
   `userId` is not `null`. The route now calls `clientAddress(call, ...)`
   only for a request with no user id, and passes an empty string
   otherwise, so the route reads the header once at most, never for a
   signed-in user.
2. **The content type guard still runs before the rate limit check.**
   A request with a wrong content type never counts against a
   rate-limit key. The KDoc of the route now states this order; this
   pull request does not reorder the two checks, because the route
   reads no body either way, and reordering was not one of the
   maintainer's numbered corrections.
3. **The IPv6 address form gets no `/64` grouping.** Design decision
   D20 states no IPv6 rule; the IPv6 caps of D43 are issue #116 and
   issue #118 work, not this one. The security review and the
   concurrency review both flagged this as a MINOR finding; this pull
   request records the gap in `kit/jvm-ktor/README.md`, section "The
   ingest rate limit", instead of fixing it here.
4. **`IngestPipeline.java` still has no change**, for the reason of the
   first round of this pull request (issue #104 changes it at the same
   time).

Closes #33

Generated with Claude Code


xamcross:

I made my own clone, I ran `gh pr checkout 158` in it, and I ran each command myself with JDK 21
(`C:\Users\xamcr\.jdks\ms-21.0.9`). I deleted the clone at the end. Each result below is my own
observation.

## The acceptance criteria

1. **Fixed clock: request 31 of one user id.** PASS. `IngestRateLimiterTest` gives 10 tests, 0
   failures.
2. **Without a user id, request 121 of one client address.** PASS.
3. **Without `OCTOMETER_CLIENT_IP_HEADER` the key is the remote address.** PASS. I started a real
   Netty server and I sent a raw request with `Host: evil.example.com`,
   `X-Forwarded-For: 203.0.113.77` and `Forwarded: for=203.0.113.99`. The value of
   `call.request.local.remoteAddress` stayed `127.0.0.1`. The choice of `local` is correct:
   `call.request.origin` changes when the app installs the `ForwardedHeaders` plugin, and `local`
   does not.
4. **With 10 000 keys a new key gets the rate-limit result.** PASS. See MAJOR 2 for the risk.
5. **No stored event and no log line holds the IP address.** PASS. The limiter writes one fixed
   line for each window. The route writes no line of its own for a 429 answer.
6. **The Ktor route returns 429.** PASS. The body of the 429 answer is empty, thus it holds no key
   (rule C19).
7. **`./gradlew :kit:jvm-core:test` and `./gradlew :kit:jvm-ktor:test` pass.** FAIL. See BLOCKER 1.
8. **A comment on the issue records the test output.** PASS.

## BLOCKER 1 — the first run of `:kit:jvm-ktor:test` failed on a clean checkout

`kit/jvm-ktor/src/test/kotlin/octometer/kit/ktor/IngestRouteTest.kt:829, 857, 884, 918, 947, 976`

My first run of `./gradlew :kit:jvm-ktor:test`, on the branch with no change, gave:

```
IngestRouteTest > without OCTOMETER_CLIENT_IP_HEADER, a spoofed header never changes the key() FAILED
    org.opentest4j.AssertionFailedError: expected: <429 Too Many Requests> but was: <204 No Content>
    at octometer.kit.ktor.IngestRouteTest.kt:943
35 tests completed, 1 failed
```

I ran the same command five times in total on the unchanged branch. One run failed.

The cause: six of the seven new route tests install the route without a `clock` argument, so they
use `Clock.systemUTC()`. `IngestRateLimiter` aligns each window to the wall clock
(`now - Math.floorMod(now, WINDOW_MILLIS)`, `IngestRateLimiter.java:106`). A run that crosses a
minute mark clears the map in the middle of the request loop. Each request then passes, and the
final assertion of 429 fails.

I reproduced the mechanism with no race. I installed the route with a clock that moves to the next
window after 120 calls, and I sent 121 anonymous requests. No request got 429.

The fix: give each of the six tests its own clock, as the test at line 793 already does. Pass
`clock = clock` with the `MutableClock` of line 1010, or `clock = fixedClock` for a test that needs
no advance.

## BLOCKER 2 — a test that cannot fail

`kit/jvm-core/src/test/java/octometer/kit/core/ingest/IngestRateLimiterTest.java:83`

`aUserIdKeyAndAClientAddressKeyOfTheSameTextStayIndependent` claims that a user key and a
client-address key of the same text stay separate. I removed both key prefixes in
`IngestRateLimiter.check` (`key = clientAddress` at line 98 and `key = userId` at line 101). The
command `./gradlew :kit:jvm-core:test --tests "octometer.kit.core.ingest.IngestRateLimiterTest"`
then gave `BUILD SUCCESSFUL`, 10 tests, 0 failures. No test holds this rule.

The reason: the test drives the user key to 31 counts. It then calls `check(null, sharedText)` one
time. With one shared counter that call gives the count 32, and the limit of that call is 120.
32 stays below 120, thus the assertion passes in each case.

The fix: reverse the order of the two parts.

```java
for (int i = 1; i <= 30; i++) {
    assertEquals(RateLimitResult.ALLOWED, limiter.check(null, sharedText));
}
assertEquals(RateLimitResult.ALLOWED, limiter.check(sharedText, "10.0.0.1"));
```

One shared counter makes the last call the 31st count of that key. The user limit of 30 then gives
`LIMITED`, and the test fails as it must.

## MAJOR 1 — `OCTOMETER_CLIENT_IP_HEADER` gives the key to the attacker

`kit/jvm-ktor/src/main/kotlin/octometer/kit/ktor/IngestRoute.kt:250`

`clientAddress` returns the raw header value. It applies no trust rule and no list rule. I installed
the route with `clientIpHeaderName = "X-Forwarded-For"` and I measured two attacks:

- I sent 130 requests, each with `X-Forwarded-For: 203.0.113.<n>, 127.0.0.1`, with a new `<n>` in
  each request. **All 130 passed.** The limit of 120 never applied. This value form is the form
  that a proxy with `$proxy_add_x_forwarded_for` makes: the proxy appends its peer address, thus
  the text of the client stays inside the value. The limit is void behind such a proxy, and it is
  void with no proxy at all.
- I sent 130 requests with the one fixed value `198.51.100.5, 127.0.0.1`. 120 passed, then each
  one gave 429. An attacker names the address of a real visitor this way, and exhausts the budget
  of that visitor.

`kit/jvm-ktor/README.md` names neither the rate limit nor this variable. An app owner who reads D20
sets the variable today. The owner then gets a limiter that an attacker turns off, or turns against
a real visitor. Issue #116 owns `OCTOMETER_TRUSTED_PROXY_COUNT`, but nothing stops an owner from
the use of the header variable now. My answer to the question "is the present form safe until
#116": the default (the variable stays unset) is safe, and each other configuration is not.

The fix, one of the two:

1. Read the last element of the comma list. This equals the D20 default of one trusted proxy.
   ```kotlin
   val headerValue = headerName?.let { call.request.header(it) }
       ?.substringAfterLast(',')?.trim()?.takeIf { it.isNotEmpty() }
   ```
2. Or ignore the variable until #116 closes, and write one warning line at the start of the app.

In each case, add a section to `kit/jvm-ktor/README.md`. State that the app sets the variable only
when a proxy that the app trusts writes that header. State the safe default: the variable stays
unset, and the key is then the remote address of the connection.

## MAJOR 2 — a full map blocks every real user

`kit/jvm-core/src/main/java/octometer/kit/core/ingest/IngestRateLimiter.java:116`

One map holds a user key and a client-address key together. Above 10 000 keys the limiter answers
`LIMITED` for each key that the map does not hold yet. I filled the map with 10 000 anonymous keys.
I then called `check("real-user-1", "198.51.100.5")`. The result was `LIMITED`. A second real user
got `LIMITED` too.

An attacker reaches this state with 10 000 small requests in one window. That is about 170 requests
each second. One host with a routed IPv6 /64 prefix is enough (see MINOR 1), and one host with the
header of MAJOR 1 is enough. The map clears at each window, thus the attacker repeats the attack in
each minute. Each real user of the app then gets 429 for the whole minute. Without the limiter the
app still answers each real user. The limiter itself thus makes the outage. The threat model of a
public ingest route names exactly this case.

I know that the acceptance criterion of #33 states this behaviour, and that the Related section
moves the two LRU maps to #118. I still report it at this level, because #33 carries the label
`production-blocker`, and the present shape is a denial of service of the whole route.

The fix, one of the two:

1. Pull the two-map shape of D20 into this pull request: one map of 5 000 user keys, one map of
   20 000 anonymous keys. An anonymous flood then leaves each signed-in user free.
2. Or make the full map fail open: return `ALLOWED` for a new key, and keep the one warning line.
   The LRU eviction of #118 has the same effect, because it removes an old key and never blocks a
   new one.

Please choose the issue that holds the fix before this pull request merges.

## MAJOR 3 — the key text has no length limit

`kit/jvm-core/src/main/java/octometer/kit/core/ingest/IngestRateLimiter.java:92` and
`kit/jvm-ktor/src/main/kotlin/octometer/kit/ktor/IngestRoute.kt:250`

Nothing limits the length of a client address. With `OCTOMETER_CLIENT_IP_HEADER` set, the attacker
chooses the key text and the length of that text. Only the header limit of the engine bounds it. I
sent a header value of 7 000 characters to a real Netty server with the default settings, and the
server accepted it. I sent a header value of 6 000 characters through the route, and the answer was
204. That key entered the map. The limiter also accepts a client address of 10 000 characters.

I measured the heap of one full map, with four `System.gc()` calls before and after:

- 10 000 keys of 45 characters: **1 MB**. The class comment at `IngestRateLimiter.java:29-31` says
  "near 2 MB". That number is correct for the default.
- 10 000 keys of 6 000 characters: **58 MB**.

The map clears at each window, thus the app also makes 58 MB of garbage in each minute. That cost
is high for a small machine.

The fix: cap the length in `clientAddress`. The text of an IPv6 address needs 45 characters at the
most.

```kotlin
val headerValue = headerName?.let { call.request.header(it) }?.takeIf { it.length <= 64 }
```

A longer value then falls back to the remote address.

## MINOR 1 — an IPv6 key holds all 128 bits

Design decision D43 states that a key is one IPv4 address, or the first 64 bits of an IPv6 address.
`clientAddress` returns the full text of the address. One host with a routed /64 prefix thus gets a
new key in each request, and the limit of 120 requests never applies to it. Issue #116 owns D43.
Please keep that link, and record the gap in `kit/jvm-ktor/README.md` until #116 closes.

## MINOR 2 — no test holds the order of the limiter and the body read

The KDoc at `IngestRoute.kt:98` states that the route answers 429 before it reads the request body.
I moved the limiter call after `receiveLimitedText` (`IngestRoute.kt:138`) and I ran
`./gradlew :kit:jvm-ktor:test`. The result was `BUILD SUCCESSFUL`, 35 tests, 0 failures. No test
holds that order.

The order in the code is correct. I sent 120 anonymous requests, then one request with a body of
20 000 bytes. The answer was 429, and not 400. A request above the body limit thus counts one time
only.

The fix: add that check as a test. A body above the 16 KB limit of rule C18, sent after the window
limit, must give 429.

## MINOR 3 — the window start can move backwards

`kit/jvm-core/src/main/java/octometer/kit/core/ingest/IngestRateLimiter.java:136-147`

The method `counters` compares the window start with `!=`. A thread that reads the clock a moment
before a boundary, and then runs after a thread of the next window, moves the window back. It
clears the map of the current window. I drove one key to its limit in window 2. I then made one
call with a clock value of window 1. The next call of the key of window 2 changed from `LIMITED` to
`ALLOWED`. The effect fails open and the exposure is a few milliseconds at each boundary, but the
fix is small.

The fix: `if (windowStart > currentWindowStart)`.

## MINOR 4 — a request with a wrong content type never counts

`kit/jvm-ktor/src/main/kotlin/octometer/kit/ktor/IngestRoute.kt:126`

The 415 guard runs before the limiter, thus such a request never reaches `check`. An attacker sends
those requests with no limit. The cost of one such request is small, because the route reads no
body. Please state this order in the KDoc, or move the limiter before the content type guard.

## MINOR 5 — the language of the description

Decision 3 of the pull request description holds "Failing open to the remote address". This is an
`-ing` form as a noun. Please write "The route falls back to the remote address". The code
comments and the commit body follow the standard.

## Other checks

- **The user id key.** The user id comes only from `resolveUserId(call)` (`IngestRoute.kt:137`).
  The route reads the body after the limit check. A `userId` field in the body can never select the
  key. Rule C16 and rule C32 stay true. PASS.
- **A key from a `Host` value or a `Forwarded` header.** No such key exists (see criterion 3).
- **The growth of the memory over 24 hours.** The limiter replaces the whole map at each window and
  it holds one `long` for the warning. No state grows across windows. PASS.
- **The diff.** It holds five source files. It holds no secret, no database file, and no build
  output.
- **My test counts (JDK 21).** `:kit:jvm-core:test`: 148 tests, 0 failures, 0 errors, 0 skipped.
  `:kit:jvm-ktor:test`: 35 tests; 4 runs of 5 gave 0 failures, and 1 run gave 1 failure (BLOCKER 1).

## The mutation checks

I broke the code six times. I reverted each change.

| # | The change | The result |
|---|---|---|
| 1 | `count > limit` becomes `count > limit + 1` | 4 tests fail. Good. |
| 2 | The map cap becomes 1 000 000 keys | `whenTheMapHolds10000KeysANewKeyIsLimitedUntilTheWindowEnds` fails. Good. |
| 3 | The warning always writes a line | 2 log tests fail. Good. |
| 4 | Both key prefixes are removed | **No test fails.** BLOCKER 2. |
| 5 | The limiter call moves after the body read | **No test fails.** MINOR 2. |
| 6 | `clientAddress` always reads `X-Forwarded-For` | The spoofed-header test fails. Good. |

xamcross:
**Review (Java concurrency engineer): CHANGES REQUESTED**

I made my own clone, I ran `gh pr checkout 158`, and I ran each command with JDK 21
(Microsoft 21.0.9). Each number below comes from a run that I made myself.

## The acceptance criteria of issue #33

|---|---|---|---|
| 1 | Fixed clock: request 31 of one user id is limited, a request after the window passes | PASS | `./gradlew :kit:jvm-core:test` = 148 tests, 0 failures. Mutation M1 (`USER_LIMIT_PER_WINDOW` 30 to 31) fails `request31OfOneUserIdInTheSameWindowIsLimited`. Mutation M2 (no new map at a window change) fails `aRequestOfTheSameUserIdAfterTheWindowEndsPasses`. |
| 2 | Without a user id, request 121 of one client address is limited | PASS | Mutation M3 (`count > limit + 1`) fails `withoutAUserIdRequest121OfOneClientAddressIsLimited`. |
| 3 | Without `OCTOMETER_CLIENT_IP_HEADER` the key is the remote address | PASS | The route test at `IngestRouteTest.kt:918` sends 120 different `X-Forwarded-For` values and still gets 429. Mutation M9 (`clientAddress` ignores the header) fails the test at line 884. |
| 4 | With 10 000 keys a new key is limited until the window ends | PASS | Mutation M4 (no cap) and mutation M5 (`>=` to `>`) each fail `whenTheMapHolds10000KeysANewKeyIsLimitedUntilTheWindowEnds`. See finding MAJOR 4 for my judgement of this rule. |
| 5 | No stored event and no log line holds the IP address | PASS | The warning text is a constant (`IngestRateLimiter.java:160`). Mutation M7 (the address in the warning) fails `aRejectionWritesAMaximumOfOneLogLineForTheWindowWithNoKeyInIt`. `IngestPipeline.ingest` never gets the address (`IngestRoute.kt:154`). |
| 6 | The Ktor route returns 429 | PASS | Mutation M8 (the 429 branch never fires) fails 5 route tests. |
| 7 | `:kit:jvm-core:test` and `:kit:jvm-ktor:test` pass | PASS | 148 tests / 0 failures, and 35 tests / 0 failures. `./gradlew build` = BUILD SUCCESSFUL in 2 m 47 s. |
| 8 | A comment on the issue records the test output | PASS | The comment is on issue #33. |

Each criterion passes. The findings below are defects that the criteria do not cover.

## The checks of my role

- **The exact count under contention.** 20 runs, 1 000 requests, 8 threads, one user key,
  a fixed clock: each run gives exactly 30 allowed and 970 limited. No lost update inside
  one window. `compute` has no side effect; `computeIfAbsent` only creates one
  `AtomicInteger` (`IngestRateLimiter.java:120`). Good.
- **The window change under contention.** It has a race. See BLOCKER 1.
- **The memory after one day.** 1 440 windows with 10 000 new keys each, a 256 MB heap:
  739 KB stay at the end. The keys of an old window go. Good.
- **The cost.** 42 ns for one call on the hot path (1 M calls, one thread, an existing key
  over its limit). 60 ns for one call on the allowed path (1 M calls, a new window each 30
  calls). 69 ns for one call as wall time with 8 threads against one key (1 M calls).
  114 ns for one call when the map holds 10 000 keys and each key is new (1 M calls).
  A rejected request allocates 56 bytes: one key text, measured with
  `getCurrentThreadAllocatedBytes()` over 1 M rejected calls. The cost is small. Good.
- **The IPv6 rule.** The code uses the full address text as the key. D20 states no IPv6
  rule, thus the code breaks no rule. One host with a `/64` prefix still has 2^64 keys.
  See MINOR 11.
- **The header forms**, through a real Netty server and a raw socket:
  `X-Client-Ip: 203.0.113.9, 198.51.100.2` gives the key `ip:203.0.113.9, 198.51.100.2`;
  a leading space and a trailing space give `ip:203.0.113.9` (Netty trims them, thus a
  space makes no second key); an empty value gives the key `ip:` (MINOR 6); a folded
  value gives `ip:203.0.113.9 198.51.100.2` (Netty replaces the line break with a space,
  thus no line break reaches the key or the log); two header lines give the **first**
  value (MAJOR 3); a value of 6 000 characters gives a key of 6 000 characters
  (BLOCKER 2).
- **The changed old test** (`IngestRouteTest.kt:406`). It still proves what it proved
  before. The assertion `totalMillis >= 900` still proves that the 8-wide dispatcher
  serializes the 64 slow calls, and the latch still proves that other `Dispatchers.IO`
  work stays free. With one user id the 64 calls get a fast 429, the store never runs,
  and the 900 ms assertion fails, thus the change is necessary. See MINOR 13 for a style
  point.
- **The mutation check.** I broke the code 10 times. 9 mutations fail the right test.
  One mutation passes: see MINOR 7.
- **The test merge with pull request #156.** `git fetch origin pull/156/head:pr156`, then
  `git merge --no-commit --no-ff pr156`: "Automatic merge went well". No conflict.
  I aborted the merge.
- **The scope and the content.** The diff holds 5 source files. No secret, no database
  file, no build output. The language of the new comments and of the commit body follows
  ASD-STE100.

## Findings

### BLOCKER 1 — a thread at the window boundary resets the live window

`kit/jvm-core/src/main/java/octometer/kit/core/ingest/IngestRateLimiter.java:136-147`

`check` reads the clock at line 105 and calls `counters(windowStart)` at line 107. The
two steps are not one atomic step. A thread that reads the clock just before a boundary,
and then loses its processor (a GC pause, a full run queue), reaches line 141 after other
threads already moved the map to the next window. Line 141 compares with `!=`, thus that
late thread builds a **third** map and moves `currentWindowStart` **backwards**. Every
count of the live window goes away. Each key then gets a full new quota inside the same
minute.

My evidence. I gave the limiter a clock that returns the old window for one thread named
`straggler`, and the new window for each other thread. The straggler waits inside
`millis()` until the other thread finishes, which models exactly this interleave. 5 runs
of 5:

```
run 1 allowedInW2=30 request31=LIMITED afterTheStraggler=ALLOWED extraAllowed=29 *** THE WINDOW RESET ***
```

The key used its full quota of 30 and got `LIMITED`. After the one straggler it got 30
more allowed requests in the same window. The straggler's own request is counted in no
live window, because its increment went into a map that nothing reads again (that is why
30 follow, and not 29).

A start of a window is public: it is the wall clock, aligned to a multiple of 60 seconds.
An attacker who sends a burst at each boundary raises the chance of a straggler. The
system clock is also not monotonic: one backward NTP step across a boundary gives the
same reset with no concurrency at all.

The same non-atomic pair breaks the fast path at lines 137-138: a thread can read
`currentWindowStart` of the old window and then `counters` of the new one.

The fix. Hold the pair in one immutable object, read it one time, and never move the
window backwards:

```java
private record Window(long start, ConcurrentHashMap<String, AtomicInteger> counters) { }

private volatile Window window = new Window(Long.MIN_VALUE, new ConcurrentHashMap<>());

private ConcurrentHashMap<String, AtomicInteger> counters(long windowStart) {
    Window current = window;                        // one read of the pair
    if (current.start() == windowStart) {
        return current.counters();
    }
    synchronized (windowLock) {
        current = window;
        if (windowStart > current.start()) {        // the window never moves back
            current = new Window(windowStart, new ConcurrentHashMap<>());
            window = current;
        }
        return current.counters();
    }
}
```

A late thread then counts in the new window. That is one extra count for each straggler,
and it is safe. Please add the test with the latch clock above: it fails today and it
passes after the fix.

### BLOCKER 2 — the key text has no length limit, thus the map holds up to 80 MB

`IngestRateLimiter.java:29-32` and `:98`, with
`kit/jvm-ktor/src/main/kotlin/octometer/kit/ktor/IngestRoute.kt:250-253`

`MAX_TRACKED_KEYS` limits the count of the keys. It does not limit the size of one key.
With `OCTOMETER_CLIENT_IP_HEADER` set, the client writes the key text itself, and the
default `maxHeaderSize` of the Netty engine is 8 192 bytes. My raw-socket probe shows
that a header value of 6 000 characters reaches the route and becomes the key.

My measurement of the heap that 10 000 keys hold, after two full collections:

| Key length | Retained heap |
|---|---|
| 45 characters (an IPv6 text) | 1 494 240 bytes (1.4 MB) |
| 8 000 characters (one header value) | 81 726 808 bytes (77 MB) |

The class comment at lines 29-32 promises "near 2 MB". 10 000 cheap requests give 77 MB
instead, and they hold it until the window ends. Issue #33 blocks production because one
client can fill a 512 MB cluster. This class must not give that same client a second way
to do it. A user id of 10 000 characters gives the same result through `resolveUserId`
(I ran it: the key works, the limit works, the memory grows).

The fix, in `IngestRoute.kt`:

```kotlin
private const val MAX_CLIENT_ADDRESS_LENGTH = 64

private fun clientAddress(call: ApplicationCall, headerName: String?): String {
    val headerValue = headerName
        ?.let { call.request.header(it) }
        ?.takeIf { it.isNotBlank() && it.length <= MAX_CLIENT_ADDRESS_LENGTH }
    return headerValue ?: call.request.local.remoteAddress
}
```

45 characters hold each IPv6 text, thus 64 is generous. Please also cut a key above a
fixed length inside `check`, because `resolveUserId` of the app is the second source.
Then correct the "near 2 MB" sentence.

### MAJOR 3 — with the header set, a client bypasses the limit in full

`IngestRoute.kt:251`

The route takes the raw header value as the key. A proxy that **appends** (the normal
rule of `X-Forwarded-For`) writes `<the value of the client>, <the real address>`, thus
the client controls a part of the key. A proxy that appends a **second header line**
leaves the line of the client first, and `call.request.header(name)` returns the first
line: my probe sends `X-Client-Ip: 1.1.1.1` and `X-Client-Ip: 2.2.2.2`, and the route
reads `1.1.1.1`.

My evidence for the bypass: with `clientIpHeaderName = "X-Forwarded-For"`, 300 anonymous
requests with a different header value each give 300 answers 204 and 300 stored events.
The limit of 120 never fires. The same 10 000 varied values also fill the map, which
gives finding MAJOR 4 to every other visitor.

Issue #33 assigns the trusted proxy rule to #116, thus the deferral of
`OCTOMETER_TRUSTED_PROXY_COUNT` is correct. The deferral of the **last element** rule is
not, because it leaves a `production-blocker` limit with no effect in the one deployment
that the variable exists for. D20 sets the default count to 1, counted from the right,
which is the last element. That default needs no new variable:

```kotlin
val headerValue = headerName
    ?.let { call.request.headers.getAll(it)?.lastOrNull() }   // the last header line
    ?.substringAfterLast(',')                                  // the last element
    ?.trim()
    ?.takeIf { it.isNotEmpty() && it.length <= MAX_CLIENT_ADDRESS_LENGTH }
return headerValue ?: call.request.local.remoteAddress
```

Issue #116 then only adds the count. If the maintainer keeps the raw value, the KDoc at
lines 100-106 must state plainly that the variable is safe only with a proxy that
**replaces** the header.

### MAJOR 4 — the full map blocks each new visitor, and one map mixes the two limits

`IngestRateLimiter.java:116-119`

This is my answer to the question of the maintainer about the rule "REJECT a new key when
the map is full". The rule is correct against the acceptance criterion. It is not
acceptable for a `production-blocker` limit, for two reasons.

1. It turns a limit for one key into a switch for every key. After 10 000 distinct keys
   in one window, each new visitor gets 429 until the window ends. 10 000 cheap requests
   buy that. One host with an IPv6 `/64` prefix is enough; a botnet is not necessary.
   With the header of MAJOR 3 one connection is enough.
2. One map holds the user keys and the anonymous keys together. An anonymous flood takes
   the 10 000 slots, thus a signed-in user with no counter yet also gets 429. D20 splits
   the map in two (5 000 user keys, 20 000 anonymous keys) for exactly this reason.

My recommendation: take the two maps with the LRU eviction of D20 into this pull request,
and give issue #118 only the tuning that stays. The reason: the eviction changes the
failure from "each new visitor is blocked" to "the oldest key loses its count", which
costs one attacker a little and costs a legitimate visitor nothing. The alternative that
the design also allows, "a new key passes with no count", is weaker: it lets the same
10 000 keys pass with no limit at all.

The smallest change that still helps, if the maintainer keeps one map for now: hold two
maps with a fixed cap each (5 000 user keys and 20 000 anonymous keys, the numbers of
D20), so that an anonymous flood never blocks a signed-in user. The maintainer decides
the level of this finding.

### MAJOR 5 — four new route tests read the system clock and flake at a window boundary

`IngestRouteTest.kt:829`, `:857`, `:918`, `:947`

Only the test at line 793 injects a clock. The four tests above install the route with no
`clock` argument, thus the limiter reads `Clock.systemUTC()`. Each one sends 31 or 121
requests in a loop and then asserts 429. A 60-second boundary inside that loop clears the
map, and the last request gives 204. The test then fails for no defect of the code.

The measured durations of these four tests on my machine are 0.122 s, 0.05 s, 0.158 s and
0.139 s, thus about 0.8 % of the runs fail on this machine. A slow or busy CI machine
raises that number by the same factor as it raises the durations.

The fix: give each of the four tests its own `MutableClock`, exactly as the test at line
793 does.

### MINOR findings

6. `IngestRoute.kt:251`. An empty header value becomes the key `ip:`, because `?:` only
   catches `null`. Each client with an empty header then shares one counter of 120
   requests. The KDoc at lines 103-105 promises a fallback to the remote address. The
   `takeIf { it.isNotBlank() }` of BLOCKER 2 fixes it.
7. A test that cannot fail. I moved the limiter check **after** the body read
   (`IngestRoute.kt:143`) and ran `:kit:jvm-ktor:test`: 0 failures. The test at line 857
   only proves that the check runs before the **parser**, not before the **body read**.
   The fix: send a body above 16 KB from a limited client and assert 429. I ran that
   test against this branch: it gives 429 today, and it gives 400 with the check after
   the body read.
8. `IngestRouteTest.kt:976`. The test asserts only that no message holds "The Octometer
   ingest route failed". The criterion is "no log line holds the IP address". Please
   assert that no captured message holds the client address and no message holds the user
   id.
9. `IngestRoute.kt:138`. `clientAddress(call, clientIpHeaderName)` runs for each request,
   also for a signed-in user, where `check` ignores it. Read the header only when the
   user id is `null`.
10. `IngestRoute.kt:137`. `resolveUserId` runs before the limit check, thus a rejected
    request still pays the session lookup of the app. The KDoc already asks for a short
    function. One sentence about the rate limit helps the reader of the app.
11. D20 states no IPv6 rule, and the code uses the full address text. One home prefix
    (`/64`) gives one client 2^64 keys, thus the limit of 120 has no effect against it,
    and MAJOR 4 follows. Please record a decision: the first 64 bits of an IPv6 address
    make the key. Issue #116 or issue #118 can own it.
12. `IngestRateLimiter.java:154-164`. `lastWarnedWindowStart` can move backwards together
    with the window of BLOCKER 1, thus one window can write two warning lines. The fix of
    BLOCKER 1 removes this case too.
13. `IngestRouteTest.kt:417`. `java.util.concurrent.atomic.AtomicInteger` is written in
    full, although the file already imports `AtomicBoolean` and `AtomicReference` from
    that package. Add the import.

## Summary

The limiter is small, it is fast (42 ns for one call), it allocates little (56 bytes for
one rejection), and it is exact inside one window (20 runs of 8 threads give exactly 30).
The window change and the key text are the two defects that must go before a merge.

xamcross:
## Correction round 1

Commits `9d88c66` (the code and test correction) and `4c2b033` (the
README note). This comment answers the security review
(https://github.com/xamcross/octometer/pull/158#issuecomment-5773081361)
and the concurrency review
(https://github.com/xamcross/octometer/pull/158#issuecomment-5773088140),
finding by finding.

### BLOCKER and MAJOR findings

| Review | Finding | File and line | Result |
|---|---|---|---|
| Concurrency BLOCKER 1 | A stalled thread at a window boundary could move the window backwards and reset every count. | `IngestRateLimiter.java:157` (`windowFor`), `IngestRateLimiter.java:230` (`Window` record) | Fixed. One `Window` record in one `volatile` field; the window changes only when `requestedWindowStart > current.start()`. New test `IngestRateLimiterTest.aStragglerThreadAtAWindowBoundaryNeverResetsTheLiveWindow` (the latch-clock test), 5 runs. |
| Security MINOR 3 (same defect) | The window start comparison used `!=`, so a stale clock read could move it backwards. | `IngestRateLimiter.java` (old line 141) | Fixed by the same change as Concurrency BLOCKER 1. |
| Concurrency BLOCKER 2 | A key text had no length limit; a 6 000-character header value cost 58-77 MB per window. | `IngestRateLimiter.java:196` (`shortenKey`), `IngestRoute.kt:45` | Fixed. A key above 64 characters becomes its SHA-256 hex text (`MessageDigest`, no dependency). New test `IngestRateLimiterTest.aKeyTextAbove64CharactersIsShortenedAndTwoLongKeysNeverJoin` (two 6 000-character values sharing a 64-character prefix, so a truncating shortener would join them; the hash does not). |
| Security MAJOR 3 (same defect) | Same finding, at MAJOR severity. | `IngestRateLimiter.java:92` (old), `IngestRoute.kt:250` (old) | Fixed by the same change. |
| Security MAJOR 1 | `OCTOMETER_CLIENT_IP_HEADER` read the raw header value; a proxy that appends its peer address let an attacker choose the key, or exhaust a named victim's budget. | `IngestRoute.kt:321` (`clientAddress`) | Fixed. The route reads the last element of the last header line (the default trusted proxy count of one, design decision D20). New tests: "a forged left element never changes the key", "a header value above 64 characters gives the remote address", "a value with no IP address form gives the remote address". |
| Concurrency MAJOR 3 (same defect) | Same finding: two header lines gave the first line; a proxy that appends gave the client full control of the key. | `IngestRoute.kt:250` (old) | Fixed by the same change. |
| Security MAJOR 2 | A full map (10 000 keys, one cap for both kinds of key) gave `LIMITED` to every new key until the window ended; an anonymous flood locked out every real user. | `IngestRateLimiter.java:116` (old) | Fixed. Two maps, 5 000 user keys and 20 000 client address keys, each a `LinkedHashMap` in access order with its own LRU eviction. New tests `tenThousandAnonymousKeysBlockNoUserAndNoNewVisitor`, `aFullClientAddressMapEvictsItsOldestKeyAndGivesTheNewKeyAFreshCounter`. |
| Concurrency MAJOR 4 (same defect) | Same finding, plus: one shared map let an anonymous flood evict a signed-in user's own counter. | `IngestRateLimiter.java:116` (old) | Fixed by the same change; `tenThousandAnonymousKeysBlockNoUserAndNoNewVisitor` drives a real user to its limit first, then floods, then confirms the user's own `LIMITED` result survives. |
| Security BLOCKER 1 | Six route tests read `Clock.systemUTC()`; a run that crossed a minute boundary made the last assertion fail. | `IngestRouteTest.kt:829, 857, 884, 918, 947, 976` (old) | Fixed. Each of the six tests now passes `clock = fixedClock`. `./gradlew :kit:jvm-ktor:test` ran 5 times in sequence: 0 failures each run. |
| Concurrency MAJOR 5 (same defect) | Same finding, with a measured flake rate of about 0.8% per run on the reviewer's machine. | Same lines | Fixed by the same change. |
| Security BLOCKER 2 | `aUserIdKeyAndAClientAddressKeyOfTheSameTextStayIndependent` drove the user key to its limit first, so a shared counter would still pass; the reviewer removed both key prefixes and no test failed. | `IngestRateLimiterTest.java:83` (old) | Fixed. The test now drives the client address key first (30 of its 120), then makes one user-id check of the same text. The prefix mutation no longer applies, because the corrected code uses two separate map objects, not one map with two text prefixes; I instead mutated `windowFor` to share one `KeyMap` between the two kinds of key (an equivalent regression) and confirmed the renamed test `aClientAddressKeyAndAUserIdKeyOfTheSameTextStayIndependent` fails, then reverted. |

### MINOR findings

11 of 13 distinct MINOR findings are corrected (2 MINOR findings, one from each review, are the same IPv6 gap and stay open for issue #116/#118).

| Review | Finding | File and line | Result |
|---|---|---|---|
| Security MINOR 1 | An IPv6 key uses the full 128 bits; a `/64` prefix gets unlimited keys. | `IngestRoute.kt:321` | Not fixed. Design decision D20 states no IPv6 rule; the `/64` rule is D43, owned by issue #116/#118. Recorded in `kit/jvm-ktor/README.md`, "The ingest rate limit", "A known gap". |
| Security MINOR 2 | No test held the order of the rate limit check and the body read. | `IngestRoute.kt:203` (`receiveLimitedText` call) | Fixed. New test "a client already limited gets 429 for a body above the maximum size, never 400". Mutated the route to read the body before the limit check: the new test failed with 400; reverted. |
| Security MINOR 3 | The window start could move backwards. | (see Concurrency BLOCKER 1 above) | Fixed by the BLOCKER 1 change. |
| Security MINOR 4 | The 415 content-type guard runs before the limit check, so a wrong-content-type request is never rate-limited. | `IngestRoute.kt:175` | Fixed (the documentation option the review offered). A KDoc comment now states the order and the reason; the check order itself is unchanged, because reordering was not one of the maintainer's numbered corrections. |
| Security MINOR 5 | "Failing open" is an `-ing` form as a noun. | Pull request description | Fixed. The pull request description now reads "falls back to the remote address" throughout. |
| Concurrency MINOR 6 | An empty header value gave the key `ip:`, shared by every client with an empty header. | `IngestRoute.kt:250` (old) | Fixed by the header rewrite: `takeIf { it.isNotEmpty() ... }`. |
| Concurrency MINOR 7 | A test that cannot fail: the limiter check moved after the body read, and no test failed. | `IngestRouteTest.kt:857` (old) | Fixed by the same new test as Security MINOR 2. |
| Concurrency MINOR 8 | A log-line test asserted only a fixed string, not the absence of a key. | `IngestRouteTest.kt:976` (old) | Fixed. The test now asserts that no captured message holds a known client address value, using `clientIpHeaderName` and a fixed value. |
| Concurrency MINOR 9 | `clientAddress(call, ...)` ran for every request, also for a signed-in user, whose key `check` never reads. | `IngestRoute.kt:188` (old) | Fixed. The route now calls `clientAddress` only when `userId == null`. |
| Concurrency MINOR 10 | `resolveUserId` runs before the limit check, so a rejected request still pays its cost; ask for one KDoc sentence. | `IngestRoute.kt` KDoc | Fixed. One sentence added to the `resolveUserId` parameter doc. |
| Concurrency MINOR 11 | Same IPv6 `/64` gap as Security MINOR 1. | `IngestRateLimiter.java` | Not fixed, for the same reason; same README note. |
| Concurrency MINOR 12 | `lastWarnedWindowStart` could move backwards together with the window. | `IngestRateLimiter.java` (old) | Fixed by the BLOCKER 1 change: `warnOncePerWindow` now always receives `current.start()`, the window that a request actually counted in, never a stale computed value. |
| Concurrency MINOR 13 | `AtomicInteger` was written with its full package name, although the file already imports two sibling classes from the same package. | `IngestRouteTest.kt:420` (old) | Fixed. Added the import; the call site now reads `AtomicInteger()`. |

### Test results (JDK 21, `C:\Users\xamcr\.jdks\ms-21.0.9`)

- `./gradlew :kit:jvm-core:test`: BUILD SUCCESSFUL, 152 tests, 0 failures, 0 errors, 0 skipped (includes the 13 tests of `IngestRateLimiterTest`, up from 10).
- `./gradlew :kit:jvm-ktor:test`, run 5 times in sequence: BUILD SUCCESSFUL every time, 39 tests, 0 failures, 0 errors, 0 skipped each run (up from 35; this round adds 4 tests).
- `./gradlew build`: BUILD SUCCESSFUL. `kit/jvm-mongo`: 32 tests, 0 failures (container tests ran against Docker), plus 32 tests of `testMongoSyncNewest`, 0 failures. `monitor/backend`: 316 tests, 0 failures.
- CI checks of commit `4c2b033`: Frontend, JVM modules, Monitor backend on Windows, Secret scan, Tracker, and the dependabot config check all pass (`gh pr checks 158`).

### The map form (design decision D20, acceptance criterion 4)

Two `LinkedHashMap` instances in access order, one lock for each map, not
`ConcurrentHashMap`. The class comment of `IngestRateLimiter` states the
reason: an LRU eviction needs a write on each read (the access-order
reorder), and a lock-free map cannot give an exact eviction order under
contention. The acceptance criterion of an exact count under contention
needs that exact order more than it needs a lock-free path.

### Mutation checks of this round

Each new test was checked this way: break the code, run the test, see
the failure, restore the code.

1. `windowFor`'s monotonic guard (`>` to `!=`): `aStragglerThreadAtAWindowBoundaryNeverResetsTheLiveWindow` failed. Reverted.
2. `shortenKey` (hash to truncate): `aKeyTextAbove64CharactersIsShortenedAndTwoLongKeysNeverJoin` failed. Reverted.
3. `KeyMap.removeEldestEntry` (always `false`): `aFullClientAddressMapEvictsItsOldestKeyAndGivesTheNewKeyAFreshCounter` failed. Reverted.
4. `windowFor`'s two maps shared as one `KeyMap` with a cap of 10 000: both `aClientAddressKeyAndAUserIdKeyOfTheSameTextStayIndependent` and `tenThousandAnonymousKeysBlockNoUserAndNoNewVisitor` failed. Reverted.
5. `KeyMap.incrementAndGet`'s lock removed: `oneThousandRequestsFrom8ThreadsAgainstOneKeyGiveExactlyTheLimitOfAcceptedRequests` (the 20-run test) failed. Reverted.
6. `clientAddress`'s length and address-form check removed: both new "above 64 characters" and "no IP address form" tests failed. Reverted.
7. `clientAddress`'s comma split reversed (`substringBeforeLast`): "a forged left element never changes the key" failed. Reverted.
8. The rate limit check moved after the body read: "a client already limited gets 429 for a body above the maximum size, never 400" failed with 400. Reverted.

Each mutation above was made, run, observed as a failure, and reverted
by me in this round.

xamcross:

I made a second clone, I ran `gh pr checkout 158` in it, and I ran each check myself with JDK 21
against `4c2b033`. I deleted the clone at the end. Each result below is my own observation, not a
copy of the answer of the author.

## My BLOCKER and MAJOR findings

| Finding | My check | Result |
|---|---|---|
| BLOCKER 1, the wall clock | Each of the 13 rate-limit route tests passes `clock = fixedClock` or a `MutableClock`. I ran `./gradlew :kit:jvm-ktor:test` three times: `BUILD SUCCESSFUL`, 39 tests, 0 failures each run. | **Fixed** |
| BLOCKER 2, a test that cannot fail | I made `check` read one map for both kinds of key (`IngestRateLimiter.java:130`). `aClientAddressKeyAndAUserIdKeyOfTheSameTextStayIndependent` then failed (13 tests, 1 failed). I reverted the change. | **Fixed** |
| MAJOR 1, the forged header | Attack A: 130 requests, each with a new left element `203.0.113.<n>, 127.0.0.1`. 120 passed, then each one gave 429 (130 of 130 passed before). Attack B: 130 requests with the forged victim value `198.51.100.5, 127.0.0.1`. 120 passed, then 429; one later request with `10.0.0.1, 198.51.100.5` gave 204, thus the budget of the named visitor stays free. Two header lines: the last line wins, and a forged first line changes no key. A value of 70 characters and a value with no address form each fall back to the remote address: 120 passed, then 429. `kit/jvm-ktor/README.md:40` holds the rule "Set this option only behind a proxy". | **Fixed** |
| MAJOR 2, a full map | After 10 000 anonymous keys a signed-in user got `ALLOWED`. After 35 000 anonymous keys a signed-in user and a new visitor each got `ALLOWED`. | **Fixed** |
| MAJOR 3, the key length | 20 000 keys of 6 000 characters now cost 2 MB, the same as 20 000 keys of 45 characters (58 MB for 10 000 long keys before). Two 6 000-character values that share a 64-character prefix stay apart: the first reached its limit, the second still got `ALLOWED`. | **Fixed** |

## My MINOR findings

| # | Finding | My check | Result |
|---|---|---|---|
| 1 | An IPv6 key holds all 128 bits | `kit/jvm-ktor/README.md:52-56` states the gap and names #116 and #118. | Open, recorded |
| 2 | No test held the order of the limiter and the body read | I moved the limiter check after `receiveLimitedText` (`IngestRoute.kt:198`). The test "a client already limited gets 429 for a body above the maximum size, never 400" failed (39 tests, 1 failed). I reverted the change. | Fixed |
| 3 | The window start could move backwards | I drove one key to its limit in window 2, made one call with a clock value of window 1, then checked the key again. The result stayed `LIMITED` (it changed to `ALLOWED` before). | Fixed |
| 4 | A request with a wrong content type never counts | `IngestRoute.kt:170-174` states the order and the reason. | Fixed |
| 5 | "Failing open" in the description | The description holds no `-ing` form as a noun now. | Fixed |

## Two notes for the maintainer, at no severity

1. Acceptance criterion 4 of issue #33 reads "a new key gets the rate-limit result until the window
   ends". The corrected code evicts the oldest key instead, as design decision D20 states. The code
   is right, and the criterion text is now old. Please correct the issue body, and decide whether
   #118 still has work.
2. An LRU eviction gives one property that the maintainer must accept: a key that reached its limit
   loses its counter when the map evicts it. I drove one key to its limit, then I added 25 000 new
   keys, and the first key got `ALLOWED` again. An attacker needs 20 000 addresses for this, and an
   attacker with 20 000 addresses already has 20 000 fresh budgets. The IPv6 gap of MINOR 1 is the
   cheap form of the same attack, and #116 owns it.

## Test counts of my own runs (JDK 21)

- `./gradlew :kit:jvm-core:test`: 152 tests, 0 failures, 0 errors, 0 skipped.
- `./gradlew :kit:jvm-ktor:test`, three runs: 39 tests, 0 failures, 0 errors, 0 skipped each run.

No BLOCKER and no MAJOR finding stays. I approve this pull request.

xamcross:
**Review (second review, Java concurrency engineer): APPROVE**

I made a second clone, I ran `gh pr checkout 158` on `4c2b033`, and I ran each command
with JDK 21 (Microsoft 21.0.9). Each number below comes from a run that I made myself. I
confirm each BLOCKER finding and each MAJOR finding of my first review as corrected.

## BLOCKER 1 — the window never moves backwards. Fixed.

`IngestRateLimiter.java:157-170`, `:230`

I ran my own latch-clock probe again: one thread named `straggler` reads the clock of the
old window, waits inside `millis()` until another thread uses the whole quota of the new
window, and then reaches `windowFor`. 5 runs of 5:

```
run 1 allowedInW2=30 request31=LIMITED straggler=LIMITED afterTheStraggler=LIMITED extraAllowed=0 ok, no reset
```

The straggler now counts in the live window and gets `LIMITED`. No count goes away, and
no key gets a second quota. Before the correction the same probe gave 30 extra allowed
requests in each run.

My mutation N1 (`requestedWindowStart > current.start()` back to `!=`) fails
`aStragglerThreadAtAWindowBoundaryNeverResetsTheLiveWindow`, and nothing else. The new
test holds the fix.

## BLOCKER 2 — the key text is bounded. Fixed.

`IngestRateLimiter.java:196-220`, `IngestRoute.kt:321`

| My measurement | Round 1 | Round 2 |
|---|---|---|
| 10 000 keys of 6 000 characters | 77 MB | 935 KB |
| Both maps at their cap (20 000 + 5 000 keys) | not applicable | 3.5 MB |
| A header value of 6 000 characters | becomes the key | gives the remote address (the first 429 arrives at request 121, one shared key) |

My mutation N3 (a truncation of 64 characters instead of the hash) fails
`aKeyTextAbove64CharactersIsShortenedAndTwoLongKeysNeverJoin`. The two-map memory now
stays near 3.5 MB, which matches the class comment.

## MAJOR 3 — the client cannot choose the key. Fixed.

`IngestRoute.kt:321-333`

I sent 130 requests for each header form against the route, with
`clientIpHeaderName = "X-Client-Ip"`. The number is the request of the first 429:

| The header form | The first 429 |
|---|---|
| A forged left element, `9.9.9.<n>, 203.0.113.9` | request 121 |
| Two header lines, the client line first (a real Netty server, a raw socket) | request 121 |
| A value of 6 000 characters | request 121 |
| A value with no address form, `evil-<n>` | request 121 |
| An empty value | request 121 |

Each form gives one shared key, thus the limit of 120 holds. In round 1 the first form
gave 300 answers 204 out of 300 requests.

## MAJOR 4 — the two LRU maps. Fixed, and the lock is an acceptable price.

`IngestRateLimiter.java:239-271`

My probe, with my own numbers:

- After a flood of 10 000 anonymous keys: a new visitor gets `ALLOWED`, a new user gets
  `ALLOWED`, and a user that already reached its own limit of 30 stays `LIMITED`. The
  flood takes nothing from the user map.
- A full client address map evicts its oldest key: that key then gets a full new counter
  of 120 allowed requests, not 119. The eviction is exact.
- 50 000 distinct new keys, each its first request: 0 rejections. A full map never
  rejects a new key any more.
- The exact count under contention: 20 runs, 1 000 requests, 8 threads, one key. Each
  run gives exactly 30 allowed and 970 limited. 0 wrong runs.

My mutations: N2 (`removeEldestEntry` returns `false`) fails
`aFullClientAddressMapEvictsItsOldestKeyAndGivesTheNewKeyAFreshCounter`; N5 (one map for
both kinds of key) fails `aClientAddressKeyAndAUserIdKeyOfTheSameTextStayIndependent`;
N6 (the lock of `KeyMap.incrementAndGet` removed) fails the 20-run contention test. The
lock is necessary, and the test proves it.

**The price of the lock.** I compiled the first form again from commit `cfad080` and I
measured both forms in the same way:

| The measurement (1 M calls) | The first form (`ConcurrentHashMap`) | This form (`LinkedHashMap` under a lock) |
|---|---|---|
| One thread, one key | 42 ns for one call | **37 ns** for one call |
| 8 threads, one key | 69 ns | 117-146 ns |
| 8 threads, 8 keys of one map | 82 ns | 253 ns |
| The allocation of a rejected request | 56 bytes | **16 bytes** |

One thread is now faster, because the key needs no text prefix any more. Full contention
across 8 keys costs about 3 times more, thus about 3.9 M checks each second on my
machine. One ingest request parses JSON and writes to MongoDB, which costs some hundreds
of microseconds; 253 ns is under one part in a thousand of that. I accept the trade: the
exact eviction order of D20 needs the lock, and the class comment states the reason.

## MAJOR 5 — each route test injects a clock. Fixed.

`IngestRouteTest.kt` holds 11 `clock = fixedClock` arguments; no rate-limit test reads the
wall clock. I ran `./gradlew :kit:jvm-ktor:test --rerun` five times: 39 tests, 0 failures,
0 errors, 0 skipped in each of the five runs.

## The 8 MINOR findings of my first review

|---|---|---|---|
| 6 | An empty header value gave the key `ip:` | Fixed | `IngestRoute.kt:325` uses `takeIf { it.isNotEmpty() ... }`; my probe with an empty value gives the first 429 at request 121, thus the remote address is the key. |
| 7 | No test held the order of the check and the body read | Fixed | My mutation N4 moves the limit check after the body read: `a client already limited gets 429 for a body above the maximum size, never 400` fails. In round 1 the same mutation broke no test. |
| 8 | The log test asserted too little | Fixed | `IngestRouteTest.kt:1010` asserts that no message holds the known client address `203.0.113.55`. |
| 9 | `clientAddress` ran for a signed-in user too | Fixed | `IngestRoute.kt:190-195` calls `clientAddress` only when `userId == null`. |
| 10 | `resolveUserId` runs before the check, with no note | Fixed | The KDoc at `IngestRoute.kt:150-152` states it. |
| 11 | An IPv6 key uses all 128 bits | **Open, and the reason holds** | D20 states no IPv6 rule, and D43 (line 513 of the design document) owns "the first 64 bits of an IPv6 address". Issue #116 owns D43. `kit/jvm-ktor/README.md:52-56` records the gap plainly. The LRU eviction also removes the worst part of this gap: my probe shows that 50 000 keys now block no other visitor, thus one host with a `/64` prefix can only avoid its own limit. I accept the deferral. |
| 12 | A window could write two warning lines | Fixed | `IngestRateLimiter.java:140` passes `current.start()`, the real window of the count. |
| 13 | `AtomicInteger` with its full package name | Fixed | The import is at `IngestRouteTest.kt:35`. |

## One new MINOR finding (14)

`IngestRateLimiter.java:203-220`. `sha256Hex` calls `MessageDigest.getInstance` for each
request with a long key. My measurement of one call of `check`: 49 ns for a key of 64
characters, 535 ns for a key of 65 characters, 780 ns for 300 characters, 6 344 ns for
6 000 characters. An app whose user id holds more than 64 characters (a JWT subject, for
example) thus pays about 0.5 microseconds on each request. The cost is small against the
rest of the route, and an attacker cannot reach this path through the header any more,
because the route drops a value above 64 characters. The fix, when you want it: hold one
`MessageDigest` in a `ThreadLocal` and call `reset()` before each digest. This finding
blocks nothing.

## The test output of this round

- `./gradlew :kit:jvm-core:test`: 152 tests, 0 failures, 0 errors, 0 skipped.
- `./gradlew :kit:jvm-ktor:test --rerun`, five runs: 39 tests, 0 failures each run.
- `./gradlew build`: BUILD SUCCESSFUL in 3 m 0 s.
- My own mutations of this round: 6 applied, 6 fail the right test, each one reverted.
- The diff holds 6 files. No secret, no database file, no build output.

Each BLOCKER finding and each MAJOR finding is closed. Two MINOR findings stay open for
the IPv6 rule of issue #116, with a reason that I accept. I approve this pull request.

xamcross:
## Merge of main (#156)

Merge commit `34ea5b8` brings in `e4dd1d2` (issue #104,
`OCTOMETER_PATH_PATTERNS`). The README conflict of `kit/jvm-ktor` keeps
both sections. `IngestRouteTest.kt` merged with no conflict and keeps
each test of both sides.

Test results (JDK 21):
- `./gradlew :kit:jvm-core:test`: BUILD SUCCESSFUL, 224 tests, 0 failures.
- `./gradlew :kit:jvm-ktor:test`, run 3 times in sequence: BUILD SUCCESSFUL
  each time, 41 tests, 0 failures, 0 errors, 0 skipped each run.

CI checks of commit `34ea5b8`: Frontend, JVM modules, Monitor backend on
Windows, Secret scan, and Tracker all pass.

Issue #115, opened 2026-09-21
Show the view Anonymous sessions


xamcross:
**Blocked by #113 and #53.** The view reads the API of #113, the row "(anonymous)" of #52 opens it, and the session link opens the element view of #53. The component tests can start now with a stub response.

**Goal.** The route `/apps/:appId/sessions` shows one row for each anonymous session. A session row opens the actions of that visitor.

**Implementation steps.**
1. In `monitor/frontend`, add the route `/apps/:appId/sessions`. Keep `anonymous`, `firstPath`, and `page` in the query parameters.
2. Load `GET /api/apps/{appId}/sessions?anonymous=true&firstPath=&page=` through the poll store of D29.
3. Build the flat table of D28 with the columns session, first page, source, start time, clicks, and user.
4. Write the session cell as `<th scope="row"><a routerLink>` to `/apps/:appId/elements`, with the `sessionId` value.
5. Read `sessionId` from the query parameters in the element view of #53, and send it to the API.
6. Open this view from the row "(anonymous)" of the level 2 view of #52.

**Acceptance criteria.**
- [ ] The table shows one row for each anonymous session.
- [ ] The view shows each path and each host with Angular text interpolation only. It holds no `[innerHTML]`, no `[href]`, no `[attr.*]`, and no `DomSanitizer.bypass*`.
- [ ] A unit test asserts that a path `/a<img src=x onerror=alert(1)>` appears as text.
- [ ] `firstPath` filters the rows, and the view keeps it in the query parameters.
- [ ] The session cell links to `/apps/:appId/elements` with the `sessionId` value.
- [ ] The element view of #53 reads `sessionId` and shows the elements of that session.
- [ ] The row "(anonymous)" of the level 2 view opens this view.
- [ ] A session without a start row shows `(unknown)` and the time of its first click.
- [ ] A session start without `referrerHost` shows the source `(direct)`.
- [ ] The start time uses the `<time>` element of D31, and the column header names the zone.
- [ ] The pager has "Previous", "Next", and "Page 2 of 100", each with `aria-disabled`.
- [ ] Zero rows shows "No anonymous session yet."
- [ ] A 404 goes to the parent view with a message.
- [ ] No table element has a CSS `display` value, and each target is a minimum of 24 by 24 CSS px.
- [ ] `@for` tracks the stable key of D29.
- [ ] The view keeps `RefreshBar` of #92 between the `<h1>` and the `<table>`.
[...]

Issue #32, opened 2026-09-21
Add the app management page


xamcross:
**Goal.** On `/manage`, the owner adds an app, replaces its connection string, and deletes an app.

**Implementation steps.**
1. In `monitor/frontend`, show the app status list on the route `/manage`.
2. Add the form for `POST /api/apps` with the fields of D13.
3. Add the control that replaces the connection string with `PATCH /api/apps/{id}`.
4. Add the delete with a native `<dialog>` and `DELETE /api/apps/{id}`.
5. Move the focus to the `<h1>` after a delete.
6. Send each result message to the `role="status"` region.

**Acceptance criteria.**
- [x] The connection string field shows no stored value.
- [x] An API 400 message shows next to the field.
- [x] The `<dialog>` has the default focus on "Cancel", shows the event count, and deletes only after the owner types the name.
- [x] The `DELETE` request holds the header `Content-Type: application/json`.
- [x] After a delete the focus moves to the `<h1>`.
- [x] "App added" and "App deleted" go to the `role="status"` region.
- [x] Each status is an icon plus text.
- [x] No table element has a CSS `display` value, and each target is a minimum of 24 by 24 CSS px.
- [x] `npm test` passes in `monitor/frontend`.
- [x] A comment on this issue records the test output.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, D13, D32, D33.

xamcross:
Body edit on 2026-09-21: #19 closed. The first line now names only the open blockers: #15.

xamcross:
Body edit on 2026-09-22: #15 closed. No open blocker remains, thus the first line and the label `blocked` are removed. The work can start.

xamcross:
The pull request is #161. It adds the app management page at `/manage`.

`npm test` in `monitor/frontend` (Vitest, jsdom):

```
Test Files  22 passed (22)
     Tests  225 passed (225)
```

`npm run lint`: all files pass. `npm run build`: the production build
completes with no error.

A real-browser check (Chrome, 1280 CSS px) added an app, renamed it,
deleted it, and confirmed the focus moved to the `<h1>` after the delete.
The pull request text has the full evidence.

xamcross:
Closed on 2026-09-22 by the merge of #161 (commit 738707d). Each reviewer approved in the second review:
- Angular engineer: https://github.com/xamcross/octometer/pull/161#issuecomment-5774675281
[...]

PR #159, opened 2026-09-22
feat(monitor): show the user table of one app (#52)

### LINKED ISSUE #52: Show the user table (level 2)
**Goal.** The route `/apps/:appId/users` shows one row for each app user.

**Implementation steps.**
1. In `monitor/frontend`, replace the placeholder of the route `/apps/:appId/users`. Keep `page` and `q` in the query parameters.
2. Load `GET /api/apps/{appId}/users?page=&q=` through the poll store.
3. Build the flat table of D28: user, clicks, sessions, and unique elements.
4. Add the filter field for `q`.
5. Add the pager.
6. Add the D30 state for a 404.

**Acceptance criteria.**
- [x] `userId: null` shows as "(anonymous)" and links with `anonymous=true`. #115 changes this link to the view "Anonymous sessions".
- [x] A click on a row cell opens the link of the name cell.
- [x] The pager has "Previous", "Next", and "Page 2 of 100" with `aria-disabled`.
- [x] A long user id wraps without an ellipsis.
- [x] A 404 goes to the parent view with a message.
- [x] The status region announces a page change, and a refresh keeps the focus.
- [x] A reload and the Back button keep `page` and `q`.
- [x] The numbers use `Intl.NumberFormat`, right alignment, and `tabular-nums`.
- [x] `@for` tracks the stable key of D29.
- [x] The view keeps `RefreshBar` of #92 between the `<h1>` and the `<table>`.

**Related.**
- #50 gives the API.
- #92 gives the refresh bar.
- #115 changes the link of the row "(anonymous)" to the view "Anonymous sessions".

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, D28 to D31 and D33.

xamcross:
## Summary

The route `/apps/:appId/users` now shows one row for each app user. It
reads `GET /api/apps/{appId}/users?page=&q=` through the poll store, with
the query parameters `page` and `q`, so a reload and the Back button keep
them. The table copies the pattern of the app list: one flat `<table>`, a
caption, scoped headers, a scroll region, a row link, and right-aligned
numbers. The filter field sends the request after a debounce, or at once
on a submit. A 404 answer sends the view to `/apps`, with a message.

## Acceptance criteria

- [x] `userId: null` shows as "(anonymous)" and links with `anonymous=true`.
  Evidence: `users.html` line 44 (`rowQueryParams`), tests "shows
  '(anonymous)' for a row with userId null" and "links the anonymous row
  to the elements view with anonymous=true" in `users.spec.ts`.
- [x] A click on a row cell opens the link of the name cell. Evidence:
  `onRowCellClick` in `users.ts`, tests "opens the row link on a click on
  a different cell of the row" and the modifier-key / selection /
  secondary-button guard tests.
- [x] The pager has "Previous", "Next", and "Page 2 of 100" with
  `aria-disabled`. Evidence: `users.html` pager block, tests "shows
  'Previous', 'Next', and 'Page X of Y'" and the two `aria-disabled`
  tests.
- [x] A long user id wraps without an ellipsis. Evidence: `users.css`
  `th[scope='row']` rule, test "wraps a long user id with overflow-wrap
  anywhere, and no ellipsis".
- [x] A 404 goes to the parent view with a message. Evidence:
  `redirectOnAppNotFound` in `users.ts`, `notFoundAppId` input and
  message in `apps.ts` / `apps.html`, test "navigates to /apps with the
  app id, and does not render a table" and the two new `apps.spec.ts`
  tests.
- [x] The status region announces a page change, and a refresh keeps the
  focus. Evidence (updated in correction round 1): the announce runs
  after the answer arrives, not before the request, and it states the
  match count too. See `announceResult` in `users.ts`, tests "announces
  the number of users that match, after the answer arrives", "announces
  'No user matches' when the filter finds no row", and "announces the
  new page only after the answer arrives, not before the request"; test
  "keeps a focus outside tbody across a refresh".
- [x] A reload and the Back button keep `page` and `q`. Evidence: `page`
  and `q` are `input()` bindings read from the route query parameters,
  test "reads the page and q input from the route, so a reload keeps
  them".
- [x] No table element has a CSS `display` value, and each target is a
  minimum of 24 by 24 CSS px. Evidence: `users.css` holds no `display` on
  a table element, test "adds no CSS display value on a table element",
  and the two 24x24 target tests.
- [x] The numbers use `Intl.NumberFormat`, right alignment, and
  `tabular-nums`. Evidence: `formatCount` in `users.ts`, `.num` rule in
  `users.css`, test "formats each count with Intl.NumberFormat,
  right-aligned with tabular numbers".
- [x] `@for` tracks the stable key of D29. Evidence: `track row.userId`
  in `users.html`, test "tracks each row with the stable key userId: a
  second answer that drops a row and changes the order keeps the tr node
  of each surviving row" (this test fails when the track expression
  changes to `$index`, confirmed locally).
- [x] The view keeps `RefreshBar` of #92 between the `<h1>` and the
  `<table>`. Evidence: `users.html`, tests "puts the refresh bar directly
  after the heading" and "keeps the refresh bar before the table".
- [x] `npm test` passes in `monitor/frontend`. Evidence: 235 tests pass
  (19 test files) after correction round 1, see the run below and the
  pull request comment.
- [x] A comment on this issue records the test result. Evidence: a
  comment follows this pull request.

## Confirm results

- The design document `docs/superpowers/specs/2026-09-21-octometer-design.md`
  (D13, D28 to D31, D33, D44) and the API route
  `monitor/backend/src/main/kotlin/octometer/monitor/users/UserTotalsRoute.kt`
  on `main` gave the response shape: `page`, `pageCount`, `rows` with
  `userId` (nullable), `clicks`, `sessions`, `uniqueElements`.
- The 15 table rules of the second accessibility review of #124
  (`https://github.com/xamcross/octometer/pull/124#issuecomment-5767934545`)
  are each followed; each one has a matching test, and I broke several
  rules locally (the CSS `display` rule, the `track` key, the HTML
  escape, and the 404 status check) and saw each matching test fail.

- The filter field sends the request 400 ms after the last key stroke,
  or at once on a form submit (Enter, or the Filter button). Both paths
  reset the page to 1, and do nothing when the value equals the current
  `q`.
- The row link uses `[routerLink]` with the array form and
  `[queryParams]`, never `[innerHTML]`, `[href]`, `[attr.*]`, or
  `DomSanitizer.bypass*`. Angular URL-encodes the user id.
- A 404 answer sends the view to `/apps` with the query parameter
  `notFoundAppId=<id>`. The `Apps` component (`apps.ts` / `apps.html`)
  reads this parameter and shows the message "App `<id>` is not
  registered." above the table. This is the smallest change I could make
  to the shared `Apps` view; it does not touch `apps.css`,
  `app.routes.ts`, or `app.ts`.
- The pager shows the `page` field of the API answer, not only the route
  value, because the API can clamp an out-of-range page (D13).

## Test run

```
Test Files  19 passed (19)
     Tests  235 passed (235)
```

`npm run lint` and `npm run build` both pass with no warning. This count
is the result after correction round 1; see the pull request comment for
the full list of findings and fixes.

## Shared files changed

- `monitor/frontend/src/app/apps/apps.ts`: one new `input()` for
  `notFoundAppId`.
- `monitor/frontend/src/app/apps/apps.html`: one new conditional
  paragraph.
- `monitor/frontend/src/app/apps/apps.spec.ts`: three new tests for the
  message.

I did not run the app in a real browser (jsdom runs no layout); a
reviewer with a Chrome browser should measure the reflow at 1280, 640,
and 320 CSS px, per the design rule 15 of the accessibility review.

## Correction round 1

The Angular review and the accessibility review of this pull request
found one BLOCKER, six MAJOR, and twelve MINOR points. A later comment
on this pull request lists each finding, the fix, and the test. The
branch merges `origin/main` first (a merge commit, not a rebase).

Closes #52

Generated with Claude Code


xamcross:

I made my own clone of the repository. I ran `gh pr checkout 159` in it. I ran each command
myself. I made each judgement from my own evidence, and not from the pull request text.

## The commands that I ran

| Command | Result |
| --- | --- |
| `npm ci` | It fails on my machine: the npm of my portable Node 24.21.0 misses `node-gyp`. `npm ci --ignore-scripts` then installed 497 packages with 0 vulnerabilities. The fault is in my environment, not in the pull request. |
| `npm test`, run 1 | `Test Files 19 passed (19)` / `Tests 221 passed (221)` |
| `npm test`, run 2 | `Test Files 19 passed (19)` / `Tests 221 passed (221)` |
| `npm test`, run 3, the reverse file order | `Test Files 19 passed (19)` / `Tests 221 passed (221)` |
| `npm run build` | It completes with no warning. Initial total 272.01 kB raw, 75.68 kB transfer. The lazy chunk `users` is 7.05 kB raw, 2.48 kB transfer. |

For run 3 I added a temporary `vitest-base.config.ts` with a sequencer that reverses the file
order, and I ran `ng test --runner-config vitest-base.config.ts`. `users.spec.ts` ran as file 13
in the normal order and as the last file in the reverse order. I deleted that file after the run.
I saw no leak of a global, of a timer, or of a listener: the three runs give the same count.

I broke the code 10 times. Nine mutations gave the correct failure, and each one failed in the
test that covers the rule:

| The mutation | The failure |
| --- | --- |
| `track row.userId` to `track $index` | 1 failed: the stable-key test |
| `FILTER_DEBOUNCE_MS` 400 to 150 | 1 failed: the debounce test |
| `rowQueryParams` always `userId` | 2 failed: the two anonymous-row tests |
| the 404 check to 410 | 1 failed: the 404 test |
| `overflow-wrap: anywhere` removed | 1 failed: the long user id test |
| the `page: null` reset removed | 3 failed: the three filter tests |
| the app id removed from the not-found message | 1 failed: the new `apps.spec.ts` test |
| `announcer.announce` removed | 1 failed: the announce test |
| `currentPage` reads the route, not the answer | 2 failed: the two pager tests |
| the guard `value === this.qValue()` removed | **0 failed**: see MINOR 2 |

I found no test that cannot fail. Each new test holds a real assertion.

## The acceptance criteria

1. `userId: null` shows "(anonymous)" and links with `anonymous=true`: **PASS**.
   `users.html:41-50`; the mutation of `rowQueryParams` fails the two tests.
2. A click on a row cell opens the link of the name cell: **PASS**. `users.ts:134-144`. The
   modifier key, the secondary button, and a selection each stop the navigation.
3. The pager holds "Previous", "Next", and "Page 2 of 100" with `aria-disabled`: **PASS**.
   `users.html:66-82`.
4. A long user id wraps with no ellipsis: **PASS in `jsdom`**. `users.css:23-26`. I did not
   measure the reflow in a browser (rule 15 below).
5. A 404 goes to the parent view with a message: **PARTIAL**. The first load of an unknown app
   id redirects. A 404 after a good answer does not redirect. See MAJOR 2.
6. The status region announces a page change, and a refresh keeps the focus: **PASS**.
   `users.ts:189`; the focus test keeps the filter field focused across a refresh.
7. A reload and the Back button keep `page` and `q`: **PASS**. I checked this with the real
   router and the real routes, and not with a manual input. A load of
   `/apps/7/users?page=3&q=bob` sends exactly one request `/api/apps/7/users?page=3&q=bob`, the
   filter field shows `bob`, and the pager shows `Page 3 of 9`. A click on Next gives the URL
   `/apps/7/users?page=4&q=bob` and one request `page=4&q=bob`.
8. No table element holds a CSS `display` value, and each target is a minimum of 24 by 24 CSS
   px: **PASS for the table and for each button**. The filter field itself has no minimum size.
   See MINOR 3.
9. `Intl.NumberFormat`, right alignment, and `tabular-nums`: **PASS**. `users.ts:16`,
   `users.css:28-32`.
10. `@for` tracks the stable key of D29: **PASS**. `users.html:41`.
11. `RefreshBar` stands between the `<h1>` and the table: **PASS**. `users.html:1-2`.
12. `npm test` passes: **PASS**. 221 tests, three runs.
13. A comment on the issue records the test result: **PASS**.

## The 15 table rules of the second accessibility review of #124

Rules 1, 2, 3, 4, 5, 9, 10, 12, 13, and 14: **PASS**. Rules 6 and 7 do not apply, because this
table holds no time column and no status column. Rule 8: **PASS**; `users.css:49-64` gives the
pointer, the two states, and the `forced-colors` variant. Rule 11: **PASS for the row link and
for each button**; see MINOR 3 for the filter field. Rule 15: I did not measure the reflow at
1280, 640, and 320 CSS px, because I ran no browser. An accessibility reviewer must do it.

## The findings

### MAJOR 1 — A failed request shows no text. The user waits for ever.

`monitor/frontend/src/app/users/users.html:17-86` and
`monitor/frontend/src/app/users/users.ts:202-211`.

The template holds only two states: the data, or the text "Octometer reads the user list.". The
component reads `store.error()` for the 404 redirect alone. The banner of the shell
(`app.ts:50-52`) shows the health error alone, and never a data error. The API gives 400 for a
`q` above 200 characters, and 400 for a `q` with a control character
(`UserTotalsRoute.kt:99-106`). A user reaches both cases with the new filter field, or with a
shared link.

My evidence: I flushed a 400 answer on the first load. The page text stayed
`"Users Pause refresh The table refreshes each 10 seconds. Filter by user id Filter Octometer
reads the user list."`. The poll sends the same request again, thus the user sees no change. With a
table already on the screen, the old rows stay and no text appears, so the filter looks broken.

D30 asks for the state of a failed monitor API. The view gives none.

The fix:
1. In `users.ts`, add a computed error text:
   `protected readonly errorText = computed(() => { const e = this.store.error(); return e instanceof HttpErrorResponse && e.status === 400 ? ((e.error as { error?: string })?.error ?? 'The request is not valid.') : null; });`
   The API sends the body `{"error": "..."}` (`monitor/backend/src/main/kotlin/octometer/monitor/Errors.kt:7`), and not `{"message": "..."}`.
2. In `users.html`, show it under the filter form:
   `@if (errorText(); as text) { <p class="error-text" role="status">{{ text }}</p> }`.
3. Add two tests: one flushes a 400 with the body `{ error: 'The filter q is too long.' }` and
   expects that sentence in the DOM; one flushes a good answer after the 400 and expects the
   text to go.

### MAJOR 2 — A change of `appId` keeps the rows of the old app, and it drops the 404 state.

`monitor/frontend/src/app/users/users.ts:100-111` and `users.ts:202-211`.

The router keeps one component instance for the two URLs `/apps/7/users` and `/apps/8/users`,
because the route config is the same. I checked this with the real router: the instance of the
second navigation equals the instance of the first one. The store keeps `data()` across the
change, and `redirectOnAppNotFound` runs only while `data() === undefined`.

My evidence, with the real router and the real routes:
- After the navigation to `/apps/8/users`, the table still shows the row `app-7-user`.
- The 404 answer of app 8 gives no redirect: the URL stays `/apps/8/users`, and the table still
  shows `app-7-user`. The user reads the data of app 7 under the app id 8, with the breadcrumb
  and the page title of app 8.
- A second case: I held the request of app 7 in flight, then I moved to `/apps/8/users`. The
  store sent **no** request for app 8, and the late answer of app 7 filled the table of the
  view of app 8.

The browser history reaches this path: a jump from `/apps/7/users` to a history entry
`/apps/8/users` runs one client navigation with no other route between the two. The acceptance
criterion "A 404 goes to the parent view with a message" then fails. I call this MAJOR and not
BLOCKER, because the first load of an unknown app id still redirects.

The fix (it touches the shared store of #25):
1. Add `reset(): void` to `PollStore` in `monitor/frontend/src/app/poll/poll-store.ts`. It sets
   `data` to `undefined`, `dataError` to `undefined`, and `firstLoadPending` to `true`.
2. In the query-key effect of `users.ts`, call `this.store.reset()` when the `appId` part of the
   key changes, before `refresh()`.
3. Drop `&& this.store.data() === undefined` from `redirectOnAppNotFound`, or compare the app id
   of the answer instead.
4. Add a test: a navigation from app 7 to app 8 shows no row of app 7, and a 404 of app 8
   redirects to `/apps?notFoundAppId=8`.

### MAJOR 3 — A pager click or a filter submit during a request in flight sends no request.

`monitor/frontend/src/app/poll/poll-store.ts:122-141` (the shared store) with
`monitor/frontend/src/app/users/users.ts:103-109`.

`exhaustMap` drops each value while a request runs. The manual refresh of a user action goes
through the same `exhaustMap`. A user action that changes `page` or `q` thus sends no request
while the poll request is in flight.

My evidence: with the first data request in flight, I set the route to page 2. The store held
one pending request `/api/apps/7/users?page=1`, and it sent none for page 2. The answer of page
1 then filled the table, and the pager showed `Previous Page 1 of 5 Next` while the URL said
`page=2`. A new click on Next computes `currentPage() + 1 = 2`, thus it navigates to the same
URL and changes nothing. The view repairs itself only at the next poll tick. The same happens
for a filter change.

The fix: keep `exhaustMap` for the scheduled ticks of D29, and give the user action its own path
that cancels the request in flight:
```ts
merge(
  scheduledTicks$.pipe(exhaustMap(runOnce)),
  manualRefresh$.pipe(switchMap(runOnce)),
)
```
with `runOnce = () => request().pipe(map(...), catchError(...))`. A test must show that a page
change during a request in flight sends the request of the new page.

### MINOR 1 — The filter field has no length limit.

`monitor/frontend/src/app/users/users.html:6-13`. The API rejects a `q` above 200 characters
(`MAX_FILTER_LENGTH` in `UserTotalsRoute.kt:17`). Add `maxlength="200"` to the input. The
browser then stops the bad request, and MAJOR 1 gets a second guard.

### MINOR 2 — The guard of an unchanged filter value has no test.

`monitor/frontend/src/app/users/users.ts:178-181`. I removed the three lines of the guard, and
all 44 tests of `users.spec.ts` still passed. The pull request text names this behaviour as a
decision. Add a test: a submit with the value of the current `q` calls `router.navigate` no
second time. Without the guard, a click on Filter with the same text resets the page to 1.

### MINOR 3 — The filter field is smaller than 24 by 24 CSS px.

`monitor/frontend/src/app/users/users.css:74-82` gives each button `min-width: 24px` and
`min-height: 24px`. The `<input type="search">` gets no rule, thus it keeps the height of the
user agent. Rule 11 asks for a minimum of 24 by 24 CSS px for each target. Add the input to that
rule.

### MINOR 4 — The announce and the URL can disagree with the clamped page.

`monitor/frontend/src/app/users/users.ts:188-193`. `goToPage` announces
`Page <n> of <pageCount>` before the answer arrives, and it uses the page count of the last
answer. The server clamps an out-of-range page (`UserTotalsRoute.kt:154`), thus the URL can keep
`page=99` while the pager shows `Page 5 of 5`. Announce the page after the answer, or write the
clamped page back into the URL with `replaceUrl: true`.

### MINOR 5 — The 404 test uses an error body that the API never sends.

`monitor/frontend/src/app/users/users.spec.ts:601` flushes `{ message: '...' }`. The API sends
`{ "error": "..." }` (`Errors.kt:7`). The test still passes, because the code reads the status
alone. Use the real shape, so the fix of MAJOR 1 gets a correct test.

### MINOR 6 — The not-found message of the apps view is easy to miss.

`monitor/frontend/src/app/apps/apps.html:4-6`. The message is a plain `<p>` with no style and no
`role`. The focus moves to the `<h1>` after the redirect, thus a screen reader user hears
"Apps" and hears nothing about the app that is not registered. Give the paragraph
`role="status"`, or announce the message through the one status region of D33.

### MINOR 7 — The language of two comments.

`monitor/frontend/src/app/users/users.ts:95-97`: "in step with" is an idiom, and "without
touching" uses an `-ing` form as a noun. `users.ts:199-200`: "under the user" is a figure of
speech. The repository writes each comment in ASD-STE100 Simplified Technical English. Two
examples: "Copies the q query parameter into the filter field, for example after the Back
button. The value does not change while q does not change, thus the text of the user stays.";
"so a 404 after a good answer, for example a deleted app, keeps the table on the screen".

## The other checks

- **The links (R6)**: I grepped the whole frontend. It holds no `[innerHTML]`, no `[href]`, no
  `DomSanitizer`, and no `bypassSecurityTrust`. Each `[attr.*]` binding carries an ARIA value or
  a `datetime` value, never a URL. The row link uses `[routerLink]` with the array form and
  `[queryParams]`.
- **A hostile user id**: I rendered the ids `<script>alert(1)</script>`, `a/b`, `a?b=c`,
  `100%25 %2F x`, and `a&b#c`. Each one shows as text, the table holds no `<script>` element,
  and each link is correct:
  `/apps/7/elements?userId=%3Cscript%3Ealert(1)%3C%2Fscript%3E`,
  `/apps/7/elements?userId=a%2Fb`, `/apps/7/elements?userId=a%3Fb%3Dc`,
  `/apps/7/elements?userId=100%2525%20%252F%20x`, `/apps/7/elements?userId=a%26b%23c`.
- **The query of the apps page**: the redirect sends `{ notFoundAppId: <appId> }` alone. The
  default handling of the router drops `page` and `q`, thus a user id never reaches the URL of
  the apps view. I saw this in the navigation arguments.
- **The filter**: the debounce is 400 ms with fake timers (399 ms sends nothing, 400 ms sends
  the request). A submit sends the request at once. A new filter value removes `page` from the
  URL: I saw the URL `/apps/7/users?q=dan` and the request `page=1&q=dan`.
- **The shared files with #32**: the branch `issue-32-app-management-page` does not exist on the
  remote, and no open pull request changes a file under `monitor/frontend/src/app/apps/`. Thus I
  made no test merge with #32. I merged `origin/main` into this branch with
  `--no-commit --no-ff`: `Automatic merge went well`. No frontend file conflicts. I aborted the
  merge.
- **The scope**: the diff holds 8 files, all in `monitor/frontend`. It holds no secret, no
  database file, and no build output.
- **The change to `apps.ts` / `apps.html` / `apps.spec.ts`**: it is small, and it fits the
  purpose of #52.

## The verdict

Three MAJOR findings stay, thus I request a change. The table, the links, the escape of each
user id, the pager, the filter, and the tests are good work. The three MAJOR findings are about
the states around the table: a failed request, a second app id, and a user action during a
request in flight.

xamcross:
**Review (accessibility specialist): CHANGES REQUESTED**

## How I checked

I made my own clone and ran `gh pr checkout 159`. I ran `npm ci --ignore-scripts`,
`npm test` and `npm run build` in `monitor/frontend`. I then built the app, and I
served it with a small fake API on port 7532. The fake API answers `GET /api/health`,
`GET /api/apps` and `GET /api/apps/1/users?page=&q=`. It holds 60 users on two pages:
an anonymous row, a user id of 60 characters with no space, a user id with HTML
characters, and 57 normal rows. I drove a real Chrome browser with the extension.

Two notes on the method:

- The browser tab was a background tab, thus `document.hidden` was true and the poll
  gate of D29 blocked each request. I set `document.hidden` to false in the page.
  The patch opens the gate. It changes no style and no layout.
- I measured each width in a same-origin iframe of an exact CSS width. A 200 percent
  text zoom is `html { font-size: 32px }`, which is the Chrome font-size setting.

## Environment note for the maintainer

The portable npm of the scratchpad fails on `npm ci`. The error is
`Cannot find module 'node-gyp/bin/node-gyp.js'` in the `esbuild` postinstall step.
`npm ci --ignore-scripts` completes, and both `npm test` and `npm run build` then work.

## Build and test

```
Test Files  19 passed (19)
     Tests  221 passed (221)
Application bundle generation complete. [9.708 seconds]
```

## Measurements in Chrome

| Width | Page scroll bar | Table region | Header line | Number line |
| 1280 | none (scrollWidth 1280 = clientWidth) | no sideways scroll | 1 | 1 |
| 640 | none (640 = 640) | no sideways scroll | 1 | 1 |
| 320 | none (320 = 320) | sideways, 467 over 304 | 1 | 1 |
| 320, text zoom 200 % | none (320 = 320) | sideways, 925 over 304 | 1 | 1 |

- The scroll region holds `role="region"`, `tabindex="0"` and `aria-labelledby`. The
  name is "The users of this app, with the click totals of each user.".
- The user id of 60 characters takes 5 line boxes in a link of 128 px. The computed
  `text-overflow` is `clip`, thus no ellipsis appears.
- Each table element keeps a native `display` value: `table`, `table-header-group`,
  `table-row-group`, `table-row`, `table-cell`, `table-caption`.
- The `<tr>` holds one attribute, the Angular content attribute. No `tabindex`,
  no `role`, no ARIA attribute.
- A click on a number cell of row 4 opened `/apps/1/elements` of that user.
- The user id with HTML characters renders as text. The link element holds 0 child
  elements, and the href encodes `<` as `%3C`.

## axe scan

axe-core 4.13.0, tags `wcag2a`, `wcag2aa`, `wcag21a`, `wcag21aa`, `wcag22aa`,
`best-practice`:

| State | Violations | Incomplete | Passes |
| --- | --- | --- | --- |
| Users, wide window | 0 | 0 | 45 |
| Users, 640 | 0 | 0 | 45 |
| Users, 320 | 0 | 0 | 46 |
| Users, 320 + text zoom 200 % | 0 | 0 | 46 |
| Users, empty filter result | 0 | 0 | 40 |
| Apps with the 404 message | 0 | 1 | 42 |

The one incomplete item is `color-contrast` on the `aria-hidden="true"` status icon of
the level 1 view. That element comes from `main`, and this pull request does not touch it.

## Contrast

| Text | Background | Ratio |
| --- | --- | --- |
| Row link `#0000EE` | page `#FFFFFF` | 9.4:1 |
| Row link `#0000EE` | row hover `#EEF2F7` | 8.36:1 |
| Number `#000000` | page `#FFFFFF` | 21:1 |
| Number `#000000` | row hover `#EEF2F7` | 18.68:1 |
| Pager button `#000000` | button `#F0F0F0` | 18.43:1 |

Each value passes WCAG 1.4.3. No new rule removes a focus outline, thus the Chrome
default focus ring stays on each control.

## Keyboard

The tab order is: skip link, main nav, Pause refresh, the filter field, Filter,
the scroll region, each of the 30 row links, Previous, Next. 40 stops in total.

- Enter in the filter field sends the request. The URL became `?q=user-01`, and the
  row count went from 30 to 10.
- The focus stayed on the filter field after the filter.
- A press of Enter on Next moved the view to page 2. The focus stayed on the same
  Next button node.
- A disabled pager button keeps `aria-disabled="true"`, `disabled = false` and
  `tabIndex = 0`. It stays in the tab order. That choice is correct: the user finds
  the control, and the button explains its own state. Keep it.

## Acceptance criteria

1. `userId: null` shows "(anonymous)" and links with `anonymous=true` — PASS. The link
   name is "(anonymous)". The href carries `anonymous`.
2. A click on a row cell opens the link of the name cell — PASS. See the measurement above.
3. The pager has "Previous", "Next" and "Page X of Y" with `aria-disabled` — PASS.
4. A long user id wraps without an ellipsis — PASS. 5 line boxes, `text-overflow: clip`.
5. A 404 goes to the parent view with a message — PASS for the redirect and the text.
   See MAJOR 1 for the announcement.
6. The status region announces a page change, and a refresh keeps the focus — PASS.
   The region text became "Page 2 of 2.". See MINOR 4 for the text itself.
7. A reload and the Back button keep `page` and `q` — PASS. I went to page 2, then I
   filtered, then I pressed Back. The route, the field and the table came back correctly.
8. No table element has a CSS `display` value — PASS. Each target is a minimum of
   24 by 24 CSS px — FAIL. See MAJOR 3.
9. The numbers use `Intl.NumberFormat`, right alignment and `tabular-nums` — PASS.
   The computed style is `text-align: right`, `font-variant-numeric: tabular-nums`,
   `white-space: nowrap`. The text is "1,234,567".
10. `@for` tracks the stable key of D29 — PASS. `track row.userId` in `users.html:41`.
11. The view keeps `RefreshBar` between the `<h1>` and the `<table>` — PASS.
12. `npm test` passes — PASS. 221 tests, 19 files.
13. A comment on the issue records the test result — PASS.

## The 15 table rules

Rules 1, 2, 3, 4, 8, 9, 10, 11, 12, 13, 14 and 15 pass, with one exception in rule 11
(MAJOR 3). Rules 5, 6 and 7 do not apply: this table holds no empty cell, no time
column and no status column. Each data cell holds a number, and I found 0 empty cells.

## Findings

### BLOCKER 1 — the filter announces nothing

`monitor/frontend/src/app/users/users.ts:178` (`submitFilter`).

I typed "user-01" in the filter field and I pressed Enter. The table went from 30 rows
to 10 rows. The region `<div role="status">` of `app.html:20` held an empty text for
1.5 s, and a mutation observer on that region recorded 0 changes. I then typed
"no-such-user-zz". The whole table left the DOM. The region text was still "". The
focus stayed in the filter field the whole time.

The new code holds one call of `Announcer.announce`, at `users.ts:189` in `goToPage`.
No path announces a filter result.

This fails WCAG 2.2 success criterion 4.1.3 Status Messages, level AA: the result count
changes, the focus does not move, and no status message reaches a screen reader. It also
breaks D33, which gives the one status region to each user action.

Fix: announce the result after each new answer, not before the request. Write the text
from an effect that reads `store.data()`, for example:

```ts
// in the constructor
effect(() => {
  const page = this.store.data();
  if (page === undefined) return;
  this.announcer.announce(`${page.rows.length} users. Page ${page.page} of ${page.pageCount}.`);
});
```

Guard the first answer, so the load of the view does not speak. Add one test for the
filter path and one test for the empty result.

### MAJOR 1 — the 404 message on the apps view announces nothing

`monitor/frontend/src/app/apps/apps.html:4`.

I opened `/apps/999/users`. The view moved to `/apps?notFoundAppId=999`. The paragraph
holds "App 999 is not registered.". The element holds no `role` and no `aria-live`. The
status region text was "". The focus moved to `<h1>Apps</h1>`, which is correct for D28.
A screen reader reads "Apps, heading level 1" and stops. The message stands after
`<app-refresh-bar />`, thus the user meets the refresh bar text first.

The escape is correct. I passed an app id with HTML characters, and the paragraph held
0 child elements.

Fix: call `Announcer.announce` from `Apps` when `notFoundAppId` holds a value, or give
the paragraph `role="status"`. Keep `RefreshBar` at its place, because D28 asks for it.
Add one test.

### MAJOR 2 — a disabled pager button shows no visible state

`monitor/frontend/src/app/users/users.html:67-81`, `monitor/frontend/src/app/users/users.css:74-82`.

On page 1 the button "Previous" holds `aria-disabled="true"`. Chrome gives the two pager
buttons the same computed style: `color: rgb(0, 0, 0)`, `background-color: rgb(240, 240, 240)`,
`border-color: rgb(0, 0, 0)`, `opacity: 1`, no text decoration. A sighted user sees no
difference between "Previous" and "Next". A click on "Previous" does nothing, and the app
says nothing. A screen-reader user hears "dimmed". The two groups get different information.

Fix: add a visible state, for example:

```css
.pager button[aria-disabled='true'] {
  color: #595959;
  background-color: #f8f8f8;
  border-color: #767676;
  cursor: default;
}
```

`#595959` on `#f8f8f8` gives 7.0:1, thus WCAG 1.4.3 still passes. WCAG 1.4.3 applies
here, because the button stays in the tab order and is not a native `disabled` control.
Add a `forced-colors` variant, the same way as `users.css:58`. Add one test.

### MAJOR 3 — the filter field is 21.33 CSS px high

`monitor/frontend/src/app/users/users.html:6-13`, `monitor/frontend/src/app/users/users.css:74-82`.

`getBoundingClientRect()` of `#user-filter` gives 170 x 21.33 CSS px at a width of 1280.
The acceptance criterion asks for a minimum of 24 by 24 CSS px for each target. Rule 11
of the maintainer comment asks for `display: inline-flex` plus `min-width` and
`min-height`. The pull request gives that floor to the row link (24 px high), to the
Filter button (57.64 x 27.33) and to each pager button (79.88 x 27.33). The new field
is the one control with no floor.

To stay fair: the axe rule `target-size` passes this field, because WCAG 2.5.8 excepts a
control that the user agent sizes. The criterion of this issue is stricter than WCAG,
and the field is a new control of this issue.

Fix: add the field to the rule at `users.css:74`:

```css
.filter-form input,
.filter-form button,
.pager button { ... min-height: 24px; }
```

Add one test, the same shape as the test at `users.spec.ts:473`.

### MINOR 1 — two elements hold `aria-current="page"`

`monitor/frontend/src/app/users/users.html:74`. The last breadcrumb item
(`breadcrumb-nav/breadcrumb-nav.html:6`) already holds `aria-current="page"` for the
view "Users". The pager text adds a second one. A screen-reader user who asks for the
current page gets two answers. Use `aria-current="true"` on the pager text, or drop the
attribute and keep the plain text "Page 2 of 2".

### MINOR 2 — the landmark label "Pager"

`monitor/frontend/src/app/users/users.html:66`. "Pagination" is the common name of this
landmark, and each screen-reader user knows it. Rename the label.

### MINOR 3 — the row link does not name the target view

`monitor/frontend/src/app/users/users.html:44-49`. The link name holds only the user id,
or "(anonymous)". The level 1 view on `main` does the same, thus the pattern stays
consistent, and I do not block on it. When #115 gives the anonymous row its own view,
add a `visually-hidden` suffix to each link name, for example "Elements of".

### MINOR 4 — `goToPage` announces the wrong thing at the wrong time

`monitor/frontend/src/app/users/users.ts:189`. The method announces `Page 3 of 5.`
before the request goes out. D13 lets the API clamp an out-of-range page, thus the text
can state a page that the table never shows. The text also states a page number, and not
a result count. The fix of BLOCKER 1 repairs both points.

### MINOR 5 — `notFoundAppId` stays in the URL

`monitor/frontend/src/app/users/users.ts:209`. The parameter stays in the address of
`/apps`. A reload shows the message again, although the user already read it. Use a
router state value, or clear the parameter after the first paint.

## What I did not check

- I did not test with a real screen reader (NVDA, JAWS or VoiceOver).
- I could not measure the time of the announcement. Chrome throttles a timer in a
  background tab, thus each `setTimeout` of the `Announcer` runs late in my setup.
  Each finding above rests on the DOM state and on the code, not on a time value.

## Summary

The table itself is good work. The structure, the scroll region, the reflow at four
widths, the escape of a hostile user id, the stable track key and the keyboard order all
pass, and axe finds nothing in six states. The gap is the speech. The new filter is the
main control of this issue, and it says nothing to a screen reader. Fix BLOCKER 1 and the
three MAJOR items, then I approve.

xamcross:
@comment-round1.txt

xamcross:

I made a second clone, and I ran `gh pr checkout 159`. The head is `995af0c`. I ran my own
probes again against that head. I did not trust the response text.

## The three MAJOR findings

| # | Result | My evidence |
| --- | --- | --- |
| MAJOR 1, a failed request shows no text | **Fixed** | A 400 answer with the body `{ "error": "The filter q is too long." }` puts that exact sentence in the `role="status"` paragraph of `users.html:18-20`. A 500 answer and a network failure each give "Octometer cannot reach the API. Check the network connection.". A good answer clears the text. |
| MAJOR 2, a change of `appId` keeps the old rows | **Fixed** | With the real router and the real routes: the router still keeps one component instance for `/apps/7/users` and `/apps/8/users`. The table empties at once after the change. The request of app 7 in flight is cancelled (`cancelled === true`), and the store then sends `/api/apps/8/users?page=1`. The 404 of app 8 calls `router.navigate(['/apps'], { queryParams: { notFoundAppId: '8' } })` on the same instance. |
| MAJOR 3, a user action sends no request during a poll | **Fixed** | With a request in flight, a page change cancels it and sends `/api/apps/7/users?page=2` at once; the pager then shows "Page 2 of 5". A filter change does the same: it sends `/api/apps/7/users?page=1&q=carol`. |

The generation counter of `poll-store.ts:116,169-206` and the `reset()` of `poll-store.ts:234-239`
give the correct behaviour. A scheduled tick still waits for a request in flight, thus D29 stays.

## The seven MINOR findings

| --- | --- | --- | --- |
| 1 | No length limit on the filter field | Fixed | `maxlength="200"` at `users.html:11` |
| 2 | No test for the unchanged-value guard | Fixed | I removed the guard `value === this.qValue()`. One test failed: "does not navigate again when the filter is submitted with the unchanged value". I restored the code. |
| 3 | The filter field is below 24 CSS px | Fixed | `.filter-form input { min-height: 24px; box-sizing: border-box; }` at `users.css:74-77` |
| 4 | The announce comes too early | Fixed | `goToPage` (`users.ts:283-287`) announces nothing. The announce effect (`users.ts:186-199`) waits for a new reference of `data` or `dataError`. |
| 5 | The 404 test uses a wrong body shape | Fixed | `users.spec.ts:699,776` flush `{ error: '...' }`, the shape of `ErrorBody` |
| 6 | The apps message has no role | Fixed | `apps.ts:117-129` announces through the one shared status region of D33 |
| 7 | Two comments hold an idiom or an `-ing` noun | Fixed | `users.ts:155-157` and `users.ts:311-318` now read as Simplified Technical English |

## The commands that I ran

- `npm ci --ignore-scripts`: it completes. My portable npm needs that flag; the fault is in my
  environment.
- `npm test`, three runs: each one gives `Test Files 19 passed (19)` / `Tests 235 passed (235)`.
  Run 3 used a temporary `vitest-base.config.ts` with a sequencer that reverses the file order.
  I deleted that file after the run. I saw no leak between the files.
- `npm run lint`: `All files pass linting.`
- `npm run build`: it completes with no warning. Initial total 272.02 kB raw, 75.68 kB transfer.
  The lazy chunk `users` is 8.62 kB raw, 2.89 kB transfer.

## The verdict

No BLOCKER and no MAJOR stays. The three MAJOR findings and the seven MINOR findings of my first
review are correct now. I approve. Rule 15 of the table rules (the reflow at 1280, 640, and 320
CSS px) still needs a real browser; an accessibility reviewer owns that check.

xamcross:
**Review (second review, accessibility specialist): CHANGES REQUESTED**

I made a new clone, ran `gh pr checkout 159` (head `995af0c`), and repeated each check of
my first review in a real Chrome browser against my fake API. `npm test` gives
`Test Files 19 passed (19)` / `Tests 235 passed (235)`. `npm run build` completes.

## Round 1 items

| Item | Result | Evidence in Chrome |
| --- | --- | --- |
| BLOCKER 1 — the filter announces nothing | **Fixed** | The region spoke "10 users match. Page 1 of 1." after the rows went from 30 to 10, and "No user matches. Page 1 of 1." after the table left the DOM. The focus stayed in the filter field. The region stayed empty on the first load. |
| MAJOR 1 — the 404 message announces nothing | **Not fixed. It is worse.** | See BLOCKER A below. |
| MAJOR 2 — the disabled pager button | **Fixed** | "Previous" on page 1: `rgb(89, 89, 89)` on `rgb(248, 248, 248)`, border `rgb(118, 118, 118)`, `cursor: default`. "Next": `rgb(0, 0, 0)` on `rgb(240, 240, 240)`. The button keeps `aria-disabled="true"`, `disabled = false`, `tabIndex = 0`. |
| MAJOR 3 — the filter field size | **Fixed** | `getBoundingClientRect()` of `#user-filter` gives 170 x 24.00 CSS px at 1280, and 24.00 px at 320 with a text zoom of 200 %. |
| MINOR 1 — two `aria-current="page"` | Fixed | The pager span holds `aria-current="true"`. Only the breadcrumb holds `"page"`. |
| MINOR 2 — the label "Pager" | Fixed | `aria-label="Pages of the user list"`. |
| MINOR 3 — the row link does not name the target view | Not corrected. **I accept this.** | I wrote "I do not block on it". The level 1 view on `main` uses the same pattern. #115 gives the anonymous row its own view. Keep it until then. |
| MINOR 4 — `goToPage` announces too early | Fixed | A click on "Next" gave "30 users match. Page 2 of 2." after the table already showed row `user-031`. The count comes from the answer, thus the text waits for the answer. |
| MINOR 5 — `notFoundAppId` stays in the URL | The clear works, but it carries BLOCKER A. | The parameter left the URL. The message left the screen with it. |

## Reflow and axe, the same six states

| State | Page scroll bar | Table region | Header / number lines | axe |
| 1280 | none | no sideways scroll | 1 / 1 | 0 violations, 0 incomplete, 45 passes |
| 640 | none | no sideways scroll | 1 / 1 | 0, 0, 45 |
| 320 | none | sideways, 467 over 304 | 1 / 1 | 0, 0, 46 |
| 320, text zoom 200 % | none | sideways, 925 over 304 | 1 / 1 | 0, 0, 46 |
| Empty filter result | — | no table | — | 0, 0, 40 |
| Apps view | — | — | — | 0 violations, 1 incomplete, 42 passes |

axe-core 4.13.0, tags `wcag2a`, `wcag2aa`, `wcag21a`, `wcag21aa`, `wcag22aa`,
`best-practice`. The one incomplete item stays the `aria-hidden="true"` status icon of
the level 1 view on `main`. The user id of 60 characters still takes 5 lines and shows no
ellipsis.

## BLOCKER A — the not-found message speaks the word "undefined", then it disappears

`monitor/frontend/src/app/apps/apps.ts:118-128`.

I opened `/apps/777/users`. A mutation observer on `<app-root>` and on the status region
recorded this order:

```
MSG:    absent
MSG:    App 777 is not registered.
STATUS: App undefined is not registered.
```

The end state: the paragraph `.not-found-message` is absent, the status region holds
"App undefined is not registered.", and the URL holds no query parameter. I saw the same
result on a direct visit to `/apps?notFoundAppId=abc`.

Cause. The effect runs two times.

1. Run 1: `notFoundAppId()` is `"777"`. The effect sets `shownNotFoundAppId`, announces
   the correct text, and clears the query parameter.
2. Run 2: the router writes the missing parameter back as `undefined`, not as `null`.
   The guard `if (id === null)` at line 119 lets `undefined` pass. The effect then sets
   `shownNotFoundAppId` to `undefined`, so `@if (shownNotFoundAppId(); as id)` in
   `apps.html:4` turns false and the paragraph leaves the DOM. The effect also calls
   `announce()` a second time with `undefined` in the text. That second call replaces the
   first text, because `Announcer.announce()` cancels a pending timer.

A sighted user now sees nothing at all. A screen-reader user hears one wrong sentence.
Both groups lose the reason for the redirect. This is a regression against round 1: the
message was at least present and readable before.

Why the test does not catch it: `apps.spec.ts:170-171` writes
`setInput('notFoundAppId', null)` and calls that step "the router echoing the cleared
query parameter". The real router writes `undefined`.

Fix, one line at `apps.ts:119`:

```ts
if (!id) {
  return;
}
```

`!id` catches `null`, `undefined` and the empty string. Run 2 then returns at once,
`shownNotFoundAppId` keeps `"777"`, the paragraph stays, and the status region keeps the
correct sentence. Change the test at `apps.spec.ts:170` to `setInput('notFoundAppId',
undefined)`, and add one assertion on the status text after that step. Please check the
same guard shape in `users.ts`, where each `input()` can also receive `undefined`.

## MINOR A — the stated contrast of the disabled button

`monitor/frontend/src/app/users/users.css:96-101` and the response text state 7.0:1.
Chrome gives `rgb(89, 89, 89)` on `rgb(248, 248, 248)`, and I measure 6.6:1. The value
still passes WCAG 1.4.3. Correct the number in the comment.

## Summary

Three of the four items are fixed, and I confirm each one with my own measurement. The
fix of MAJOR 1 introduced BLOCKER A, and one guard repairs it. I approve after that one
change and its test.

xamcross:
## Correction round 2

I fixed the two points of the second accessibility review.

**BLOCKER A.** `apps.ts:119` now guards with `if (!id) return;`, not
`if (id === null) return;`. The old guard let `undefined` pass, so the
effect ran a second time and erased the not-found message. The test at
`apps.spec.ts:170` now sets `notFoundAppId` to `undefined`, and checks
the status text. I ran this test with the old guard first: it failed
with `expected undefined to be 'App 9 is not registered.'`. I then
applied the fix, and the test passed.

**MINOR A.** The comment in `users.css:96-101` and the round 1 comment
on this pull request stated a contrast of 7.0:1 for the disabled pager
button. Both now state the measured value of 6.6:1.

The branch merged `origin/main` (a merge commit, three commits, no
conflict).

## Test run

Three full runs of `npm test` pass, one of them in a reversed file
order: 19 test files, 235 tests, each time. `npm run lint` and
`npm run build` both pass with no warning. The CI checks of this pull
request are green: Frontend, JVM modules, Monitor backend on Windows,
Secret scan, Tracker.

Commit: 37e6ce9.

Generated with Claude Code

xamcross:
**Review (third review, accessibility specialist): APPROVE**

I made a new clone, ran `gh pr checkout 159` (head `37e6ce9`), and ran the probe of
BLOCKER A again in a real Chrome browser against my fake API.

## BLOCKER A — fixed

I opened `/apps/777/users`. A mutation observer on `<app-root>` and on the status region
recorded this order:

```
MSG:    absent
MSG:    App 777 is not registered.
STATUS: App 777 is not registered.
```

End state:

- The paragraph `.not-found-message` holds "App 777 is not registered.". It stays.
- The status region holds "App 777 is not registered.".
- No text holds the word "undefined". The second effect run returns at the new guard.
- The URL holds no `notFoundAppId` parameter, thus MINOR 5 still works.
- The focus sits on `<h1>Apps</h1>`, as D28 asks.

The guard at `apps.ts:121` is `if (!id)`, and the test at `apps.spec.ts:173` writes
`undefined`, plus an assertion on the status text. Both match my recommendation.

## MINOR A — fixed

`users.css:96-101` now states 6.6:1, the value that I measured.

## Commands

- `npm ci --ignore-scripts`: 0 vulnerabilities.
- `npm test`: `Test Files 19 passed (19)` / `Tests 235 passed (235)`.
- `npm run build`: completes.
- axe-core 4.13.0 on the apps view with the message: 0 violations, 42 passes. The one
  incomplete item stays the `aria-hidden="true"` status icon of the level 1 view on
  `main`, which this pull request does not touch.

Each item of my first review and my second review is now closed. I approve.

PR #161, opened 2026-09-22
feat(monitor): add the app management page

### LINKED ISSUE #32: Add the app management page

xamcross:
## Summary

This change adds the app management page at the route `/manage`. The owner
adds an app on this page. The owner renames an app on this page, or replaces
its connection string. The owner deletes an app on this page. The page
follows the rules of D11, D13, D32, and D33.

The page reads `GET /api/apps` one time, on load. The page reads the list
again after each successful change, or after a child form reports a stale
row (a 404). The page shows no `RefreshBar`: the list changes only after an
action of the owner, not on a timer.

Correction round 1 fixed each BLOCKER and MAJOR finding of the two round-1
reviews, and ten MINOR findings. See the correction comment on this pull
request for the full list.

## Acceptance criteria

- [x] The connection string field shows no stored value. The field has
  `type="password"` and `autocomplete="new-password"`. The field starts
  empty on each open of the edit form. The API never returns a stored
  value. A unit test reads `localStorage`, `sessionStorage`, the console
  (`log`, `info`, `warn`, `error`, `debug`), and the URL, and finds no
  connection string after a save.
- [x] An API 400 message shows next to the connection string field.
  `readApiErrorMessage` reads the fixed sentence of `Errors.kt`, and a
  fallback sentence for a network failure. The page shows that sentence
  near the form, linked with `aria-describedby`.
- [x] The `<dialog>` has the default focus on "Cancel", an accessible name
  from its own heading, and shows the event count (a sentence with no
  number for a `NEVER_POLLED` app). The dialog deletes the app only after
  the owner types the exact app name. A unit test confirms each point.
- [x] The `DELETE` request holds the header `Content-Type: application/json`.
  The component sends a body of `{}`. Angular's `HttpClient` then sets the
  header. A unit test reads the header from the captured request.
- [x] After a delete, the focus moves to the `<h1>`, and stays there once
  the dialog's queued `close` event later arrives. A unit test fires
  `close` from a real queued task and confirms the focus does not fall
  back to the removed trigger button.
- [x] "App added" and "App deleted" go to the `role="status"` region. Both
  actions call the shared `Announcer` service. The shell renders this one
  region at the root.
- [x] Each status is an icon plus text. The status cell repeats the pattern
  of the level 1 view (`apps.ts`). A new file, `status-format.ts`, now holds
  this pattern for both views.
- [x] No table element has a CSS `display` value. A unit test checks
  `getComputedStyle` for `table`, `caption`, `thead`, `tbody`, `tr`, each
  `th`, and each `td`, including the actions cell (the flex box now sits on
  a child `div.action-row`, not on the cell). Each target is a minimum of
  24 by 24 CSS px.
- [x] `npm test` passes in `monitor/frontend`. The suite has 295 tests in 22
  files, after the merge of `origin/main` (pull request #159). Each test
  passes, three runs, one in the reverse file order. See the evidence
  below.

## Extra rules this branch adds, where the issue gave no rule

- The connection string stays a secret at each step. Each form holds a
  write-only field for it. The API never returns it. No code logs it, and
  no code writes it to `localStorage` or to `sessionStorage`.
- A 400, a 404, a 409 (the name is taken), a 503, or a network failure
  shows a fixed or a general sentence as plain text. No new code uses
  `[innerHTML]` or `DomSanitizer.bypass*`. Each `[attr.*]` binding of this
  branch carries a fixed ARIA value or a generated id, and never a value
  from the server.
- A delete needs a typed exact match of the app name first. The dialog is a
  native `<dialog>`, never a browser `confirm()`.
- Each form control has a `<label for>`. A field error links to the
  connection string field with `aria-describedby`. A success and a failure
  both go through the one shared `Announcer`. The page adds no second live
  region.
- After a 404 on a save or a delete, the page reads the list again and
  drops the stale row, instead of keeping it on the screen.
- The focus moves to a sensible element after each transition of the edit
  form: the first field on open, the "Edit" button on cancel, and the
  "Edit" button again after a save.

## Design choices the issue did not state

- The list shows three columns: App, Status, and Actions. It does not show
  the click totals of the `/apps` view. D32 asks for "the status of each
  app." The `/apps` view already shows the level 1 metrics.
- One PATCH form covers the rename and the connection string replacement.
  The backend already accepts both fields on one request. The issue names
  only the connection string in step 3.
- The list reloads with a plain `GET /api/apps` call. It does not use the
  poll store of #25 and #92. The placeholder test "shows no refresh bar"
  stays green.
- The delete dialog shows the `clicks` field of `GET /api/apps` as the event
  count, and a text with no number for a `NEVER_POLLED` app (the pattern of
  the level 1 view). This is the nearest count that the API gives for one
  app.

## Test evidence

`npm test` in `monitor/frontend` (Vitest, jsdom), after the merge of
`origin/main`:

```
Test Files  22 passed (22)
     Tests  295 passed (295)
```

Three runs gave the same counts, one in the reverse file order (a
temporary `vitest-base.config.ts`, removed after the run).

`npm run lint`: all files pass. `npm run format:check`: all files use the
Prettier style. `npm run build`: the production build completes with no
warning.

### Manual check in a real browser (round 1, Chrome, the claude-in-chrome extension)

The backend: the Ktor monitor of this branch, mode `dev`, a free port
(18801, not the shared port 7431), a temporary `OCTOMETER_DATA_DIR`. The
folder was removed after the check. The frontend: `ng serve` on the default
dev port 4200. `proxy.conf.json` pointed at the backend port for the check
only, then it was reverted (no diff against `main`).

At 1280 CSS px:
- Added an app with the placeholder connection string
  `mongodb+srv://user:password@cluster0.example.invalid/exampledb`. The row
  showed the status "Not polled yet".
- Renamed the app with Edit and Save. The row showed the new name.
- Deleted the app. The dialog opened with the default focus on "Cancel".
  "Delete this app" stayed disabled until the typed text matched the app
  name. The delete succeeded. `document.activeElement` was the `<h1>` right
  after.
- Read the browser console after each step. No line held the connection
  string. The address bar stayed on `/manage` at each step.

At 320 CSS px: the extension's window-resize call did not change the real
viewport of this display (it stayed near 2552 CSS px wide, most likely a
maximised window on a high-DPI screen). As a check, the page body was
forced to a 320 px width in the page itself. The empty-state text and the
"Add an app" form stayed inside that width, with no horizontal overflow.
A second full add pass at this forced width did not complete: the local
safety classifier blocked each attempt to type the placeholder connection
string in that window. The table's narrow-width behaviour thus relies on
the unit tests (`overflow-x: auto`, `position: relative` on the scroll
wrapper) and on the same pattern already proven for the level 1 view. This
is a gap against the letter of the issue's read-first list, named here
rather than hidden.

The second accessibility review of this pull request later confirmed the
15 table rules, the axe-core scan, and the 320 CSS px reflow with its own
real-browser session; see the round-1 review comment on this pull request
for that evidence.

## Confirm results

- Node: v24.21.0 (the portable Node of the scratchpad). npm: 11.19.0 in
  round 1, 11.6.2 in the correction round. `npm ci` failed in this sandbox
  both times: the bundled npm's own `node-gyp` file is missing
  (`node_modules/npm/node_modules/node-gyp/bin/` is empty in the given Node
  build). The workaround: `npm ci --ignore-scripts`, then
  `node node_modules/esbuild/install.js` for the one install script the
  build and the tests need. This is a finding about the shared tool, not a
  change to the project.
- JDK 21 at `C:\Users\xamcr\.jdks\ms-21.0.9`. Used only for the manual
  backend check above, never for the frontend build or the frontend tests.

Closes #32

Generated with Claude Code


xamcross:
## Repair of the two failed checks

**Frontend.** The format step failed. Six files of the manage module did not match the Prettier style. `prettier --write` fixes each file. No file changes its logic.

**Secret scan.** The scan reads the full commit history of the pull request, not only the current file text. An early commit (`a92504b`) still holds its first placeholder connection string. A later commit already replaced the value in the working tree, but the old commit stays on the remote branch. This adds that exact placeholder to the `.gitleaks.toml` allowlist, next to the existing test URI. The value is not a real credential: the host has the reserved TLD `.invalid`, and the user name and the password are the literal words `user` and `password`.

The branch also merges `origin/main` (the Vitest 5 dependency update), with a merge commit.

`npm test` passes with 225 tests in 22 files, in three separate runs (one in the reverse file order). `npm run lint` and `npm run build` pass. All five CI checks are now green.

xamcross:
**Review (security and accessibility engineer): CHANGES REQUESTED**

I made my own clone, I ran `gh pr checkout 161`, and I ran each check myself. I
also ran the built app in Chrome against the Ktor backend of this branch, on a
free port, with a temporary data folder. I removed the folder at the end.

## The acceptance criteria

| # | Criterion | Result |
| - | - | - |
| 1 | The connection string field shows no stored value. | PASS |
| 2 | An API 400 message shows next to the field. | PASS |
| 3 | The dialog focuses "Cancel", shows the count, needs the typed name. | PASS |
| 4 | The `DELETE` request holds `Content-Type: application/json`. | PASS |
| 5 | After a delete the focus moves to the `<h1>`. | PASS |
| 6 | "App added" and "App deleted" go to the `role="status"` region. | PASS |
| 7 | Each status is an icon plus text. | PASS |
| 8 | No table element has a CSS `display` value. Each target is 24 by 24 px. | **FAIL** |
| 9 | `npm test` passes in `monitor/frontend`. | PASS |
| 10 | A comment on the issue records the test output. | PASS |

### The evidence of each PASS

**1. The secret.** In the running page I added an app with the placeholder
string of the reserved host `cluster0.example.invalid`. After the save I read:
the field value (empty), each console call (I wrapped `log`, `info`, `warn`,
`error`, `debug`, and `trace` before the save; 0 lines), `localStorage` (0
hits), `sessionStorage` (0 hits), `document.cookie` (0 hits), `location.href`
(0 hits), and `document.documentElement.outerHTML` (0 hits). The edit form
opens with an empty field, `type="password"`, and `autocomplete="off"`. The
SQLite file of the temporary data folder holds no part of the string, as D11
states.

**2. The error text.** I submitted a connection string with a rare marker in
the password part, plus a raw question mark. The page showed the fixed
sentence "Each option of a connection string needs a value." The element
`#add-app-error` holds text only (`children.length === 0`). The field carries
`aria-describedby="add-app-error"`. The marker appears in no error text, in no
page text, and in no HTML. I also forced a real 503: I replaced the secrets
folder with a plain file. The page showed "The secret store is not available.
Try again." and it echoed no part of the submitted value.

**3. The requests and the guard of D12.** The real requests gave `POST` 201,
`PATCH` 204, and `DELETE` 204. I captured the request options of the page: the
`POST` and the `DELETE` both carry `Content-Type: application/json`. Against
the backend I confirmed the guard: a `POST` with `text/plain` gives 415, and a
`DELETE` with no `Origin` header gives 403. Each answer carries
`X-Content-Type-Options: nosniff`, the CSP text of D12, and `Cache-Control:
no-store`. No answer carries an `Access-Control-Allow-Origin` header.

**4. The delete.** The dialog is the native `<dialog>`, and `:modal` is true.
The default focus is "Cancel". "Delete this app" stays disabled until the typed
text matches. A real Escape key closes the dialog and returns the focus to the
"Delete" trigger. A click on "Cancel" does the same. After a successful delete
`document.activeElement` is the `<h1>`, and the region `role="status"` holds
"App deleted".

**5. Accessibility.** I ran axe-core 4.10.2 with the tags `wcag2a`, `wcag2aa`,
`wcag21a`, `wcag21aa`, `wcag22aa`, and `best-practice`, in four states: the
list, the open edit form, the open delete dialog, and each of these at 1280 and
at 320 CSS px. Result: **0 violations** in each state. One `incomplete` item
(`color-contrast`) appeared for the open dialog; axe cannot measure a colour
behind the backdrop. The window-resize call of the extension does not change
the real window on this machine, so I measured the reflow in an iframe of an
exact CSS width. At 320 CSS px the page has no horizontal scroll bar
(`scrollWidth - clientWidth` is 0 for the document and for the body). The table
overflows its own `.table-scroll` container by 13 px, and that container has
`overflow-x: auto`, `tabindex="0"`, `role="region"`, and a name from the
caption. This is the pattern of rule 3. The target sizes: Edit 51x29, Delete
67x29, Cancel 70x29, Delete this app 117x29, Add app 86x32. The tab order runs
from the skip link to the nav, to the table container, to the row controls, and
then to the add form.

**6. The tests.** `npm ci --ignore-scripts` and `npm test` in
`monitor/frontend`: 22 files, 225 tests, all pass. I broke the code four times
and I saw the right test fail each time:

| Break | The failing test |
| - | - |
| `type="password"` → `type="text"` | "gives the connection string field type password and autocomplete off" |
| Remove the field reset in `onCreated` | "clears the form fields" (plus the sentinel test) |
| Add a `console.warn` of the connection string | "never writes the connection string to the console, or to the URL, after a save" and the same test of the edit form |
| `canConfirm` always true | "keeps the Delete button disabled until the typed text matches the app name exactly" |

## The findings

### BLOCKER 1 — a table cell gets a CSS `display` value

`monitor/frontend/src/app/manage/manage.css:38-43`, with
`monitor/frontend/src/app/manage/manage.html:30`.

The rule `.actions { display: flex; ... }` also matches the cell
`<td class="actions">`. In the running page I read
`getComputedStyle(document.querySelector('td.actions')).display`. The result is
`"flex"`, at 1280 CSS px and at 320 CSS px. Acceptance criterion 8 and rule 2
of the 15 table rules both forbid a CSS `display` value on a table element. A
browser takes the role of a cell from its CSS display type, so this cell is no
longer a cell for a screen reader, and the column "Actions" no longer applies
to it.

The test at `monitor/frontend/src/app/manage/manage.spec.ts:122-129` reads only
`table` and `tbody`, so it misses the cell. axe reads the DOM and not the
computed display, so the axe pass does not cover this rule either.

**The fix.** Put the flex box on a child element, and leave the cell with no
display value:

- `manage.html:30-33`: wrap the two controls in
  `<div class="action-row"> ... </div>` inside the cell.
- `manage.css:38`: rename the selector `.actions` to `.action-row`.
- `manage.spec.ts:122`: read each `th` and each `td` of the row too, and expect
  `table-cell` for each one.

### MAJOR 1 — the edit control loses the keyboard focus three times

`monitor/frontend/src/app/manage/edit-app-form.ts:51-61` and `:88-94`, with
`monitor/frontend/src/app/manage/edit-app-form.html:2` and `:33-36`.

I measured `document.activeElement` in the running page:

- I focused the "Edit" button and I activated it. The active element became
  `BODY`.
- I focused "Cancel" and I activated it. The active element became `BODY`.
- I focused "Save" and I activated it, and the save gave 204. The active
  element became `BODY`.

The reason: the `@if` removes the button that the user just activated, and no
code moves the focus. A keyboard user and a screen-reader user go back to the
start of the document three times, in a table that can hold many rows. This
breaks WCAG 2.4.3 Focus Order. The delete dialog does this correctly, so the
page is also inconsistent with itself.

**The fix.**

- In `open()`, move the focus to the name field of the new form. Use
  `afterNextRender` or an `effect` on the `viewChild` of the field.
- In `cancel()` and in `onUpdated()`, move the focus back to the "Edit" button
  of the same row. The delete dialog shows the pattern at
  `delete-app-dialog.ts:82-84`.
- Add one test for each of the three transitions.

### MINOR 1 — no test reads the browser storage

`monitor/frontend/src/app/manage/add-app-form.spec.ts:221-246`, and the same
test in `edit-app-form.spec.ts:191` and `manage.spec.ts:291`.

The pull request text and the body of commit a92504b both say that a unit test
found no connection string in `localStorage`. No test reads `localStorage` or
`sessionStorage`. I confirmed this with a search of `monitor/frontend/src`. My
browser check covers the two stores, but the claim of the text is wrong.

**The fix.** Add two assertions to each sentinel test: the joined content of
`localStorage` and of `sessionStorage` holds no part of the value. Correct the
text of the pull request.

### MINOR 2 — the console sentinel misses three console methods

`monitor/frontend/src/app/manage/add-app-form.spec.ts:223-225`.

The test wraps `console.log`, `console.warn`, and `console.error`. A call of
`console.info`, `console.debug`, or `console.trace` still writes the value to
the DevTools console, and the test stays green.

**The fix.** Wrap the six names in one loop, and read the calls of each spy.

### MINOR 3 — a cancel of the edit form keeps the typed value in memory

`monitor/frontend/src/app/manage/edit-app-form.ts:58-61`.

`cancel()` sets only `editing`. The signal `connectionString` keeps the typed
text until the next `open()`. The doc comment says "It discards each typed
value", and that is wrong. The save path is clean: `onUpdated()` clears the
signal at line 91. The DOM holds no copy, because the template uses a property
binding. This is a defence-in-depth point only.

**The fix.** Add `this.connectionString.set('');` to `cancel()`. Then the
comment is correct.

### MINOR 4 — the dialog has no accessible name

`monitor/frontend/src/app/manage/delete-app-dialog.html:3-4`.

The `<dialog>` element carries no `aria-labelledby` and no `aria-label`. A
screen reader announces "dialog" with no name. The rule `aria-dialog-name` of
axe does not apply to a native `<dialog>`, so the axe pass does not cover it.

**The fix.** Give the `<h2>` an id, and add
`[attr.aria-labelledby]` to the `<dialog>` with that id.

### MINOR 5 — the dialog states "0 events" for an app that the monitor never polled

`monitor/frontend/src/app/manage/delete-app-dialog.html:5-8`.

The dialog shows `clicks` as the event count. For a `NEVER_POLLED` app the
dialog reads "This removes 0 events of &lt;name&gt;." I saw this text in the
running page. The level 1 view shows "No data" for the same state, because the
count is unknown and not zero (rule 5 of the 15 table rules, and
`apps.html:33-37`).

**The fix.** For the status `NEVER_POLLED`, show a sentence with no number, for
example "Octometer holds no event of this app yet."

### MINOR 6 — one error id serves four fields

`monitor/frontend/src/app/manage/add-app-form.html:10`, `:23`, `:35`, `:45`.

Each of the four controls points `aria-describedby` at `add-app-error`. A
screen reader reads the connection string error on the name field and on the
collection list too. The user cannot tell which field the message describes.

**The fix.** Keep the link on the connection string field, or give each field
its own error id and set only the id of the field that the message names.

### MINOR 7 — the focus restore after a delete depends on a race

`monitor/frontend/src/app/manage/delete-app-dialog.ts:82-84` and `:104-109`,
with `manage.ts:71-78`.

`onDeleted()` calls `close()`. A browser fires the `close` event on a later
task, not at once. `onDialogClose()` then focuses the trigger button. The
parent focuses the `<h1>` before that task runs. In my run the `<h1>` took the
focus at 1756 ms and the row went away at 1775 ms, so the trigger button was
gone before the event arrived, and the result was correct. The margin is 19 ms.
A slower answer of `GET /api/apps` moves the trigger focus after the `<h1>`
focus, and the user then loses the focus when the row goes away. The unit test
cannot see this, because the `<dialog>` polyfill of the specs fires the `close`
event at once (`delete-app-dialog.spec.ts:39-42`).

**The fix.** Set a flag in `onDeleted()`. Skip the trigger focus in
`onDialogClose()` when the flag is set. Clear the flag after the close.

### MINOR 8 — `autocomplete="off"` and the password manager of the browser

`monitor/frontend/src/app/manage/add-app-form.html:19` and
`edit-app-form.html:22`.

The value `off` follows the issue. Chrome ignores `off` for a password field in
some cases, and it can offer to store the value in the password manager of the
browser. The value `new-password` states the intent more clearly.

**The fix.** Change the value to `new-password` in both forms, and update the
two tests.

## Check 2: the `.gitleaks.toml` change

`.gitleaks.toml:34-39`.

- The new entry is a second regex in the allow-list of the rule
  `mongodb-uri-password`. It demands the literal user name "user", the literal
  password "password", and the host `cluster0.example.invalid`. RFC 2606
  reserves the top-level domain `.invalid`, so that host cannot resolve.
- **A later real secret cannot pass this entry.** A real credential carries
  other text in the user part and in the password part, so the rule still
  fires. One small gap: the regex has no end anchor, and gitleaks matches a
  substring. A URI with the same two literal words and a host that starts with
  the placeholder host plus a dot also passes. Such a URI holds no real
  credential, so the risk is small. **The fix:** end the regex with `$`. The
  rule regex stops at the host, so the anchor still matches the real finding.
- **The old commit holds only the placeholder.** `git show a92504b` shows the
  placeholder eight times, in four spec files, and no other URI. I also read
  each commit of the branch with the rule regex. The branch history holds
  exactly two URIs: the old test fixture and this placeholder. The allow-list
  names both. The working tree holds no `.invalid` URI any more.
- **After a squash merge:** main then holds no commit with the placeholder. The
  job `secrets` of `.github/workflows/ci.yml:134-148` checks out the repository
  with `fetch-depth: 0`, so it reads each branch that GitHub still serves.
  **The entry must stay while the branch `issue-32-app-management-page` stays
  on GitHub.** After the merge **and** the delete of the branch, the entry can
  go. I suggest one follow-up issue: delete the second regex when the branch
  goes away.

## The scope and the other checks

- The diff touches 20 files: the frontend `manage` module, `apps/apps.ts`, a
  new `status-format.ts`, and `.gitleaks.toml`. The change to `apps.ts` moves
  six shared helpers into `status-format.ts` with no change of behaviour. No
  backend file changes. No secret, no database file, and no build output is in
  the diff.
- No new code uses `[innerHTML]`, `DomSanitizer`, or a dynamic `[attr.*]` value
  from the server. The error text and `lastError` both go through an
  interpolation, so Angular escapes them.
- The language of each new comment and of each commit body follows ASD-STE100.
- The shell holds one `role="status"` region. The page adds no second one.

Fix BLOCKER 1 and MAJOR 1. The eight MINOR findings are for you to judge; I
suggest MINOR 1, MINOR 2, and MINOR 3, because they guard the one secret of the
monitor.

xamcross:

I made my own clone of the repository. I ran `gh pr checkout 161` in it. I ran each
command myself. I made each judgement from my own evidence, and not from the pull
request text.

## The commands that I ran

| Command | Result |
| --- | --- |
| `npm ci` | It fails on my machine. The npm of the portable Node 24.21.0 misses `node-gyp`. `npm ci --ignore-scripts` plus `node node_modules/esbuild/install.js` installed 489 packages with 0 vulnerabilities. The fault is in my environment, and not in the pull request. |
| `npm test`, run 1 | `Test Files 22 passed (22)` / `Tests 225 passed (225)` |
| `npm test`, run 2 | the same two lines |
| `npm test`, run 3, the reverse file order | `Test Files 22 passed (22)` / `Tests 225 passed (225)` |
| `npm run format:check` | `All matched files use Prettier code style!` |
| `npm run build` | It completes with no warning. Initial total 273.79 kB raw, 74.33 kB transfer. The lazy chunk `manage` is 14.51 kB raw, 3.69 kB transfer. |

For run 3 I added a temporary `vitest-base.config.ts` with a sequencer that reverses the
file order, and I ran `ng test --runner-config vitest-base.config.ts`. I deleted the file
after the run. The printed file order changed between run 1 and run 3, and the counts
stayed the same. I saw no leak of a global, of a timer, or of a listener between the
files. The builder runs each file in one shared `jsdom` document: the schema of
`@angular/build:unit-test` gives `isolate` the default `false`.

## The mutation matrix

I broke the code 14 times. I ran the full suite after each change, then I restored the
file. 13 mutations gave the right failure. One gave none.

| The mutation | The failure |
| --- | --- |
| `track row.appId` to `track $index` | 1 failed: the stable-key test |
| the `Content-Type` header of the DELETE goes | 1 failed: the header test |
| `canConfirm` accepts each non-empty text | 1 failed: the typed-name test |
| the default focus on "Cancel" goes | 1 failed: the default-focus test |
| the add form keeps the connection string after a save | 2 failed |
| the edit form prefills the connection string field | 2 failed |
| `type="password"` to `type="text"` | 1 failed |
| `focusHeading()` goes from `onAppDeleted` | 1 failed: the delete-focus test |
| `readApiErrorMessage` reads `body.message` | 5 failed: each fixed-sentence test |
| `errorMessage.set(message)` to `set(null)` | 2 failed |
| the 24 px floor of the Delete trigger goes | 1 failed: the target-size test |
| `announce('App added')` goes | 1 failed: the announce test |
| the add form never resets the name field | 1 failed |
| the guard `if (this.pending()) { return; }` goes | **0 failed**: see MINOR 2 |

## The acceptance criteria of #32

1. The connection string field shows no stored value — **PASS**. `add-app-form.html:16-24`
   and `edit-app-form.html:19-26` give `type="password"` and `autocomplete="off"`. The
   edit form starts the field empty (`edit-app-form.ts:51-56`). `AppSummary` holds no
   connection string, and `AppRow` holds none. My probe after a save printed:
   ```
   the field value after the save = ""
   localStorage = []        sessionStorage = []        cookie = ""
   url = http://localhost:3000/        innerHTML holds the secret = false
   ```
   See MINOR 1 for the proof inside the branch.
2. An API 400 message shows next to the field — **PASS**. `add-app-form.ts:89-96` reads
   the fixed sentence of `Errors.kt` with `readApiErrorMessage`, and
   `add-app-form.html:53-55` shows it with an `aria-describedby` link.
3. The `<dialog>` starts with the focus on "Cancel", shows the event count, and deletes
   only after the owner types the name — **PASS**.
4. The `DELETE` request holds `Content-Type: application/json` — **PASS**.
   `delete-app-dialog.ts:18` and `:96-97`.
5. After a delete the focus moves to the `<h1>` — **FAIL in a real browser**. See
   BLOCKER 2.
6. "App added" and "App deleted" go to the `role="status"` region — **PASS**. Both call
   the shared `Announcer`, and `app.html:21` holds the one region.
7. Each status is an icon plus text — **PASS**. `manage.html:23-29` with the new
   `status-format.ts`.
8. No table element has a CSS `display` value — **FAIL**. See BLOCKER 1. Each target is a
   minimum of 24 by 24 CSS px — **PASS** for each button and for each field.
9. `npm test` passes in `monitor/frontend` — **PASS**. 225 tests, 22 files, three runs.
10. A comment on the issue records the test output — **PASS**.

## The findings

### BLOCKER 1 — the actions cell is a flex box, thus a table element holds a CSS `display` value

`monitor/frontend/src/app/manage/manage.html:30` with
`monitor/frontend/src/app/manage/manage.css:38-43`.

The rule `.actions { display: flex; … }` matches `<td class="actions">` of the same
component. The cell then stops to be a `table-cell`. The header comment at
`manage.css:1-4` states the opposite of the fact: "No table element gets a `display`
value in this file."

My evidence. I added a temporary spec that reads `getComputedStyle` of each table element
of the page. The run printed:

```
{"table":"table","thead":"table-header-group","tbody":"table-row-group","tr":"table-row",
 "th[scope=row]":"table-cell","td.status":"table-cell","td.actions":"flex","caption":"table-caption"}
```

The test at `manage.spec.ts:122-129` reads the `<table>` and the `<tbody>` alone, thus it
cannot catch this cell. D33 and acceptance criterion 8 both forbid the value.

The fix:
1. In `manage.html:30-33`, put the two controls in a child element:
   `<td class="actions"><div class="actions-group"><app-edit-app-form … /><app-delete-app-dialog … /></div></td>`.
2. In `manage.css:38`, change the selector to `.actions-group`, and keep the flex rules
   there.
3. Extend the test at `manage.spec.ts:122` over `thead`, `tr`, `th[scope="row"]`,
   `td.status`, `td.actions`, and `caption`. The rule then holds for each table element.

### BLOCKER 2 — after a delete the focus goes back to the Delete button, not to the `<h1>`

`monitor/frontend/src/app/manage/delete-app-dialog.ts:82-84` and `:104-109`, with
`monitor/frontend/src/app/manage/manage.ts:71-74`.

`onDeleted()` calls `close()` first, then `deleted.emit()`. The parent then calls
`focusHeading()`. The HTML Standard fires the `close` event from a queued element task
("close the dialog", the last step), and not at once. `onDialogClose()` thus runs **after**
`focusHeading()`, and it moves the focus back to the trigger button of the deleted row.
The list answer then removes that button from the DOM, and the focus falls to the
`<body>`.

My evidence. I copied the `<dialog>` stand-in of `delete-app-dialog.spec.ts`, and I gave
it the order of the standard: `close()` restores the focus at once, and the `close` event
comes from a queued task. The run printed:

```
right after the 204: activeElement = <h1 tabindex="-1">Manage</h1>
after the queued close task: activeElement = <button type="button" class="delete-trigger"…
```

The stand-in of the branch (`manage.spec.ts:18-21`, `delete-app-dialog.spec.ts:39-42`)
fires `close` at once. The order then flips, and the test at `manage.spec.ts:248` passes
for the wrong reason. I state my evidence exactly: I ran no browser. My evidence is the
order of the code plus the order of the standard.

The fix:
1. In `delete-app-dialog.ts`, mark the successful path, and skip the focus restore one
   time:
   ```ts
   private restoreFocusOnClose = true;

   private onDeleted(): void {
     this.pending.set(false);
     this.restoreFocusOnClose = false;
     this.dialogEl().nativeElement.close();
     this.announcer.announce('App deleted');
     this.deleted.emit();
   }

   protected onDialogClose(): void {
     if (!this.restoreFocusOnClose) {
       this.restoreFocusOnClose = true;
       return;
     }
     this.triggerButton().nativeElement.focus();
   }
   ```
2. Change each stand-in to `setTimeout(() => this.dispatchEvent(new Event('close')), 0)`.
   Make the focus test `await` one task before it reads `document.activeElement`. The test
   then fails for the real cause.

### MAJOR 1 — a failed request gives no text, no speech, and no way back

`monitor/frontend/src/app/manage/manage.ts:42-47`,
`monitor/frontend/src/app/manage/add-app-form.ts:89-96`,
`monitor/frontend/src/app/manage/edit-app-form.ts:96-103`, and
`monitor/frontend/src/app/manage/delete-app-dialog.ts:111-118`.

`reload()` drops each error with `error: () => undefined`. `readApiErrorMessage` returns
null for a network failure, thus each form sets the message to null and announces
nothing. The page runs no poll and holds no refresh control, thus the owner must reload
the browser page.

My evidence. I failed the first `GET /api/apps` with a status 0 answer. The page text
stayed:

```
"ManageOctometer reads the app list.Add an appNameConnection stringDatabaseCollection…Add app"
```

The pull request text says that the shell banner covers a network failure. That is not
correct. `poll/refresh-interval-state.ts:70-72` holds `filter(…)` plus `take(1)`, thus the
health stream completes at the first good answer. `error()` then stays undefined for the
rest of the session, and the computed `monitorApiError` at `app.ts:50` never shows the
banner again. A monitor that stops after the app start therefore gives the owner a silent
page. The tests `add-app-form.spec.ts:210` and `edit-app-form.spec.ts:162` lock in that
silence.

The fix:
1. Give `readApiErrorMessage` a fallback for a status 0 answer, for example
   `'The monitor did not answer. Try again.'`.
2. Show that text in each form, and announce it.
3. Add a failed state to `Manage`: a signal `listError`, a text in `manage.html`, and a
   "Try again" button that calls `reload()`.
4. Add one test for each path: a status 0 answer shows a text, and the button sends a new
   `GET /api/apps`.

### MAJOR 2 — a 404 on a delete or on a save keeps the stale row on the screen

`monitor/frontend/src/app/manage/delete-app-dialog.ts:111-118` and
`monitor/frontend/src/app/manage/edit-app-form.ts:96-103`, with the test
`delete-app-dialog.spec.ts:168-180`.

The API answers 404 with "The app is not registered." when a second window already removed
the app. The component shows the sentence, and it sends no new `GET /api/apps`. The list
keeps the row of an app that no longer exists. A second delete gives 404 again.

My evidence:

```
GET /api/apps requests after the 404 = 0
the row of the deleted app is still on the screen = traficio
```

The fix: add an output, for example `staleList`, and emit it from `onFailed` for the
status 404. `Manage` then calls `reload()`. Keep the sentence on the screen. Extend the
test at `delete-app-dialog.spec.ts:168` to expect one new `GET /api/apps`.

### MAJOR 3 — the edit form loses the focus after a save and after a cancel

`monitor/frontend/src/app/manage/edit-app-form.ts:59-61` and `:88-94`.

`cancel()` and `onUpdated()` both set `editing` to false. Angular then removes the
`<form>`, and the focus of the Cancel button or of the Save button goes with it. A
keyboard user lands on the `<body>`, and the next Tab press starts at the top of the page.
The delete dialog handles the same case correctly (`onDialogClose`), thus the page is not
consistent with itself.

My evidence:

```
the focus on the Cancel button of the edit form = true
after the cancel, activeElement = BODY
after the save, activeElement = BODY
```

The fix: move the focus back to the "Edit" button of the row. Read the button with a
`viewChild`, and call `afterNextRender(() => this.editToggle()?.nativeElement.focus())`
once the template shows it again. Add one test for the cancel path and one test for the
save path.

## The MINOR findings

### MINOR 1 — the secret sentinel reads no console line, and no test reads a browser store

`manage.spec.ts:291-329`, `add-app-form.spec.ts:221-246`, and
`edit-app-form.spec.ts:191-218`.

Each of the three tests walks `spy.mock.calls`. That array is empty in each run, thus the
loop body never runs, and the test asserts nothing about the console. My probe counted the
calls of the three spies after a full add: the total was **0**. No test reads
`localStorage`, and no test reads `sessionStorage`, although the body of commit `a92504b`
claims "no test sees it in the console, in local storage, or in the URL after a save".
`manage.spec.ts:328` reads `root().innerHTML`; that string never holds the `value`
property of an input, thus the assertion cannot fail either.

I grepped the whole `monitor/frontend/src` tree. It holds no `localStorage` and no
`sessionStorage`, thus the behaviour is right today. Only the proof is absent.

The fix: plant a control line in one test, for example `console.log('a control line')`,
and assert that the sentinel finds it. Then add
`expect(JSON.stringify(Object.entries(localStorage))).not.toContain(secret)`, and the same
line for `sessionStorage` and for `document.cookie`.

### MINOR 2 — the guard of a second submit has no test

`add-app-form.ts:62-64`. I removed the three lines of the guard, and all 225 tests still
passed. A disabled submit button also stops an implicit submit in a browser, thus the
guard is a second defence. My probe shows the behaviour of today is right: three submit
events during one request in flight gave 1 `POST /api/apps`, and the submit button was
disabled. Add that test, so a later change keeps the rule.

### MINOR 3 — the stand-in of `<dialog>` can leak into each later spec file

`manage.spec.ts:63-67` and `delete-app-dialog.spec.ts:70-74`.

Each `afterEach` calls `httpMock.verify()` before `removeDialogPolyfill()`. `verify()`
throws for an open request. The removal then never runs, and
`HTMLDialogElement.prototype.showModal` stays for each later spec file, because the runner
shares one `jsdom` document. Put the removal in a `finally` block, or call it first.

### MINOR 4 — `EditAppForm.cancel()` keeps the typed connection string in memory

`edit-app-form.ts:59-61`. `cancel()` sets `editing` to false alone. The signal keeps the
typed secret until the next `open()`. Add `this.connectionString.set('')` to `cancel()`,
and add one test.

### MINOR 5 — the pull request text says the new code uses no `[attr.*]` binding

`add-app-form.html:10,23,35,45`, `edit-app-form.html:13,25`, and
`delete-app-dialog.html:19` each hold `[attr.aria-describedby]`. Each value is a fixed id
or null, thus each binding is safe. Correct the sentence of the pull request text.

### MINOR 6 — a comment names an element that this page does not hold

`manage.css:34`. The comment says that `position: relative` makes the container the
containing block "of a hidden span". `manage.html` holds no hidden span. The comment comes
from the level 1 view. Correct it, or drop the rule with it.

### MINOR 7 — `reload()` has no guard for a second request

`manage.ts:42-47`. Two fast actions start two `GET /api/apps` requests. The older answer
can arrive last, and the table then shows stale rows. Keep the subscription and cancel it,
or send each reload through a `Subject` with `switchMap`.

### MINOR 8 — the allowlist entry of `.gitleaks.toml` outlives its commit

`.gitleaks.toml:38`. The second regex covers one placeholder of commit `a92504b` of this
branch alone. A squash merge drops that commit, thus the entry then widens the allowlist
for no reason. Remove the entry after the merge, or record the reason in #32.

## The other checks

- **The injection rules (D33).** I grepped the whole frontend. The source code holds no
  `[innerHTML]`, no `[href]`, no `[src]`, no `DomSanitizer`, and no `bypassSecurityTrust`.
  Each `[attr.*]` binding carries an ARIA value or a `datetime` value, and never a URL.
- **A hostile app name.** I rendered the name `<script>alert(1)</script>` plus an ``
  with an `onerror` attribute. The row header text held the exact characters, the row
  header held 0 child elements, the dialog heading held 0 child elements, and the page
  held 0 `<script>` elements and 0 `` elements.
- **The data flow.** The list comes from `GET /api/apps` in the level 1 row form of #18
  (`apps/app-row.ts`). The add sends `POST /api/apps` with the four fields of D13. The
  rename and the connection string replacement share one `PATCH /api/apps/{id}`, and an
  empty field means "keep the old value". The delete sends `DELETE /api/apps/{id}`. Each
  fixed sentence of `Errors.kt` shows as plain text for a 400, a 404, a 409, and a 503.
- **A request in flight.** Each form blocks a second submit; it does not queue it. I judge
  that correct for this page. A queue would send a second `POST` with a cleared form, and
  the owner expects one new app for one submit.
- **The shared component instance defect of #159.** It does not apply here. The route
  `/manage` carries no path parameter, thus one instance serves one URL.
- **The `exhaustMap` defect of #159.** It does not apply here. `Manage` sends a plain
  `HttpClient` request for each reload, with no `exhaustMap`.
- **The shared files.** This branch changes only `apps.ts` of the level 1 view, and that
  change is a pure move of the status helpers and the time helpers into the new
  `status-format.ts`. It changes no line of `app.routes.ts`, `app.ts`, `styles.css`,
  `apps.html`, or `apps.spec.ts`.
- **The test merge with #159.** I fetched `pull/159/head` as `pr159`. `git merge
  --no-commit --no-ff pr159` answered "Auto-merging
  monitor/frontend/src/app/apps/apps.ts" and "Automatic merge went well". No file
  conflicted. I then ran the suite and the build on the merged tree: `Test Files 22 passed
  (22)` / `Tests 269 passed (269)`, and the production build completed. The merged
  `apps.ts` keeps the import of `../status-format` and each call of the moved helpers. I
  aborted the merge.
- **The scope.** The diff holds 20 files. It holds no secret, no database file, and no
  build output.
- **The language.** Each new comment and each commit body follow ASD-STE100. I found two
  claims that the code does not support: the `[attr.*]` sentence (MINOR 5) and the "local
  storage" sentence of commit `a92504b` (MINOR 1).

## The verdict

Two BLOCKER findings and three MAJOR findings stay, thus I request a change. The API
contract, the write-only field, the typed confirmation, the fixed sentences, the stable
track key, and the mutation results are good work. The gap is at the edges: one CSS rule
breaks the table semantics, the focus after a delete lands on the wrong element in a real
browser, and a failed request leaves the page silent.

xamcross:
## Correction round 1

I made my own clone at `C:/Users/xamcr/octo-wt/issue-32c`. I ran `gh pr checkout 161` in it. I ran each command myself. `origin/main` had moved: it now holds pull request #159 (the level 2 view). I merged `origin/main` into the branch with a merge commit (`48ead8f`), not a rebase. No file conflicted.

### The BLOCKER findings

**BLOCKER (both reviews) — a table cell got a CSS `display` value.**
Fixed. `monitor/frontend/src/app/manage/manage.css` and `manage.html`. The flex box now sits on a child `<div class="action-row">` inside `<td class="actions">`, and the cell itself carries no `display` value. The test at `manage.spec.ts` now reads the computed `display` of `table`, `caption`, `thead`, `tbody`, `tr`, `th[scope="row"]`, `td.status`, and `td.actions`.

Mutation test: I put `display: flex` back on `.actions`, ran `manage.spec.ts`, and saw "adds no CSS display value on any table element" fail with `td.actions: expected 'flex' to be 'table-cell'`. I restored the file and saw the test pass again.

**Angular BLOCKER 2 — the focus after a delete lost a race with the dialog's queued `close` event.**
Fixed. `delete-app-dialog.ts:71-72` adds the flag `restoreFocusOnClose`. Line 134 clears it in `onDeleted`. Lines 106-112 (`onDialogClose`) skip the trigger-focus step once when the flag is false, and reset it. Each stand-in `<dialog>` of the spec files now fires `close` from a real `setTimeout(..., 0)`, not at once.

Mutation test: I removed the flag check from `onDialogClose`, ran `delete-app-dialog.spec.ts`, and saw "does not move the focus back to the trigger once the queued close event later arrives" fail (`expected <button> not to be <button>`). I restored the file and saw the test pass again.

### The MAJOR findings

**Angular MAJOR 1 — a failed request showed no text.**
Fixed. `app-registry-api.ts:38` (`NETWORK_FAILURE_MESSAGE`) and lines 51-52 give a fixed sentence for a status-0 answer. `manage.ts:42` adds a `listError` signal, `manage.html` adds a "Try again" button, and `manage.ts:65-66` announces the text. `add-app-form.ts`, `edit-app-form.ts`, and `delete-app-dialog.ts` already show and announce `errorMessage()`, so they show the new text with no further change.

Mutation test: I reverted the status-0 branch of `readApiErrorMessage` to `return null`, ran `manage.spec.ts`, `add-app-form.spec.ts`, and `edit-app-form.spec.ts`, and saw one test fail in each file. I restored the file and saw all three pass again.

**Angular MAJOR 2 — a 404 on a delete or a save kept the stale row on the screen.**
Fixed. `delete-app-dialog.ts:60` and `edit-app-form.ts` each add a `staleList` output, emitted on a 404 (`delete-app-dialog.ts:147-149`). `manage.html` binds `(staleList)="reload()"` on each child.

Mutation test: I removed the `staleList.emit()` call from both files, ran `delete-app-dialog.spec.ts`, `edit-app-form.spec.ts`, and `manage.spec.ts`, and saw one test fail in each of the first two files and two tests fail in `manage.spec.ts` (four in total). I restored the files and saw all four pass again.

**Angular MAJOR 3 and security MAJOR 1 — the edit control lost the focus three times.**
Fixed. `edit-app-form.ts:82` moves the focus to the name field in `open()`. Line 89 (`cancel()`) and line 123 (the save path) move the focus back to the "Edit" button, the pattern of `DeleteAppDialog.onDialogClose`.

Mutation test: I removed the three `afterNextRender` calls, ran `edit-app-form.spec.ts`, and saw all three new focus tests fail. I restored the file and saw all three pass again.

### The MINOR findings named in the correction brief

1. **The `$` anchor of the allow-list regex.** Fixed. `.gitleaks.toml:36-39`. Each regex now ends with `$`.
2. **`autocomplete="new-password"`.** Fixed. `add-app-form.html:12`, `edit-app-form.html:22`. Mutation test: reverted both to `off`, ran the two spec files, saw one test fail in each, restored, saw both pass.
3. **A cancel clears the typed secret from the signal.** Fixed. `edit-app-form.ts:88` (`cancel()`). Mutation test: removed that line, ran `edit-app-form.spec.ts`, saw "clears the typed connection string signal on Cancel, and keeps it out of memory" fail, restored, saw it pass.
4. **The dialog has an accessible name.** Fixed. `delete-app-dialog.html:3-4` (`[attr.aria-labelledby]`, `[id]` on the `<h2>`). Mutation test: removed both bindings, ran `delete-app-dialog.spec.ts`, saw "gives the dialog an accessible name from its own heading" fail, restored, saw it pass.
5. **"0 events" for a never-polled app.** Fixed. `delete-app-dialog.html:6-12`. A `NEVER_POLLED` app now shows "Octometer holds no event of NAME yet." (D30: "show '–', not 0"). Mutation test: reverted the template, ran the spec, saw the test fail on the old "0 events" sentence, restored, saw it pass.
6. **One error id for each field.** Fixed, with a choice: `aria-describedby` now names the connection string field alone, in both the add form and the edit form. I removed the link from the other fields (the name, the database, and the collection fields of the add form; the name field of the edit form). The API gives one sentence with no field name, so one consistent field gives a clearer link than two or four fields that may or may not apply. I updated the edit form's 404 test to read the connection string field's id, since it read the name field's id before.
7. **The race in the focus restore after a delete.** Fixed by the flag of BLOCKER 2 above; the flag removes the race itself, not only the one timing that a browser happened to hit in the reviewer's run.
8. **A storage assertion in the sentinel test.** Fixed. `manage.spec.ts`, `add-app-form.spec.ts`, and `edit-app-form.spec.ts` now assert that `JSON.stringify(Object.entries(localStorage))`, and the same for `sessionStorage`, hold no part of the secret. Mutation test: inside `add-app-form.spec.ts`, I wrote the secret into `localStorage` right before the assertion, ran the test, saw it fail and report the secret, then removed the line.
9. **The console sentinel covers `log`, `info`, `warn`, `error`, and `debug`.** Fixed, same three spec files. Each test also opens with a control line (`console.log('a control line')`) and asserts the spy caught it, so the loop below is not vacuous, the exact gap the Angular review named. Mutation test: I added `console.info(this.connectionString())` to `AddAppForm.onCreated`, ran `add-app-form.spec.ts`, and saw the sentinel test fail and report the secret (the old three-method sentinel would have missed a `console.info` call). I removed the line and saw the test pass again.
10. **A test for the second-submit guard.** Fixed. `add-app-form.spec.ts`: "sends exactly one POST when the form is submitted three times while a request is pending". Mutation test: I removed the `if (this.pending()) { return; }` guard from `add-app-form.ts`, ran the test, and saw it fail with "found 3 requests". I restored the file and saw it pass again.

### Other findings of the two reviews, corrected without a place in the brief

- **Angular MINOR 3 — the dialog stand-in could leak into a later spec file.** Fixed. `manage.spec.ts` and `delete-app-dialog.spec.ts`: `afterEach` now calls `httpMock.verify()` inside a `try`, and `removeDialogPolyfill()` inside the matching `finally`, so a thrown `verify()` cannot skip the removal.
- **Angular MINOR 5 — the pull request text claimed no `[attr.*]` binding.** Fixed the sentence in the pull request text: the branch does use `[attr.aria-describedby]` and `[attr.aria-labelledby]`, each with a fixed or a generated value, and never a value from the server.
- **Angular MINOR 6 — a stale comment named a "hidden span" that `manage.html` does not hold.** Fixed. `manage.css`: the comment now describes the containing-block rule in general terms.
- **Angular MINOR 7 — `reload()` had no guard against two requests in flight.** Fixed. `manage.ts:52-53`: `reload()` now keeps the `Subscription` and unsubscribes the previous one before it starts a new request, so a late answer of an old request cannot overwrite a fresh one.

### Not corrected

- **Angular MINOR 8, and the security review's gitleaks note — the allowlist entry for commit `a92504b` outlives that commit.** Not corrected. The entry must stay while the branch keeps that commit on GitHub, because the `secrets` job of `.github/workflows/ci.yml` checks out the repository with `fetch-depth: 0` and reads each branch GitHub still serves. Removing the entry now would break the secret scan of this same branch. I record this as a follow-up for the maintainer: drop the entry after the squash merge and the branch delete.

### Commands and results

- `npm ci --ignore-scripts`, then `node node_modules/esbuild/install.js` (the bundled npm's own `node-gyp` is missing on this Node build, the same finding as round 1).
- `npm test`, run 1 (after the merge of `origin/main`): `Test Files 22 passed (22)` / `Tests 295 passed (295)`.
- `npm test`, run 2: the same two lines.
- `npm test`, run 3, reverse file order (a temporary `vitest-base.config.ts` with a sequencer that reverses the file list, run with `ng test --runner-config vitest-base.config.ts`, removed after the run): `Test Files 22 passed (22)` / `Tests 295 passed (295)`.
- `npm run lint`: `All files pass linting.`
- `npm run format:check`: `All matched files use Prettier code style!`
- `npm run build`: completes with no warning.
- CI on commit `48ead8f` (the merge commit, the head of the branch): Frontend, JVM modules, Monitor backend on Windows, Secret scan, and Tracker all pass.

### New commits

- `4ba101d` — the table-cell CSS fix, the failed-list state, and the reload-after-stale-row wiring of `Manage`.
- `3b3d133` — the delete-dialog focus race, the dialog name, and the never-polled sentence.
- `317289c` — the edit-form focus moves, the cancel-clears-secret fix, and `autocomplete="new-password"`.
- `311c254` — the network-failure sentence, the error-id scope, and the sentinel-test fixes of the add form.
- `046c891` — the gitleaks regex anchor.
- `48ead8f` — the merge of `origin/main` (brings in pull request #159).

The head of the branch is `48ead8f`.

xamcross:
**Review (second review, security and accessibility engineer): APPROVE**

I made a second clone, I ran `gh pr checkout 161` (head `48ead8f`), and I ran
each check myself. I drove the built app in Chrome against the Ktor backend of
this branch, on a free port, with a temporary data folder. I removed the folder
at the end.

## The BLOCKER and the MAJOR of my first review

**BLOCKER 1 — a table cell got a CSS `display` value. FIXED.**
In the running page I read the computed display of each table element:

| Element | `display` |
| - | - |
| `table` | `table` |
| `caption` | `table-caption` |
| `thead` | `table-header-group` |
| `tbody` | `table-row-group` |
| `tbody tr` | `table-row` |
| `th[scope="row"]` | `table-cell` |
| `td.status` | `table-cell` |
| `td.actions` | `table-cell` |
| `.action-row` (a `<div>`) | `flex` |

The same result holds at 1280 and at 320 CSS px. The flex box now sits on the
child `<div class="action-row">` of `manage.html:31`, and `manage.css:39` names
that class. Mutation test: I renamed `.action-row` back to `.actions` and ran
`npm test`. One test failed with `td.actions: expected 'flex' to be
'table-cell'`. I restored the file.

**MAJOR 1 — the edit control lost the keyboard focus three times. FIXED.**
I measured `document.activeElement` in Chrome after each step:

| Step | Active element |
| - | - |
| A click on "Edit" | `input#edit-app-name-1` |
| A click on "Cancel" | `button.edit-toggle` |
| A successful "Save" | `button.edit-toggle` |
| A successful delete | `h1` |

## The eight MINOR findings

| - | - | - | - |
| 1 | The `$` anchor of the allow-list regex | FIXED | `.gitleaks.toml:37-38`: each regex ends with `$`. The job "Secret scan" of the head commit `48ead8f` passes. |
| 2 | `autocomplete="new-password"` | FIXED | The running page reports `new-password` on the add field. `edit-app-form.html:22` holds the same value. |
| 3 | A cancel clears the typed secret | FIXED | `edit-app-form.ts:88`. After a cancel the page HTML holds no part of the value, and the field is empty on the next open. |
| 4 | The dialog has an accessible name | FIXED | The `<dialog>` carries `aria-labelledby="delete-app-heading-1"`, and that id names the `<h2>` "Delete &lt;name&gt;?". |
| 5 | The text for a never-polled app | FIXED | The dialog reads "Octometer holds no event of &lt;name&gt; yet. This cannot be undone." |
| 6 | One error id for each field | FIXED | After a 400, `aria-describedby` is `add-app-error` on the connection string field, and `null` on the name, the database, and the collection fields. |
| 7 | The focus race after a delete | FIXED | I wrapped `HTMLElement.prototype.focus` for a full delete. The page made 2 focus calls, none of them on the trigger button, and the final active element is the `<h1>`. |
| 8 | The storage assertion and the console sentinel | FIXED | Each of the three sentinel tests reads `localStorage` and `sessionStorage`, spies on `log`, `info`, `warn`, `error`, and `debug`, and proves the spy with a control line first. |

One small point for a later change, and not a blocker: the sentinel still
leaves `console.trace` out. The five names cover each realistic case.

## The other checks

- **The secret.** After a real save of the placeholder string of the host
  `cluster0.example.invalid` I found no part of it in the console (0 lines), in
  `localStorage`, in `sessionStorage`, in a cookie, in the URL, or in the page
  HTML.
- **The error text.** A connection string with a rare marker and a raw question
  mark gave the fixed sentence "Each option of a connection string needs a
  value." The marker appears in no text and in no HTML.
- **The delete.** The dialog is modal, the default focus is "Cancel", the
  confirm button stays disabled until the typed name matches, a real Escape key
  and a click on "Cancel" both return the focus to the "Delete" trigger.
- **axe-core 4.10.2** (`wcag2a`, `wcag2aa`, `wcag21a`, `wcag21aa`, `wcag22aa`,
  `best-practice`): **0 violations and 0 incomplete items** in the list, in the
  open edit form, and in the open delete dialog, at 1280 and at 320 CSS px. At
  320 CSS px the page has no horizontal scroll bar; the table scrolls inside
  its own labelled container.
- **The tests.** `npm test` in `monitor/frontend`: 22 files, **295 tests**, all
  pass.

## The follow-up

The allow-list entry for commit `a92504b` must stay while the branch keeps that
commit on GitHub. Please delete the second regex of `.gitleaks.toml` after the
squash merge and the delete of the branch. I agree with the reason in the
response.

No BLOCKER and no MAJOR stays. I approve from the security and accessibility
side.

xamcross:

I made a second clone. I ran `gh pr checkout 161` in it. The head is `48ead8f`. I ran
each command myself, and I ran my round 1 probes again against the new code.

## The commands that I ran

| Command | Result |
| --- | --- |
| `npm ci --ignore-scripts` plus `node node_modules/esbuild/install.js` | 0 vulnerabilities. The bundled npm of the portable Node misses `node-gyp`; that is my environment. |
| `npm test`, run 1 | `Test Files 22 passed (22)` / `Tests 295 passed (295)` |
| `npm test`, run 2 | the same two lines |
| `npm test`, run 3, the reverse file order | `Test Files 22 passed (22)` / `Tests 295 passed (295)` |
| `npm run format:check` | `All matched files use Prettier code style!` |
| `npm run build` | It completes with no warning. Initial total 274.49 kB raw, 74.51 kB transfer. The lazy chunk `manage` is 16.41 kB raw, 4.12 kB transfer. |

The count of 295 matches the report of the correction agent. Run 3 used a temporary
`vitest-base.config.ts` with a sequencer that reverses the file order. I deleted the file
after the run. The working tree is clean after each experiment.

## The BLOCKER findings and the MAJOR findings

I ran each probe of round 1 again, with the `<dialog>` stand-in of the HTML Standard: the
focus goes back at once inside `close()`, and the `close` event comes from a queued task.

| My finding of round 1 | The probe output of today | Result |
| --- | --- | --- |
| BLOCKER 1 — a table cell held `display: flex` | `{"table":"table","caption":"table-caption","thead":"table-header-group","tbody":"table-row-group","tr":"table-row","th[scope=row]":"table-cell","td.status":"table-cell","td.actions":"table-cell","div.action-row":"flex"}` | **FIXED** |
| BLOCKER 2 — the focus went back to the Delete button | `right after the 204 = H1` / `after the queued close task = H1` / `after the list reload = H1` | **FIXED** |
| MAJOR 1 — a failed request was silent | `page text = "ManageThe monitor did not answer. Try again.Try again…"` / `the Try again button is present = true` / `GET requests after the click = 1` | **FIXED** |
| MAJOR 2 — a 404 kept the stale row | `GET after the delete 404 = 1` / `the 404 sentence is on the screen = true` / `GET after the save 404 = 1` | **FIXED** |
| MAJOR 3 — the edit form lost the focus | `after the Edit click = INPUT` / `after the Cancel = BUTTON edit-toggle` / `after the Save = BUTTON edit-toggle` / `after the list reload = BUTTON edit-toggle` | **FIXED** |

The flex box now sits on `<div class="action-row">` inside `<td class="actions">`
(`manage.html`, `manage.css:39-44`). The test at `manage.spec.ts:165-185` reads the
computed `display` of `table`, `caption`, `thead`, `tbody`, `tr`, `th[scope="row"]`,
`td.status`, and `td.actions`. The flag `restoreFocusOnClose`
(`delete-app-dialog.ts:72,107-112,134`) removes the race, and each spec stand-in now
fires `close` from `setTimeout(…, 0)` (`manage.spec.ts:25`,
`delete-app-dialog.spec.ts:46`), thus the test at `delete-app-dialog.spec.ts:201` can
fail for the real cause.

## The 8 MINOR findings of round 1

| # | The finding | Result |
| --- | --- | --- |
| 1 | The secret sentinel read no console line, and no test read a browser store | **FIXED**. Each of the three spec files opens with a control line and asserts the spy caught it. Each file also asserts that `localStorage` and `sessionStorage` hold no part of the secret. The sentinel covers `log`, `info`, `warn`, `error`, and `debug`. |
| 2 | The guard of a second submit had no test | **FIXED**. `add-app-form.spec.ts:131`: "sends exactly one POST when the form is submitted three times while a request is pending". |
| 3 | The `<dialog>` stand-in could leak into a later spec file | **FIXED**. `manage.spec.ts:75-79` and `delete-app-dialog.spec.ts:82-86` put `httpMock.verify()` in a `try` and the removal in the `finally`. |
| 4 | `cancel()` kept the typed connection string | **FIXED**. `edit-app-form.ts:88`. |
| 5 | The pull request text denied each `[attr.*]` binding | **FIXED**. The text now names `[attr.*]` and states that each value is fixed or generated. |
| 6 | A comment named a "hidden span" that the page does not hold | **FIXED**. `manage.css:32`. |
| 7 | `reload()` had no guard for a second request | **FIXED**. `manage.ts:45,53`: it keeps the `Subscription` and cancels the older request. |
| 8 | The gitleaks allowlist entry outlives its commit | **ACCEPTED, DEFERRED**. The agent gives a correct reason: the `secrets` job reads each branch that GitHub serves, thus the entry must stay while the branch holds commit `a92504b`. The agent records the follow-up for the maintainer. I agree. |

I also read the other changes of this round: `autocomplete="new-password"` on each
connection string field, the accessible name of the `<dialog>`, the sentence for a
`NEVER_POLLED` app, and the one error link on the connection string field. Each one is
correct, and none breaks a rule of D11, D30, D32, or D33.

## The one new MINOR finding

### MINOR 9 — a failed reload shows no text while the table is on the screen

`monitor/frontend/src/app/manage/manage.html:50-53` with `manage.ts:63-67`.

`listError` shows in the `@else if` branch alone. A first-load failure therefore shows the
sentence and the "Try again" button, which is the fix of MAJOR 1. A **later** reload that
fails keeps the table, sets `listError`, and shows no text. My probe:

```
R2-F the sentence is on the screen = false
R2-F the table is still on the screen = true
```

`onListFailed` still calls `announce(message)`, thus a screen reader hears the failure.
The sighted owner sees the old rows with no mark. The impact is small: each action of the
owner shows its own sentence, and the stale rows are the rows of the last good answer, as
D30 asks. I do not block the approval on it.

The fix for a later round: show the same paragraph above the table, for example
`@if (listError(); as message) { <p class="list-error">{{ message }}</p> }` before the
`@if (apps(); …)` block, with the "Try again" button next to it.

## The other checks

- I grepped the whole `monitor/frontend/src` tree again. It holds no `[innerHTML]`, no
  `[href]`, no `[src]`, no `DomSanitizer`, and no `bypassSecurityTrust` outside a spec
  file.
- The branch merges `origin/main` with the merge commit `48ead8f`, which brings in pull
  request #159. The suite covers both views in one run: 295 tests in 22 files.
- The shared file `apps.ts` keeps the pure move of the status helpers and the time helpers
  into `status-format.ts`. The branch changes no line of `app.routes.ts`, `app.ts`,
  `styles.css`, or the files of the level 2 view.
- The working tree stayed clean after each probe. I pushed nothing, and I changed no
  branch.

## The verdict

No BLOCKER finding and no MAJOR finding stays. Each of my five findings of round 1 is
fixed, and I confirmed each one with my own probe and not with the report. Seven MINOR
findings are fixed, and the eighth is deferred with a correct reason. One new MINOR
finding stays, and it does not block. I approve.

PR #155, opened 2026-09-22
feat(kit): publish the JVM kit through JitPack

xamcross:
## Summary

This change adds the `maven-publish` plugin to `kit/jvm-core`, `kit/jvm-mongo`,
and `kit/jvm-ktor`, with the group id `com.github.xamcross.octometer` and the
artifact ids of design decision D25. It adds `jitpack.yml` at the root, with
a JDK 21 build and an install command that publishes only the three kit
modules. This change does not add a git tag and does not make a GitHub
release; issue #39 owns the release workflow.

## Acceptance criteria

- [x] The coordinates in the comment come from the JitPack build log, not
  from a guess. Evidence: the issue comment on #37 quotes the log lines
  under "Build artifacts".
- [x] The JitPack build log of a test build shows success. This branch has
  no tag yet, since a tag and a release stay with the owner and with issue
  #39. The evidence is a build of the head commit `876f956` instead:
  `BUILD SUCCESSFUL in 23s`, `Exit code: 0`, and three lines under "Build
  artifacts". Log: https://jitpack.io/com/github/xamcross/octometer/876f956/build.log
- [x] A clean Gradle project outside the repository resolves
  `octometer-kit-ktor`. Two checks, both outside the clone: one against
  `mavenLocal()` after `./gradlew publishToMavenLocal`, and one against the
  real JitPack repository, at the head commit `876f956` (see the next item
  for both outputs).
- [x] `./gradlew dependencies` in that project shows `octometer-kit-core`
  as a transitive dependency of `octometer-kit-ktor`. From the JitPack
  network check:
  ```
  compileClasspath - Compile classpath for source set 'main'.
  \--- com.github.xamcross.octometer:octometer-kit-ktor:876f956
       +--- com.github.xamcross.octometer:octometer-kit-core:876f956
       \--- org.jetbrains.kotlin:kotlin-stdlib:2.4.20
  ```
- [x] The build log shows no artifact of the monitor and no artifact of a
  tool. The log lists exactly three artifacts:
  `octometer-kit-core`, `octometer-kit-ktor`, `octometer-kit-mongo`.
- [x] A comment on the issue records the log URL and the coordinates. See
  the issue comment.

## Confirm results

- JitPack multi-module group id and the `jdk:` list form: confirmed on
  `docs.jitpack.io`, page "Guide to publishing libraries", on 2026-09-22.
  A multi-module Gradle build publishes under
  `com.github.<user>.<repo>:<module>:<version>`.
[...]

PR #152, opened 2026-09-22
docs(contract): align four texts with design version 1.1

xamcross:
## Summary

This pull request aligns four texts of `contract/README.md` with design version 1.1, per
issue #139. Each change is a text correction. No rule changes its normative meaning.

- Rule C40 now states the five omit cases of `referrerHost`, from design decision D42.
- The Purpose paragraph now states five table views, with the two views of design decision
  D44.
- Rule C19 now states the 204 answer of a daily anonymous cap, from design decision D43.
- Rule C38 now states that rule C33 runs first, with a new example file for the case
  `octo:session-start` with a trailing space.
- Rule C42 now defines an invalid route pattern, in line with `prepareRoutes` of
  `kit/tracker/src/path-match.ts`.
- The contract version moves from 1.1 to 1.2, with a new paragraph in the change history.

## Acceptance criteria

- [x] Rule C40 holds the five omit cases, and it agrees with design decision D42 word for
  word in the list of cases. Evidence: `contract/README.md`, rule C40, the sentence that
  starts "The tracker sends no `referrerHost` field for one of five cases".
- [x] No line of `contract/README.md` says "three table views". Evidence: the Purpose
  paragraph now says "five table views".
- [x] Rule C19 gives 429 for a rate limit and 204 for a daily anonymous cap, as design
  decisions D20 and D43 state. Evidence: `contract/README.md`, rule C19.
- [x] Rule C38 states which rule runs first, and an example file shows the result for
  `octo:session-start` with a space at the end. Evidence: `contract/README.md`, rule C38,
  and `contract/examples/ingest-invalid-C38-session-start-trailing-space.json`.
- [x] Rule C42 defines an invalid route pattern, and the definition agrees with
  `prepareRoutes` of `kit/tracker/src/path-match.ts` on `main`. Evidence:
  `contract/README.md`, rule C42, the paragraph "**An invalid route pattern.**".
- [x] `./gradlew :kit:jvm-core:test` passes. Evidence: `BUILD SUCCESSFUL in 12s`, run with
  JDK 21 (`C:\Users\xamcr\.jdks\ms-21.0.9`) in this clone.
- [ ] A comment on this issue records the commit. This item needs the commit hash of this
  pull request; a separate issue comment follows this pull request.

## Notes for the maintainer

- `kit/tracker/src/index.ts` does not yet send `document.referrer`; issue #108 is still
  open. Rule C40 states the five omit cases as design decision D42 states them, ahead of
[...]

PR #162, opened 2026-09-22
feat(tracker): send the session start from start()

### LINKED ISSUE #107: Send the session start from the tracker
**Goal.** `start()` sends one entry `octo:session-start` at once for a new session, thus the monitor sees the first page of a visitor who leaves after 2 seconds.

**Evidence.** Decided on 2026-09-21: the browser reviewer showed that a delay of 5 seconds loses the visitor who leaves after 2 seconds. That visitor is the subject of the owner question, thus the session start goes at once. The rate limits and the caps control a flood.

**Implementation steps.**
1. In `kit/tracker`, make `start()` send one entry `octo:session-start` in its own request when the session id is new.
2. Give that entry `ageMs: 0` and the `path` of #106. Wait for no click and for no `flushIntervalMs`.
3. Send the request again one time after a network error, a 5xx response, or a 429 response.
4. Send no session start when `start()` reads an existing id from `sessionStorage`. With a blocked `sessionStorage`, keep the id in a module variable for the life of the document.
5. While `document.prerendering` is true wait for `prerenderingchange`, and while `document.visibilityState` is `hidden` wait for `visibilitychange`. Run one time only.
6. Send nothing while `navigator.webdriver` is true, and add the option `ignoreWebdriver` that turns this filter off.

**Acceptance criteria.** The tests use jsdom and fake timers.
- [x] A new session id sends one request with one entry `octo:session-start` before any click.
- [x] The entry holds `ageMs: 0` and the `path` of rule R3.
- [x] The request waits for no click and for no `flushIntervalMs`.
- [x] A 500 response makes exactly one more attempt, and a second failure throws no error into the page.
- [x] A second `start()` with the same `sessionStorage` id sends no session start.
- [x] With a blocked `sessionStorage`, one document sends exactly one session start.
- [x] With `document.prerendering` true the tracker sends nothing until `prerenderingchange`, then it sends one time.
- [x] With `document.visibilityState` `hidden` the tracker waits for `visibilitychange`, then it sends one time.
- [x] With `navigator.webdriver` true the tracker sends nothing.
- [x] With `ignoreWebdriver` true and `navigator.webdriver` true the tracker sends the entry.
- [x] A `data-octo` value that starts with `octo:` records no click (rule C38).
- [x] `npm test` passes in `kit/tracker`.

**Related.**
- #106 gives the path rule.
- #13 closed on 2026-09-21. Its reviewer found that `getOrCreateSessionId()` runs inside the flush today. Step 1 must move that call into `start()`, else the entry cannot go at once.
- #108 adds the source field to the same entry.
- #117 adds the server-side robot filter, which is the control.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, the contract version 1.1 of #101, and the rules T1, T2, T3, T4, T5, and C38.

xamcross:
## Summary

`start()` now resolves the session id at once and sends one entry
`octo:session-start` for a new session, in its own request, before any
click (contract rule C38, design decision D41). It waits for a visible,
non-prerendering document, then it sends the entry one time; a `stop()`
call before that moment cancels the wait. The `navigator.webdriver`
filter blocks the whole tracker, no click and no session start, unless
the new option `ignoreWebdriver` is `true`. A blocked `sessionStorage`
keeps the session id in a module variable, so the tracker still sends
one session start for the life of the document.

**Correction round 1** (maintainer decision, 2026-09-22): a new session
id no longer enters `sessionStorage` inside `start()`. It enters the
store only when the session start request goes out, after the wait
below. See "Decisions where the issue was silent", item 9.

Closes #107

## Acceptance criteria

- [x] A new session id sends one request with one entry `octo:session-start` before any click.
  Evidence: `kit/tracker/src/index.test.ts` → `the session start` → `sends one request with one octo:session-start entry, before any click and with no wait`. This holds for a visible document. A hidden or a prerendering document can still let a click go out first; see the README section "The session start".
- [x] The entry holds `ageMs: 0` and the `path` of rule R3 (contract rule C39).
  Evidence: same describe → `adds the path field of rule C39 when the routes option gives one`, `adds no path field without the routes option`.
- [x] The request waits for no click and for no `flushIntervalMs`.
  Evidence: the test above asserts a fetch call with no `clickElement()` call and no `vi.advanceTimersByTime()` call.
- [x] A 500 response makes exactly one more attempt, and a second failure throws no error into the page.
  Evidence: `the retry rule of the session start` → `retries one time after a 500 response, after a wait`, `drops the entry after a second 500 response, and throws no error into the page`.
- [x] A second `start()` with the same `sessionStorage` id sends no session start.
  Evidence: `sends no session start when a second start() reads the same stored id (a page reload)`, plus the new test `sends no second session start on a reload after the id enters storage at the send, for a document that started hidden`.
- [x] With a blocked `sessionStorage`, one document sends exactly one session start.
  Evidence: `a blocked sessionStorage (the module-level fallback id)` → `sends exactly one session start for the life of the document, also across a stop() and a later start()`, plus the new test `gives one session start with one id, not two, when sessionStorage reads succeed but each write fails` (reliability review MINOR 1, a partial block).
- [x] With `document.prerendering` true the tracker sends nothing until `prerenderingchange`, then it sends one time.
  Evidence: `the wait for a visible, non-prerendering document` → `sends nothing while document.prerendering is true, then sends one time after prerenderingchange`, plus the new tests for a `stop()` during the wait and for a `stop()` and a later `start()`.
- [x] With `document.visibilityState` `hidden` the tracker waits for `visibilitychange`, then it sends one time.
  Evidence: same describe → `waits for a hidden document, then sends one time after visibilitychange to visible, and no more after a later hide-show cycle` (extended for TypeScript review MINOR 3), plus the new "the storage write timing" tests.
- [x] With `navigator.webdriver` true the tracker sends nothing.
  Evidence: `the navigator.webdriver filter` → `sends nothing, also for a later click, when navigator.webdriver is true`.
- [x] With `ignoreWebdriver` true and `navigator.webdriver` true the tracker sends the entry.
  Evidence: same describe → `sends the session start entry when ignoreWebdriver is true, also with navigator.webdriver true`.
- [x] A `data-octo` value that starts with `octo:` records no click (rule C38).
  Evidence: pre-existing test `drops a click on a data-octo value that starts with octo: in each letter case`, plus the new test `records a click on a data-octo="octo:session-start" element as no click, also right after the automatic send`.
- [x] `npm test` passes in `kit/tracker`.
  Evidence: 163 tests pass, three full runs, one in reverse file order (`path-match.test.ts`, `index.test.ts`, `batch.test.ts`).
  Evidence: a comment on issue #107, added with the first version of this pull request.

No step of this issue names a "Confirm" source, thus this section has no external lookup to record.

1. **A new session.** A session counts as new when `sessionStorage` holds no id, or when the stored id fails the UUID rule of contract rule C5. A page reload with a valid stored id sends no session start.
2. **The request form.** The session start is its own request, one entry, `ageMs: 0`, the `path` of the page when `routes` gives one. It follows the retry rule of the timer flush (one retry, after a wait), not the lifecycle rule.
3. **`keepalive`.** The request uses `keepalive: true`. The goal of the issue is a visitor who leaves the page after 2 seconds, thus the request must survive a fast page exit. Its body holds 266 bytes at most; the worst case sum with a full queue in flight is `4 * 15000 + 266 = 60266` bytes, below the 64 KB quota.
4. **The C38 guard.** A `data-octo="octo:session-start"` value on a page element still records no click. Only the call inside `start()` sends the session start.
5. **The wait.** `start()` returns at once. The session start goes out when the document is visible and not in a prerender. It goes out one time only. A `stop()` call before that moment cancels it. The removal of the pending listener in `stop()` is the guard that matters here; the run token of issue #36 is a second guard for the same case (correction of TypeScript review MINOR 4, which named the run token alone).
6. **`navigator.webdriver`.** A `true` value blocks the whole tracker, the click listener included, not only the session start. `ignoreWebdriver: true` turns this filter off. Design decision D41 now states this wider scope in one sentence (TypeScript review MINOR 9).
7. **The module-level fallback id.** A blocked `sessionStorage` keeps the id in a module variable, outside one tracker instance, so it survives a `stop()` and a later `start()`. The tests of this state use `vi.resetModules()` and a dynamic `import()`, not a test-only export from `index.ts`: the package `exports` map exposes the whole surface of `index.ts`, thus a reset function there would become part of the public API.
8. **No console leak.** No console line holds a session id or a path; a dedicated test checks `console.log`, `console.warn`, and `console.error`.
9. **The storage write timing (correction round 1, maintainer decision, 2026-09-22).** A new session id lives in a variable from the start. `sendSessionStart()` writes it into `sessionStorage` only at the moment the request goes out, after the wait for a visible, non-prerendering document. A hidden document that ends before that moment stores no id; the next document of the same tab then starts a fresh session, with its own id and its own session start. A click that is queued before the send still carries the right id, from the variable. This closes MAJOR 1 of the reliability review.

## Other changes

- `resolveSessionId()` replaces the direct body of `getOrCreateSessionId()`. It writes to `sessionStorage` only for a new id, not for a reused, valid one (an unneeded write before, now removed).
- The pre-existing test `stop() clears the pending flush timer` compared raw jsdom timer counts before and after `stop()`. `stop()` now always finds a real stored session id to remove (the seeded id of `beforeEach`), thus the raw count no longer isolates the flush timer alone. The test now also checks a direct `vi.getTimerCount()` value (reliability review MINOR 3, TypeScript review MINOR 1).
- `readStoredSessionId()` falls back to the module id when `sessionStorage.getItem()` returns `null`, not only when it throws (reliability review MINOR 1). A store that reads but fails each write then gives one session start, not two.
- `getOrCreateSessionId()` writes a fresh id at once when it creates one, because no wait for a visible document applies to that path (for example, no Web Crypto API yet at `start()` time).
- Merged `origin/main` three times during this work: for pull request #152 (contract v1.2 text) and for the Dependabot bump of `vitest` to `5.0.1` (pull request #133), during the first version of this pull request; and once more during correction round 1, for pull requests #155, #156, #158, #134, #136, and #137. None of these merges touched the tracker source.

## Test count and bundle size

- Tests: 163 passed (0 failed), up from 156 before correction round 1. Three full runs, one in reverse file order, plus a shuffled file and test order.
- `npm run typecheck`: clean.
- `npm run build`: clean (`tsc -p tsconfig.build.json && tsc -p tsconfig.build.types.json`).
- `npm pack --dry-run`: package size 25.9 kB, unpacked 103.2 kB, 17 files.
- `dist/index.js` (the tracker entry file): 14.7 kB raw, 3507 bytes gzip.
- The whole runtime (`dist/index.js` + `dist/batch.js` + `dist/path-match.js`, gzip'd together): 4723 bytes.

Generated with Claude Code


xamcross:
**Review (TypeScript library engineer): APPROVE**

I made my own clone of the repository. I ran `gh pr checkout 162` in it. I ran each
command myself. No BLOCKER finding and no MAJOR finding stays. Nine MINOR findings
follow.

## What I ran

- `npm ci --ignore-scripts`, then `npm test` in `kit/tracker`:
  `Test Files  3 passed (3)` and `Tests  156 passed (156)`.
- A second run, in the reverse file order:
  `npx vitest run src/path-match.test.ts src/index.test.ts src/batch.test.ts --no-file-parallelism`
  gives `Tests  156 passed (156)`.
- A third run, with a shuffled file order and a shuffled test order:
  `npx vitest run --sequence.shuffle.tests=true --sequence.shuffle.files=true --no-file-parallelism`
  gives `Tests  156 passed (156)`. The `vi.resetModules()` block stays isolated, also
  under a random order. The runner reports `jsdom was created 3 times`, thus each test
  file gets its own `document`.
- `npm run typecheck`: clean. `npm run build`: clean.
- `npm pack --dry-run`: 17 files, package size 24.9 kB, unpacked size 100.0 kB. The file
  list holds no test file. `dist/index.d.ts` exports only `createTracker`,
  `TrackerOptions`, and `Tracker`. The change adds no public export.
- The bundle: `dist/index.js` is 14578 bytes raw and 3478 bytes with gzip. The three
  runtime files together are 20072 bytes raw and 4688 bytes with gzip.
- 20 mutations of `kit/tracker/src/index.ts`. 15 of them made the right test fail. 5 of
  them made no test fail. Each of those 5 gives a finding below. I restored the file
  after each mutation. `git status` shows no change to a tracked file now.

## The acceptance criteria of issue #107

|---|---|---|---|
| 1 | One request, one `octo:session-start` entry, before any click | PASS | The test `sends one request with one octo:session-start entry…` holds no click and no timer call. The mutation `ageMs: 0` → `ageMs: 1` makes exactly this test fail. |
| 2 | `ageMs: 0` and the `path` of rule C39 | PASS | The mutation that drops the `click.path` write in `sendSessionStart()` makes `adds the path field of rule C39…` fail. My own probe read the body: `{"sessionId":…,"clicks":[{"element":"octo:session-start","ageMs":0,"path":"/…"}]}`. |
| 3 | No wait for a click and no wait for `flushIntervalMs` | PASS | The same test calls no `clickElement()` and no `vi.advanceTimersByTime()`. The one `fetch` call happens inside `start()`. |
| 4 | A 500 response gives one more attempt; a second failure throws nothing | PASS | The mutation `TIMER_RETRY_COUNT` → `LIFECYCLE_RETRY_COUNT` in `sendSessionStart()` makes both retry tests fail. |
| 5 | A second `start()` with the same stored id sends no session start | PASS | The mutation `if (isNewSession)` → `if (true)` makes 46 tests fail, `sends no session start for a valid stored id` among them. Also the test `sends no session start when a second start() reads the same stored id (a page reload)`. |
| 6 | A blocked `sessionStorage` gives one session start for the document | PASS | The mutation `moduleFallbackSessionId = id;` → `void id;` makes `sends exactly one session start for the life of the document…` fail. |
| 7 | `document.prerendering` true, then `prerenderingchange` | PASS | The mutation `isDocumentPrerendering()` → `false` makes the prerender test fail. |
| 8 | `document.visibilityState` hidden, then `visibilitychange` | PASS | The mutation of the hidden check → `false` makes two tests fail. |
| 9 | `navigator.webdriver` true sends nothing | PASS | The mutation `isBlockedByWebdriver()` → `false` makes `sends nothing, also for a later click…` fail. |
| 10 | `ignoreWebdriver` true sends the entry | PASS | The mutation `options.ignoreWebdriver === true` → `false` makes both `ignoreWebdriver` tests fail. |
| 11 | A `data-octo` value with the prefix `octo:` records no click (C38) | PASS | The mutation `RESERVED_ELEMENT_PREFIX.test(name)` → `false` makes 5 tests fail, the new C38 test among them. |
| 12 | `npm test` passes in `kit/tracker` | PASS | 156 tests pass, in three runs. |
| 13 | A comment on the issue holds the test output | PASS | The comment on issue #107 holds `Tests  156 passed (156)`. |

## The checks of my role

**The body form (rules C13 and C38).** PASS. The session start uses the same body as a
batch: a `sessionId` field and a `clicks` array with one entry. The entry holds
`element`, `ageMs`, and the optional `path`. I read the body of a real call in my own
probe.

**The order.** The request of the session start goes out inside `start()`, before the
page can produce a click. `start()` adds the click listener first, then it calls
`trySendSessionStart()` in the same task. One case breaks the order; see MINOR 6.

**The retry rule.** PASS. The session start uses `TIMER_RETRY_COUNT`, thus one retry
after a wait. The lifecycle rule (no retry) stays for `pagehide` and for
`visibilitychange`.

**`keepalive: true` against the 64 KB budget.** PASS, and I agree with the choice. I
measured the body of the session start: 122 bytes with a path of 121 characters. Rule
C39 limits `path` to 150 bytes, thus the maximum is near 151 bytes. The budget of
`index.ts` is `MAX_QUEUE_SIZE / MAX_BATCH_ENTRIES * MAX_BODY_BYTES`, thus 4 × 15000 =
60000 bytes. A session start in flight at the same time as a full lifecycle flush gives
60151 bytes. That stays below 65536 bytes, with a margin near 5.3 kB. The retry sends
after the first response, thus it adds no second body at the same time. See MINOR 5 for
the comment.

**The C38 guard.** PASS. `handleClick()` keeps the prefix guard with no exception. The
new test clicks a `data-octo="octo:session-start"` element right after the automatic
send, and the tracker records no click.

**No console leak.** PASS. The package writes five console lines: one for the `octo:`
prefix, three for the `routes` option, and one for an oversized click. No line holds a
session id, a path, or a pattern text.

**The `vi.resetModules()` form.** PASS. The blocked-storage block takes a fresh module
instance, thus a fallback id of an earlier test cannot reach it. The suite passes with a
shuffled test order too.

**Design decision D41 and the webdriver filter.** The text of D41 (design line 489) has
`start()` as its subject: "It then runs one time only. It sends nothing when
`navigator.webdriver` is true". The literal text covers the session start only. The
block of the click listener is wider than D41 states. Step 6 of issue #107 says "Send
nothing while `navigator.webdriver` is true", and acceptance criterion 9 says "the
tracker sends nothing". The choice thus agrees with the issue, and it breaks no rule of
the contract. I accept it. The design document needs one sentence; see MINOR 9.

## The findings

### MINOR 1 — the rewritten timer test proves less than the old one
`kit/tracker/src/index.test.ts:278`

The old test compared `vi.getTimerCount()` before and after `stop()`, thus it proved the
`clearTimeout()` call. The new test advances the clock and checks `fetchMock`. I removed
only the line `clearTimeout(timerId);` from `stop()` (line 212 of `index.ts`), and I kept
`timerId = null;`. All 110 tests of the file still pass. The new test cannot see a
leaked timer.

Fix: add one direct check to the same test. I ran this form against the mutation, and it
fails as it must:

```ts
    tracker.stop();
    expect(vi.getTimerCount()).toBe(0);

    vi.advanceTimersByTime(5000);
    expect(fetchMock).not.toHaveBeenCalled();
```

`vi.getTimerCount()` is 0 right after `stop()` with the present jsdom version. I
measured it.

### MINOR 2 — an old test no longer reaches the path that its name states
`kit/tracker/src/index.test.ts:722`

The new `beforeEach` hook stores a valid session id. `resolveSessionId()` then skips
`writeStoredSessionId()`, because a valid stored id needs no rewrite. I measured the
call count of `sessionStorage.setItem` during `start()` with a valid stored id: it is 0.
The test `falls back to an in-memory sessionId when sessionStorage.setItem throws` thus
never enters the `catch` block that it names. The mutation of that `catch` block fails
the new blocked-storage test only, not this one.

Fix: add `window.sessionStorage.clear();` as the first line of the test body, and check
the fallback: assert that the sent `sessionId` matches the UUID pattern, and that
`window.sessionStorage.getItem('octo_session_id')` is `null`.

### MINOR 3 — no test pins the "one time only" rule of the hidden path
`kit/tracker/src/index.ts:324`, `kit/tracker/src/index.test.ts:1837`

The prerender test dispatches a second event and checks the count. The hidden test
dispatches one event only. I removed `{ once: true }` from the `visibilitychange`
listener of `trySendSessionStart()`. All 110 tests still pass, but the tracker then
sends a second session start after a hide-and-show cycle.

Fix: extend the hidden test with a full cycle. I ran this form, and it fails under the
mutation:

```ts
      setVisibilityState('visible');
      document.dispatchEvent(new Event('visibilitychange'));
      setVisibilityState('hidden');
      document.dispatchEvent(new Event('visibilitychange'));
      setVisibilityState('visible');
      document.dispatchEvent(new Event('visibilitychange'));
      expect(fetchMock).toHaveBeenCalledTimes(1);
```

### MINOR 4 — decision 5 of the pull request names the wrong guard
`kit/tracker/src/index.ts:305` and `kit/tracker/src/index.ts:339`

Decision 5 says that `stop()` cancels the pending session start "through the run token of
issue #36". The real cancel comes from the listener removal in `stop()` (lines 226 to
233). I removed the run-token guard of `trySendSessionStart()`: all 110 tests pass. I
removed the run-token guard of `sendSessionStart()`: all 110 tests pass. I removed the
listener removal of `stop()` instead: all 110 tests pass too. The three guards each cover
the same case, thus no black-box test can pin one of them alone.

Fix: keep all three guards, and correct the claim. Add a short comment above the guard in
`trySendSessionStart()`: it is a second guard, and `stop()` already removes the pending
listener.

### MINOR 5 — the keepalive budget comment does not name the session start
`kit/tracker/src/index.ts:67` to `kit/tracker/src/index.ts:75`

The comment gives the budget as `MAX_QUEUE_SIZE / MAX_BATCH_ENTRIES * MAX_BODY_BYTES <
65536`. The session start now adds one more `keepalive: true` request of a maximum near
151 bytes. The sum stays below the limit, but a later reader cannot see this from the
comment.

Fix: add one line to the comment. The session start adds one request. Its maximum is
near 151 bytes: a 36-byte session id, plus a `path` of 150 bytes under rule C39. The sum
is then 60151 bytes.

### MINOR 6 — a click can go out before the session start on a hidden document
`kit/tracker/src/index.ts:318`

`start()` adds the click listener and starts the wait for a visible document. A click
during that wait enters the queue, and the timer flush sends it first. I built a probe:
with `document.visibilityState` hidden at `start()`, the request order is `["save"]`,
then `["save", "octo:session-start"]` after the visible event. Acceptance criterion 1
says "before any click". The case needs a click on a hidden document or on a prerendered
document, thus a real visitor almost never reaches it.

Fix: record the limit in one sentence in the README section "The session start". A
visitor cannot click a hidden document, thus a code change is not needed.

### MINOR 7 — a code comment overclaims
`kit/tracker/src/index.ts:185`

The comment says "A later flush() retries through getOrCreateSessionId()". A later flush
creates the session id, but it sends no session start. That session then holds no
`octo:session-start` entry.

Fix: change the text to: "A later flush() still creates the session id. This document
then sends no session start."

### MINOR 8 — the language of four texts
`kit/tracker/README.md:40`, `:45`, `:51`, and `kit/tracker/src/index.ts:177`

Three sentences of the README pass the 25-word limit of ASD-STE100: line 40 has 31
words, line 45 has 27 words, and line 51 has 34 words. The comment at `index.ts:177`
holds the idiom "tells a new session from an existing one".

Fix: split each long sentence into two. Change the comment to "This early resolve gives
the answer: the session is new, or the session already exists."

### MINOR 9 — record the wider webdriver filter in the design
`docs/superpowers/specs/2026-09-21-octometer-design.md:489`

D41 states the filter for the session start. The code blocks the click listener too. A
later reader of D41 cannot see this.

Fix: add one sentence to D41: "A `true` value blocks the whole tracker. The tracker adds
no click listener."

## Summary

- BLOCKER: 0
- MAJOR: 0
- MINOR: 9

The production code is correct for each acceptance criterion. Each finding is a test, a
comment, or a document. I approve the pull request.

xamcross:
**Review (browser reliability engineer): CHANGES REQUESTED**

I made my own clone. I ran `gh pr checkout 162` in it. I ran `npm ci --ignore-scripts` and each
command myself. I also ran the built tracker in a real Chrome browser, against a local server on
port 7642 that records each request. I stopped the server, I closed the tab, and I deleted the
clone.

## The acceptance criteria

|---|---|---|---|
| 1 | A new session id sends one request with one `octo:session-start` entry before any click | PASS for a visible document | My own run of the test, plus break 5 below. A hidden document breaks the order: see MAJOR 1. |
| 2 | The entry holds `ageMs: 0` and the `path` of rule R3 | PASS | Break 5 (`ageMs` 0 to 1) fails 1 test. The `path` tests assert `/history/:id` and the absent field. |
| 3 | The request waits for no click and for no `flushIntervalMs` | PASS | The test calls no `clickElement()` and no `vi.advanceTimersByTime()`. Chrome: the request left inside `start()`. |
| 4 | A 500 response makes one more attempt; a second failure throws no error | PASS | Break 6 (the lifecycle retry count on the session start) fails 2 tests. |
| 5 | A second `start()` with the same stored id sends no session start | PASS | Break 2 (`isNew` always true) fails more than 20 tests. |
| 6 | With a blocked `sessionStorage`, one document sends exactly one session start | PASS for a full block, FAIL for a partial block | My own probe: a store that reads but fails each write gives 2 session starts with 2 ids. See MINOR 1. |
| 7 | `document.prerendering` true sends nothing until `prerenderingchange`, then one time | PASS in jsdom | Break 1 fails 3 tests. My own probe adds the case where `prerenderingchange` arrives before the visible state; the code then waits again and sends one time. Chrome: not proven, see "What stays without a measurement". |
| 8 | A hidden document waits for `visibilitychange`, then sends one time | PASS in jsdom | Break 1 fails 3 tests. My own probe: a hidden tab for one hour sends nothing, and a `stop()` cancels it. |
| 9 | `navigator.webdriver` true sends nothing | PASS | Break 4 fails 1 test. |
| 10 | `ignoreWebdriver` true sends the entry | PASS | The test asserts the entry and a later click. |
| 11 | A `data-octo` value with the prefix `octo:` records no click | PASS | The pre-existing test plus the new test for the exact text. |
| 12 | `npm test` passes in `kit/tracker` | PASS | My own run: `Test Files 3 passed (3)`, `Tests 156 passed (156)`. `npm run typecheck` and `npm run build` are clean. |
| 13 | A comment on the issue records the test output | PASS | The comment is on issue #107. |

The diff holds 3 files: `kit/tracker/README.md`, `kit/tracker/src/index.ts`, and
`kit/tracker/src/index.test.ts`. It holds no secret, no database file, and no build output. Each
name, each number, and each rule agrees with decision D24, with decision D41, and with contract
rules C38 and C39. The language of the README, of each comment, and of the commit body follows
ASD-STE100.

---

## BLOCKER

None.

---

## MAJOR 1 — a hidden document stores the session id before the session start goes out

**File:** `kit/tracker/src/index.ts:180-204` and `kit/tracker/src/index.ts:563-576`

**The defect.** `start()` calls `resolveSessionId()` first. That function writes the new id into
`sessionStorage` at line 573. Only after that does `trySendSessionStart()` decide to wait. Two
results follow for a document that starts hidden.

1. **The click queue does not wait.** The wait gates the session start alone. A click that enters
   the queue during the wait starts the normal timer, and its request goes out first.
2. **The next document sends no session start at all.** The id is already in `sessionStorage`.
   A navigation, or a tab close, before the visible moment loses the session start for ever. The
   next `start()` in the same tab reads a valid stored id, counts the session as not new, and
   sends nothing. The clicks of that session still go out.

**My measurement in a real Chrome.** The extension window stayed occluded, thus each page of my
run loaded with `document.visibilityState === 'hidden'`. The page called `start()`, then it
clicked one `data-octo` element.

- The page state: `load vis=hidden pre=false wd=false | start() vis=hidden pre=false | clicked`.
- The server received exactly one request, at 5416 ms:
  `{"sessionId":"d81a9984-...","clicks":[{"element":"save","ageMs":5416,"path":"/p/:id"}]}`.
- The session start never arrived. The id `d81a9984-...` stayed in `sessionStorage`
  (`sessionStorage.length === 1`).
- I then navigated the tab away. That session holds a click row and no start row, for ever.

**My measurement in jsdom.** A hidden document, `start()`, one click, then 5000 ms: the request
order is `["save", "octo:session-start"]`. The session start arrives second, after the switch to
the visible state. A `pagehide` of a page that stays hidden sends the click batch and drops the
session start.

**Why this matters.** The issue text, the README, and decision D41 all promise a session start
"before any click". The promise holds only for a visible document. Decision D44 reads the start
row of each session. A session with no start row shows the path `(unknown)`, and a start row that
arrives after the first click carries a `ts` later than that click (rule C14:
`ts = receivedAt - ageMs`, and `ageMs` is 0 here). The "First pages" view and the start time of
the "Anonymous sessions" view both take the wrong value.

**The exact fix.** Choose one, and state the choice in the README and in decision D41.

- (a) Keep a second `sessionStorage` key, for example `octo_session_start_pending`. `start()`
  writes it beside the id, and `sendSessionStart()` removes it after the `fetch` call. A later
  `start()` that finds a valid id **and** this key keeps the same id and sends the session start
  again. One session then keeps one id, and it gets exactly one start row.
- (b) Add these three sentences to the README section "The session start", and to decision D41:
  "The wait holds back the session start alone. A click of a hidden document goes out first. A
  document that ends while it is hidden sends no session start, and the next document of the same
  tab sends none either, because the session id is already stored." Then name decision D44, which
  shows such a session with the path `(unknown)`.

I prefer (a). The maintainer must decide.

---

## The answers to the review questions

**Question 1, the prerender wait and the hidden tab.** Each rule holds in jsdom. My own probes,
beside the tests of the pull request:

| Case | Result |
|---|---|
| A prerender that never activates | 0 requests after 600 000 ms. A `stop()` during the prerender cancels it. |
| `prerenderingchange` while the document is still hidden | 0 requests. The code waits again for `visibilitychange`, then it sends one time. Two later visibility changes send nothing more. |
| A hidden tab for one hour | 0 requests after 3 600 000 ms. A `stop()` cancels it, and a later switch to the visible state sends nothing. |
| A hidden tab, then a switch to it | 1 request, with the one `octo:session-start` entry. |

**Question 2, the double count.** `sendSessionStart()` at `index.ts:352` passes
`TIMER_RETRY_COUNT`, thus a first request that failed, and that still reached the store, gives a
second `octo:session-start` row for the same session. Rule C24 drops a copy only for the same
`_id`, and a retry writes a new `_id`. The two rows also carry a different `ts`, because
`receivedAt` moves and `ageMs` stays 0. No later query can find the pair.

What the monitor does with two start rows of one session:

- **Level 1, the session count.** No change. The design counts sessions with
  `COUNT(DISTINCT session_id)` and with no `kind` filter. Two rows of one session give one session.
- **Level 1, the click count.** No change. That statement filters `kind = 0`, and a start row is
  `kind = 1`.
- **"First pages" and "Anonymous sessions" (decision D44).** The design does not state the
  aggregation of either statement. "First pages" is safe only when it counts
  `COUNT(DISTINCT session_id)`. "Anonymous sessions" is safe only when it takes one row for each
  session, for example with `MIN(ts)`. The issue that builds decision D44 must hold both rules.
  Please add that sentence to decision D44 now, before the work starts.

**Is the second send worth it for a session start?** Yes, with one qualification. The batch holds
one entry, thus the mid-batch `insertMany` failure of issue #36 cannot apply: an ordered
`insertMany` of one document either writes it or it does not. A 500 answer for this request
means "not written", thus the retry is right and it writes no copy. Only a dropped connection
after a commit gives a copy, and the cost of that copy is one extra `kind = 1` row, which does
not change the session count. The value is the goal of the issue: the first page of a visitor who
leaves after 2 seconds. Keep the retry.

**Question 3, the order.** A visible document sends the session start first: `start()` calls
`fetch` inside its own call frame, at `index.ts:352`, and a click at 100 ms waits for the timer of
`flushIntervalMs`. I ran the test, and I saw the one call with no `advanceTimersByTime`. A hidden
document breaks the order, because the queue does not wait. See MAJOR 1. The design must say which
of the two rules it wants: the queue waits too, or a click can go first.

**Question 4, the `keepalive` budget.** The numbers:

- One session start body, with no `path` field: **106 bytes**.
- One session start body, with a `path` of the full 150 bytes of rule C39: **266 bytes**.
- The worst case of the code caps: `MAX_QUEUE_SIZE / MAX_BATCH_ENTRIES` gives 4 requests, thus
  `4 * MAX_BODY_BYTES + 266` = **60 266 bytes**. The quota is 65 536 bytes. The margin is
  **5 270 bytes**.
- My measurement of a real full queue of 200 entries at the contract maximums, plus the session
  start: `266 + 14463 + 14463 + 14463 + 14463` = **58 118 bytes**.
- **My Chrome proof.** I sent 4 bodies of 14 512 bytes and 1 body of 266 bytes with
  `keepalive: true`, then I navigated the tab away. All 5 requests arrived. The sum was
  58 314 bytes.
- **My Chrome control.** 5 bodies of 14 512 bytes, a sum of 72 560 bytes: one request failed with
  `TypeError: Failed to fetch`. The 64 KB quota of the page is real, and it counts the sum.

The criterion holds. The margin drops by 266 bytes for each session start still in flight at the
page end. See MINOR 4: the README does not name the session start as a `keepalive` request.

**Question 5, the session id and a duplicated tab.** Chrome copies `sessionStorage` when the user
duplicates a tab. My probe: a second tracker that reads the same store finds the same valid id,
counts the session as not new, and sends no session start. The two tabs share one session id.
**This is acceptable under decision D41.** D41 gives one rule only: a new id sends the entry. A
duplicated tab holds no new id. The session count stays right, because the design counts
`COUNT(DISTINCT session_id)`. The clicks of the two tabs join one session, and the design has no
rule that separates them. Section 7 of the design also puts a new session id after 30 idle minutes
out of scope for version 1, thus the design accepts a coarse session.

**Question 6, the `navigator.webdriver` filter.** `isBlockedByWebdriver()` at `index.ts:294`
compares with `=== true`. An old browser gives `undefined`, and `undefined !== true`, thus the
tracker runs normally. I proved it: I deleted the property and `start()` sent the entry. A headless
Chrome sets the property to `true`, thus the pilot app's own tests create no session. I also
checked the state after a blocked `start()`: `started` stays false, `everStarted` stays false, and
a later `stop()` removes no stored id. That is correct. In my own Chrome the extension leaves
`navigator.webdriver` at `false`, thus the extension itself does not turn the tracker off.

**Question 7, `npm ci` and the break test.** `npm ci --ignore-scripts` and `npm test` pass: 3
files, 156 tests. I broke the code 11 times, and I restored the file after each one:

|---|---|---|
| 1 | The prerender check and the hidden check both removed | 3 tests fail |
| 2 | `resolveSessionId()` returns `isNew: true` always | more than 20 tests fail |
| 3 | **The run token check at the top of `trySendSessionStart()`** | **0 tests fail** — MINOR 2 |
| 4 | `isBlockedByWebdriver()` returns false | 1 test fails |
| 5 | `ageMs: 0` becomes `ageMs: 1` | 1 test fails |
| 6 | The session start takes `LIFECYCLE_RETRY_COUNT` | 2 tests fail |
| 7 | The session start takes `keepalive: false` | 1 test fails |
| 8 | **The run token check inside `sendSessionStart()`** | **0 tests fail** — MINOR 2 |
| 9 | **`stop()` keeps the pending visibility listener** | **0 tests fail** — MINOR 2 |
| 10 | Break 9 plus break 3 plus break 8 | 0 tests fail. The fourth guard, `sessionId === null`, still holds. All four guards removed: 1 test fails. |
| 11 | **`stop()` keeps the pending flush timer** | The test of its own name passes. A different test fails — MINOR 3 |

No test is empty. No test is always green.

**What stays without a measurement.** The switch of a real Chrome tab from hidden to visible, and
a real prerender activation. The extension window stayed occluded for my whole run, thus each page
reported `hidden` for its whole life, and no `visibilitychange` event fired. My speculation-rules
page also reported `document.prerendering === false` after the navigation, thus Chrome ran no
prerender for it. Each rule of question 1 rests on jsdom alone. Firefox, Safari, and a freeze on a
mobile browser also stay without a measurement.

---

## MINOR findings

**MINOR 1 — a store that reads but fails each write gives two session starts.**
`kit/tracker/src/index.ts:578-596`. `readStoredSessionId()` reads the module fallback id only when
`getItem` **throws**. A full `sessionStorage` quota, and an old Safari private window, let
`getItem` return `null` and make `setItem` throw. The module fallback id then holds a value that
nothing reads back. My probe, one document, a `start()` and a `stop()` and a `start()`: **2**
session starts, with **2 different** session ids. The acceptance criterion holds for a full block
only. Fix, one line in `readStoredSessionId()`:
`const stored = window.sessionStorage.getItem(SESSION_STORAGE_KEY); return stored ?? moduleFallbackSessionId;`
The module variable stays `null` while each write succeeds, thus this change alters no normal path.
Add a test with a real `getItem` and a throwing `setItem`.

**MINOR 2 — no test holds the run token of the session start, and none holds the listener removal
of `stop()`.** `kit/tracker/src/index.ts:230-233`, `:306-308`, and `:339`. The cancel path has four
guards. I removed each one alone, and also three of the four together: 0 tests fail each time. Only
the fourth guard, `sessionId === null`, keeps the test green. Fix: add one test that calls
`stop()`, then `start()` again with a hidden document, then fires `visibilitychange`. Assert
exactly one request. Add one more test for a `stop()` during the prerender wait; today only the
visibility wait has that test.

**MINOR 3 — the renamed test does not check its own name.**
`kit/tracker/src/index.test.ts:286` (`stop() clears the pending flush timer, so it never fires`).
I removed the `clearTimeout` call of `stop()` at `index.ts:211-214`. This test still passed,
because `stop()` empties the queue first, thus the stale timer finds nothing to send. A different
test caught the break. Fix: after `stop()`, call `start()` again, click one element, advance
`flushIntervalMs`, and assert exactly one click batch. A stale timer blocks `scheduleFlush()`, thus
that assertion fails when `stop()` keeps the timer.

**MINOR 4 — the README does not name the session start as a `keepalive` request of the budget.**
`kit/tracker/README.md`, the sections "The session start" and "The page lifecycle flush". The
lifecycle section states the 64 KB rule and the figure of about 58 KB for a full queue. The new
session start also uses `keepalive: true`, and it shares the same budget of the page. Fix: add one
sentence to "The page lifecycle flush": "The session start of the next section also uses
`keepalive: true`. Its body holds 266 bytes at most, thus the sum with a full queue stays near
58 KB."

**MINOR 5 — a code comment promises a retry that never happens.**
`kit/tracker/src/index.ts:185-186`: "A later flush retries through `getOrCreateSessionId()`". That
call retries the **session id** only. It never sends the session start. A page with no Web Crypto
API at `start()` time thus sends no session start at all, for the life of the document. Fix:
change the comment to "A later flush still finds an id through `getOrCreateSessionId()`. The
session start of this document does not go out."

**MINOR 6 — the README joins two statements with a semicolon, and it uses a phrasal verb.**
`kit/tracker/README.md`, the options table and the section "The session start": "Set the option
`ignoreWebdriver` to `true` only for an end-to-end test of the app itself; it turns this filter
off." and "this client-side filter only keeps test traffic out of normal use". Rule 4 of
ASD-STE100 asks for one instruction in one sentence, and rule 11 asks for no idiom. Fix: "Set the
option `ignoreWebdriver` to `true` only for an end-to-end test of the app itself. The option turns
the filter off." and "this client-side filter removes the traffic of a test tool only".

---

## Scope and the design

The change does what the issue asks, and no more. The new option `ignoreWebdriver` matches the
option list of decision D24. The entry form matches section 4.2 of the design and contract rule
C38. The `path` field matches rule C39 and rule C42. The decision to give the session start the
retry rule of the timer flush, and not of the page lifecycle flush, is right: the timer flush runs
while the page is visible, and so does the session start after its wait.

I approve after MAJOR 1 closes.

xamcross:
**Correction round 1.** I made my own clone, at `C:/Users/xamcr/octo-wt/issue-107b`. I ran `gh pr checkout 162` in it. I ran each command myself.

## MAJOR 1 (reliability review)

The maintainer decision: `start()` keeps a new session id in a variable. `sendSessionStart()` writes the id into `sessionStorage` only at the moment the request goes out, after the wait for a visible, non-prerendering document. A hidden document that ends before that moment stores no id, so the next document of the same tab starts a fresh session, with its own id and its own session start.

File and line: `kit/tracker/src/index.ts:354` (the new `writeStoredSessionId(sessionId)` call in `sendSessionStart()`), and `kit/tracker/src/index.ts:589-598` (`resolveSessionId()` no longer writes for a new id). The README section "The session start" and design decision D41 (`docs/superpowers/specs/2026-09-21-octometer-design.md:485-497`) both state this rule now, in one paragraph and one sentence.

New tests, `kit/tracker/src/index.test.ts`, describe `the storage write timing (MAJOR 1)`:
- `leaves no stored id when a hidden document ends before it becomes visible (a pagehide while hidden)`
- `stores the id at the moment the session start goes out, when the document becomes visible`
- `sends no second session start on a reload after the id enters storage at the send, for a document that started hidden`

I broke the fix (restored the old, eager write inside `resolveSessionId()`) and ran the first two tests: both failed, with the real id in storage where the test expects `null`. I restored the file, and confirmed the full suite is green again.

## Reliability review MINOR findings

| # | Finding | Result | File and line |
|---|---|---|---|
| 1 | A store that reads but fails each write gave two session starts, with two ids | Fixed | `kit/tracker/src/index.ts:600-613` (`readStoredSessionId()` now falls back to the module id on a `null` read, not only on a throw). New test: `gives one session start with one id, not two, when sessionStorage reads succeed but each write fails`. Broke the fix (removed the `?? moduleFallbackSessionId` fallback): the test failed with 2 calls and 2 ids. Restored. |
| 2 | No test held the run token and the listener removal of `stop()` together | Fixed | Two new tests: `sends exactly one session start when start() runs again after stop(), for a document that stays hidden through the first wait`, and the matching prerender test. Plus a new `stop()` during the prerender wait cancel test. Broke all four cancel guards together (both run-token checks, both listener-removal blocks in `stop()`): both new tests failed, 2 calls instead of 1. Restored. |
| 3 | The renamed test did not check its own name (the `clearTimeout()` call of `stop()`) | Fixed | `kit/tracker/src/index.test.ts:278-297`, added a direct `expect(vi.getTimerCount()).toBe(0)` check. Broke the fix (removed `clearTimeout(timerId)` from `stop()`): the test failed, 1 timer instead of 0. Restored. Same defect as TypeScript review MINOR 1. |
| 4 | The README did not name the session start in the 64 KB budget | Fixed | `kit/tracker/README.md`, "The page lifecycle flush": added the exact figures, `4 * 15000 + 266 = 60266` bytes. |
| 5 | A code comment overclaimed "A later flush() retries" | Fixed | `kit/tracker/src/index.ts:186-188`, corrected to state that a later flush only creates the session id; it sends no session start for that document. Same defect as TypeScript review MINOR 7. |
| 6 | The README joined two statements with a semicolon, and used a phrasal verb | Fixed | `kit/tracker/README.md`, "The session start": split the sentences, and changed "this client-side filter only keeps test traffic out of normal use" to "this client-side filter removes the traffic of a test tool only". |

## TypeScript review MINOR findings

| # | Finding | Result | File and line |
|---|---|---|---|
| 1 | The rewritten timer test proved less than the old one | Fixed | Same fix as reliability MINOR 3 above. |
| 2 | An old test no longer reached the path it named (the catch block of the in-memory fallback) | Fixed | `kit/tracker/src/index.test.ts:727-761`, added `window.sessionStorage.clear()` and a fresh module instance (`vi.resetModules()` + dynamic `import()`), so `setItem` actually runs and throws. Broke the fix (removed the `try`/`catch` of `writeStoredSessionId()`): the test failed with an uncaught `DOMException`. Restored. I also had to isolate this test with a fresh module import, because the write pollutes the shared module-level fallback id for every later test in the file; I found this only after the first version of the fix made 21 unrelated tests fail. |
| 3 | No test pinned the "one time only" rule of the hidden path | Fixed | Extended `waits for a hidden document...` with a hide-show-hide-show cycle. Broke the fix (removed `{ once: true }` from the `visibilitychange` listener of the wait): the test failed, 3 calls instead of 1. Restored. |
| 4 | Decision 5 of the pull request named the wrong guard | Fixed | Pull request text, item 5: now names the listener removal in `stop()` as the guard that matters, and the run token as a second guard for the same case. |
| 5 | The keepalive budget comment did not name the session start | Fixed | `kit/tracker/src/index.ts:67-77`, added the session start figures to the existing comment. |
| 6 | A click can go out before the session start on a hidden document | Fixed | `kit/tracker/README.md`, "The session start": states the limit in one paragraph. No code change; this is inherent to the maintainer decision on MAJOR 1 too, since the wait holds back the session start alone. |
| 7 | A code comment overclaimed | Fixed | Same fix as reliability MINOR 5 above. |
| 8 | The language of four texts passed the word limit, or used an idiom | Fixed | Three README sentences split, and the `index.ts:181-182` comment changed from "tells a new session from an existing one" to "gives the answer: the session is new, or the session already exists". |
| 9 | The design document did not record the wider scope of the `navigator.webdriver` filter | Fixed | `docs/superpowers/specs/2026-09-21-octometer-design.md:485-497` (D41), one added sentence: "A `true` value of `navigator.webdriver` blocks the whole tracker, the click listener included, not only the session start." |

Also, question 2 of the reliability review (the double-count aggregation rule): added one sentence to design decision D44, that "First pages" counts `COUNT(DISTINCT session_id)`, and "Anonymous sessions" takes one row for each session. Changed only that decision.

## Corrected count

- Reliability review: 1 MAJOR fixed, 6 of 6 MINOR fixed.
- TypeScript review: 9 of 9 MINOR fixed.

## Verification

- `npm ci` in `kit/tracker`: clean (`EBADENGINE` warnings only, for jsdom's newer Node requirement; the installed Node is v24.13.0).
- `npm test`: 163 tests pass, 0 failed, up from 156. Three full runs: the normal file order, the reverse file order (`path-match.test.ts`, `index.test.ts`, `batch.test.ts`), and a shuffled file and test order.
- `npm run typecheck`: clean.
- `npm run build`: clean.
- `npm pack --dry-run`: package size 25.9 kB, unpacked 103.2 kB, 17 files.
- Bundle: `dist/index.js` 14.7 kB raw, 3507 bytes gzip. The whole runtime, gzip'd together: 4723 bytes.
- Merged `origin/main` into the branch with a merge commit (pull requests #155, #156, #158, #134, #136, #137). No conflict. None of these touched the tracker source; `kit/tracker/package.json` picked up the TypeScript bump to `~7.0.2` (pull request #134's family), and the full suite still passes under it.
- Pushed. CI: all 5 checks pass (Tracker, Frontend, JVM modules, Monitor backend on Windows, Secret scan).

New commits on `issue-107-tracker-session-start`:
- `fc1d592` fix(tracker): defer the session id write to the session start send
- `30a08b8` test(tracker): pin the storage write timing and the cancel guards
- `7224505` docs(tracker): record the storage write timing and the budget
- plus one merge commit for `origin/main`

Generated with Claude Code

xamcross:
**Review (second review, browser reliability engineer): APPROVE**

I made a new clone. I ran `gh pr checkout 162` and `npm ci --ignore-scripts` in it. I ran each
check myself, in jsdom and in a real Chrome. I stopped my server, I closed my tab, and I deleted
the clone.

## MAJOR 1 — closed

The maintainer rule holds. `sendSessionStart()` writes the id at `index.ts:354`, and
`resolveSessionId()` writes none. My own probes, each one green:

| # | My probe | Result |
|---|---|---|
| a | A hidden document, one click, a timer flush, then `pagehide` | 1 click request goes out. `sessionStorage` stays empty. |
| b | A hidden document, then the visible state | 1 session start. The stored id equals the `sessionId` of that request. |
| c | A reload after the send, with the store kept | 0 requests. The id does not change. |
| d | A prerender that never activates, then a fresh visible document | The prerender stores nothing. The next document makes its own id and its own session start. |
| e | A click flush during the wait | The click and the later session start carry **one** id. The store stays empty until the send. |

**My Chrome proof.** The extension window stays occluded, thus each page loads with
`document.visibilityState === 'hidden'`. This is the exact case of my first review. The page state
is now `load vis=hidden wd=false | start() vis=hidden stored=0 | clicked`, and
`sessionStorage.length` is **0** after 7 seconds. The server got one request only:
`{"sessionId":"d8e4060c-...","clicks":[{"element":"save","ageMs":5995,"path":"/p/:id"}]}`. In my
first review the same page left the id in the store. It leaves none now.

**The break test.** I moved the write back into `resolveSessionId()`, the old, eager form. Two
tests failed: `leaves no stored id when a hidden document ends before it becomes visible` and
`stores the id at the moment the session start goes out`. I restored the file.

The new paragraph of the README, "The storage write timing", and the new sentence of decision D41
(`docs/superpowers/specs/2026-09-21-octometer-design.md:489-492`) both state the rule correctly.

## My 6 MINOR findings

|---|---|---|---|
| 1 | A store that reads `null` but fails each write gave 2 session starts | Fixed | `index.ts:602-608` returns `stored ?? moduleFallbackSessionId`. My probe now gives 1 session start. I removed the fallback: the new test fails. |
| 2 | No test held the run token and the listener removal of `stop()` | Fixed | My probe: `stop()` then `start()` on a document that stays hidden gives exactly 1 session start. A `stop()` during the prerender wait cancels it. Both cases now have a test of the pull request too. |
| 3 | The renamed timer test did not check its own name | Fixed | I removed the `clearTimeout` call of `stop()`. The test `stop() clears the pending flush timer, so it never fires` now **fails** (2 tests fail in total). It passed before. |
| 4 | The README did not name the session start in the 64 KB budget | Fixed | `README.md`, "The page lifecycle flush": it states 266 bytes and `4 * 15000 + 266 = 60266`. The comment at `index.ts:75-79` states the same. |
| 5 | A code comment promised a retry that never happens | Fixed | `index.ts:186-189` now says that a later flush creates the id only, and that the document sends no session start. |
| 6 | Two ASD-STE100 defects in the README | Fixed | The semicolon sentence is now two sentences. "keeps test traffic out of normal use" is now "removes the traffic of a test tool only". |

## Question 2, the aggregation rule of decision D44

The new sentence at `docs/superpowers/specs/2026-09-21-octometer-design.md:538-540` is correct and
complete: "First pages" counts `COUNT(DISTINCT session_id)`, "Anonymous sessions" takes one row for
each session, and a second start row from a retry changes no count of either view. This closes the
open point of my first review.

## The test run

My own run: `Test Files 3 passed (3)`, `Tests 163 passed (163)`. `npm run build` is clean. Three
breaks, each one restored: the eager storage write (2 tests fail), the `clearTimeout` of `stop()`
(2 tests fail), and the `null`-read fallback (1 test fails). No new test is empty. No new test is
always green.

**What stays without a measurement.** A real Chrome switch from hidden to visible, and a real
prerender activation. The extension window stays occluded, thus each page reports `hidden` for its
whole life. Each rule of the wait rests on jsdom alone.

xamcross:
**Review (second review, TypeScript library engineer): APPROVE**

I made a new clone. I ran `gh pr checkout 162` in it, at `7224505`. I ran each command
myself. Each of my 9 MINOR findings of round 1 is fixed. The change for MAJOR 1 of the
reliability review is correct, and three new tests pin it.

## My 9 MINOR findings of round 1

| # | Finding | Status | My evidence |
|---|---|---|---|
| 1 | The rewritten timer test proved less than the old one | Fixed | `index.test.ts:290` holds `expect(vi.getTimerCount()).toBe(0)`. I removed only `clearTimeout(timerId);` from `stop()`: the test `stop() clears the pending flush timer, so it never fires` fails. In round 1 it passed. |
| 2 | An old test did not reach the `catch` block that it names | Fixed | `index.test.ts:727` now calls `vi.resetModules()`, takes a fresh module, and calls `window.sessionStorage.clear()`. I turned the `catch` of `writeStoredSessionId()` into an `else` branch: the test fails. In round 1 the same mutation left it green. |
| 3 | No test pinned the "one time only" rule of the hidden path | Fixed | I removed `{ once: true }` from the `visibilitychange` listener of the wait: the renamed test `…and no more after a later hide-show cycle` fails. In round 1 it passed. |
| 4 | Decision 5 named the wrong guard | Fixed | Item 5 of the pull request text now names the listener removal of `stop()`, and the run token as a second guard. I removed all four cancel guards together: the two new restart tests fail. Each single guard still escapes, because the four cover the same case. |
| 5 | The keepalive budget comment did not name the session start | Fixed | `index.ts:72-79` and the README section "The page lifecycle flush". **I correct my own number of round 1.** I wrote 151 bytes. That figure is wrong: my probe used a `*` segment above 80 characters, thus the matcher gave the short path `/other`. The 266 bytes of the author is right, and `4 * 15000 + 266 = 60266` stays below 65536. |
| 6 | A click can go out before the session start on a hidden document | Fixed | `README.md`, "The session start": one paragraph states the limit. The fix needs no code change. |
| 7 | A code comment overclaimed | Fixed | `index.ts:187-189` now states that a later flush only creates the id, and that the document sends no session start. |
| 8 | The language of four texts | Fixed | The three long README sentences are split. The comment at `index.ts:181-182` now reads "gives the answer: the session is new, or the session already exists". I read each new sentence: each one stays below 25 words. |
| 9 | The design did not record the wider webdriver filter | Fixed | D41, `2026-09-21-octometer-design.md:490-491`: "A `true` value of `navigator.webdriver` blocks the whole tracker, the click listener included, not only the session start." |

## MAJOR 1 of the reliability review (the storage write time)

The code is correct. `sendSessionStart()` writes the id at `index.ts:354`, after the run
token check and the null check, and before the `fetch` call. `resolveSessionId()`
(`index.ts:589-598`) writes no id now. `getOrCreateSessionId()` (`index.ts:560-571`)
writes only on the path where `start()` found no id, for example with no Web Crypto API.
That path starts no wait, thus the write is correct there.

I read the three new tests of `the storage write timing (MAJOR 1)`. They are real tests.
My own mutations:

| Mutation | Result |
|---|---|
| Remove `writeStoredSessionId(sessionId);` from `sendSessionStart()` | 7 tests fail, the three new tests among them. |
| Put the old eager write back into `resolveSessionId()` | Exactly the two new time tests fail: `leaves no stored id when a hidden document ends…` and `stores the id at the moment the session start goes out…`. |
| `return stored ?? moduleFallbackSessionId;` → `return stored;` | The new test `gives one session start with one id, not two, when sessionStorage reads succeed but each write fails` fails. |
| Turn the `catch` of `writeStoredSessionId()` into an `else` branch | 4 tests fail. The internal `catch` thus holds, and the write at `index.ts:354` cannot throw into the page, although it sits outside the `try` of `sendSessionStart()`. |

## The 5 mutations that escaped in round 1

| Mutation | Round 1 | Now |
|---|---|---|
| `stop()` drops only the `clearTimeout()` call | No test fails | 1 test fails |
| The `visibilitychange` wait listener loses `{ once: true }` | No test fails | 1 test fails |
| `trySendSessionStart()` loses its run-token guard | No test fails | No test fails alone |
| `sendSessionStart()` loses its run-token guard | No test fails | No test fails alone |
| `stop()` keeps the pending `visibilitychange` listener | No test fails | No test fails alone |

The last three guards cover one case each other also covers. I removed all four together:
the two new restart tests fail. That is the right level of cover for a redundant guard. I
ask for no further change.

## The commands that I ran

- `npm ci --ignore-scripts` in `kit/tracker`: clean.
- `npm test`: `Test Files  3 passed (3)`, `Tests  163 passed (163)`.
- The reverse file order (`path-match.test.ts`, `index.test.ts`, `batch.test.ts`): 163 passed.
- A shuffled file order and a shuffled test order: 163 passed. The tests stay isolated,
  also with the two `vi.resetModules()` blocks in one file.
- `npm run typecheck`: clean. `npm run build`: clean. `npx tsc --version` gives 7.0.2,
  thus the merge of `origin/main` raised TypeScript, and the package still builds.
- `npm pack --dry-run`: 17 files, package size 25.9 kB, unpacked size 103.2 kB. The list
  holds no test file. `dist/index.d.ts` still exports only `createTracker`,
  `TrackerOptions`, and `Tracker`.
- The bundle: `dist/index.js` 14725 bytes raw, 3507 bytes with gzip. The three runtime
  files together: 20219 bytes raw, 4723 bytes with gzip. Each figure agrees with the
  report of the author.
- `gh pr checks 162`: all 5 checks pass.

## The result

- BLOCKER: 0
- MAJOR: 0
- MINOR: 0

I approve the pull request.

Issue #117, opened 2026-09-21
Add the daily anonymous cap and the user-agent filter


xamcross:
**Goal.** A daily cap limits the anonymous events in total and for each key. The ingest route drops a batch of a known robot.

**Why it blocks production.** Without a daily cap one address fills the cap of 200 000 events of D21, thus the monitor shows no new number for 30 days. A robot that runs JavaScript makes a session start for a page that no person opened, thus each first-page number is wrong.

**Implementation steps.**
1. In `kit/jvm-core`, add the global cap `OCTOMETER_MAX_ANON_EVENTS_PER_DAY` with the default 20 000.
2. Add the cap for each key `OCTOMETER_ANON_EVENTS_PER_KEY_PER_DAY` with the default 2 000, in a map of 20 000 keys with an LRU eviction.
3. Above a cap, drop the batch and write one warning each hour. Reset each counter after 24 hours.
4. In `kit/jvm-ktor`, give the `User-Agent` header to the ingest logic, and map a dropped batch to the response 204.
5. Drop a batch when the user agent matches `bot|crawl|spider|slurp|headless|preview|monitor|Go-http-client|python-requests|curl`. The match ignores the letter case, and the kit stores no user agent.
6. Add the tests of the acceptance criteria with an injected clock.

**Acceptance criteria.**
- [x] Above `OCTOMETER_MAX_ANON_EVENTS_PER_DAY` the kit drops the batch, and the route returns 204.
- [x] Above `OCTOMETER_ANON_EVENTS_PER_KEY_PER_DAY` the kit drops the batch of that key, and a second key still passes.
- [x] Fixed clock: the kit writes a maximum of one warning each hour.
- [x] Fixed clock: each counter resets after 24 hours.
- [x] The key map holds 20 000 keys with an LRU eviction.
- [x] The user agent `Mozilla/5.0 (compatible; Googlebot/2.1)` gives 204, and the store gets 0 events.
- [x] The user agent `PYTHON-REQUESTS/2.31` gives 204, because the match ignores the letter case.
- [x] A drop of a robot batch writes one debug line.
- [x] No stored event and no log line holds the user agent.
- [x] `./gradlew :kit:jvm-core:test` and `./gradlew :kit:jvm-ktor:test` pass.
- [x] A comment on this issue records the test output.

**Related.**
- #34 gives the event cap and the 204 response.
- #116 adds the per-minute counters.
[...]

Issue #70, opened 2026-09-21
Write the integration guide for a Spring Boot app


xamcross:
**Blocked by #40 and #71.** The guide links the Ktor guide of #40 and gives the coordinates that #71 records. The outline can start now.

**Goal.** One guide tells a reader how to add Octometer to a Spring Boot app.

**Evidence.** Observed on 2026-09-21: `cadence` sends only `/api`, `/oauth2`, and `/login/oauth2/code` from Cloudflare Pages to the backend. It enforces a cookie CSRF token.

**Implementation steps.**
1. Create `docs/integration-spring.md`. It holds only the differences and links `docs/integration-ktor.md` for each shared part.
2. Write the section on the store: the app gives `MongoDatabaseFactory.getMongoDatabase()` to the MongoDB event store.
3. Write the section on the `UserIdResolver`: it reads the user id from the security context.
4. Write the section on the tracker. The `headers` callback reads the cookie `XSRF-TOKEN` and sends `X-XSRF-TOKEN`, and `credentials: 'include'` applies when the API has another origin.
5. Write the section on the route with the Cloudflare Pages `/api` path. The section states that a chain with `oauth2ResourceServer` is not necessary for this route.
6. Add one step that sends a POST through the public origin and reads the event back.

**Acceptance criteria.**
- [ ] The guide links `docs/integration-ktor.md` for the shared parts.
- [ ] The guide holds each part of steps 2 to 6.
- [ ] A comment on this issue records the commit of the guide.

**Related.** #119 adds the pilot route list, `OCTOMETER_RECORD_ANONYMOUS=true`, and the `start()` rule for server-side rendering to the same file.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, section 2.3, D22, D23, D24.

xamcross:
Status on 2026-09-21 from the maintainer: the owner selected `investguideua` as the pilot app (Spring Boot 3.4.1, Maven, Spring Security, Angular 17.3). This guide is now on the path of the pilot, thus the issue moves to the milestone "B. Pilot numbers", and #43 waits for it. The guide must give the Maven coordinates and the JitPack repository block for a `pom.xml`, next to the Gradle form.

xamcross:
Note on 2026-09-21 from the maintainer, from the second product review of #87: https://github.com/xamcross/octometer/pull/87#issuecomment-5763057976

[...]

PR #165, opened 2026-09-22
feat(kit): stop the ingest above the event cap

### LINKED ISSUE #34: Stop the ingest above the event cap

xamcross:
## Summary

Above `OCTOMETER_MAX_EVENTS`, `MongoEventLogStore` now drops each new
batch. The store never runs a count query on each request; it caches
the count for 60 seconds (design decision D21). `kit/jvm-core` adds a
small interface and a cache class, with no MongoDB dependency. The
ingest route needs no code change: a dropped batch already gives 204,
the same answer as a stored batch (contract rule C19).

**Correction round 1** (this update) closes 2 BLOCKER, 2 MAJOR, and
11 MINOR findings of the two reviews. See the round-1 comment on this
pull request for the full list, the fix, and the evidence of each one.

## Acceptance criteria

- [x] Above `OCTOMETER_MAX_EVENTS` the store drops the batch, the
  route returns 204, and the store writes one warning.
  Evidence: `MongoEventLogStoreEventCapTest.aBatchAboveTheCapIsDroppedAndTheStoreWritesOneWarning`
  and `IngestRouteTest`, the test named
  "a batch that the store drops above the event cap still gives 204
  and an empty body".
- [x] Fixed clock: the store calls `estimatedDocumentCount()` a
  maximum of one time in 60 seconds.
  Evidence: `EventCapGuardTest.theGuardCallsTheEstimatorAtMostOneTimeInsideOneRefreshInterval`
  (1000 calls, 1 read) and
  `EventCapGuardTest.oneThousandCallsFrom8ThreadsReadTheEstimateOneTimeOnly`
  (8 threads, 1000 calls, 1 read).
- [x] The test deletes documents, as the TTL does. After the next
  count the store accepts a batch again.
  Evidence: `MongoEventLogStoreEventCapTest.aDeleteOfEveryDocumentAsTheTtlDoesLetsTheStoreAcceptABatchAgainAfterTheNextCount`.
- [x] `./gradlew :kit:jvm-mongo:test` and `./gradlew :kit:jvm-ktor:test` pass.
  Evidence below.
- [x] A comment on this issue records the test output. Posted after
  round 1 too, with the round-1 numbers.

## Test output (round 1, JDK 21)

- `./gradlew :kit:jvm-core:test`: 254 tests, 0 failures, 0 errors.
- `./gradlew :kit:jvm-ktor:test`: 44 tests, 0 failures, 0 errors.
- `./gradlew :kit:jvm-mongo:test` (Docker): 63 tests, 0 failures, 0 errors.
- `./gradlew :kit:jvm-mongo:testMongoSyncNewest` (the newest 5.x driver, Docker, run alone): 63 tests, 0 failures, 0 errors.
- `./gradlew build`: BUILD SUCCESSFUL, every module.

## The merge with `origin/main`

Pull request #157 (issue #35) merged into `main` as commit `1f7f436`
after this branch first merged `main`. A second merge of `origin/main`
(merge commit `fd01be1`) found the four conflicts that the MongoDB
review named:

- `kit/jvm-core/src/main/java/octometer/kit/core/store/InMemoryEventLogStore.java`
- `kit/jvm-core/src/test/java/octometer/kit/core/ingest/CapturingLoggerFinder.java`
- `kit/jvm-core/src/test/java/octometer/kit/core/store/InMemoryEventLogStoreTest.java`
- `kit/jvm-mongo/src/main/java/octometer/kit/mongo/store/MongoEventLogStore.java`

Each conflict was a class-comment or a test-name clash between the two
features (the event cap of this issue, the user erasure of issue #35).
Each side kept its own paragraph or its own test; no line of either
feature was dropped. `./gradlew :kit:jvm-core:test` passed after the
merge (248 tests, then 254 after round 1's new tests), and
`./gradlew build` passed for the whole workspace.

1. **Where the cap check lives.** `kit/jvm-core` adds
   `EventCountEstimator` (one method, `estimatedEventCount()`) and
   `EventCapGuard` (the 60-second cache, the cap check, and a
   once-per-hour warning), with no MongoDB dependency. This is the
   "small interface that a store implements" from my brief.
   `MongoEventLogStore` implements `EventCountEstimator` through
   `estimatedDocumentCount()`, and it holds one `EventCapGuard`
   field. `InMemoryEventLogStore` also implements
   `EventCountEstimator` (an exact count from its own queue size), so
   a test of the guard needs no MongoDB dependency.
2. **The refresh interval.** 60 seconds, from D21 directly. Round 1
   bounds the overshoot: `EventCapGuard.countAppended` adds each
   accepted batch to the cached estimate between two refreshes, so a
   flood inside one window stops at the cap too (security MAJOR 1).
   Test: `EventCapGuardTest.countAppendedAddsToTheCachedEstimateInsideOneRefreshInterval`
   and `MongoEventLogStoreEventCapTest.aFloodInsideOneRefreshIntervalStopsAtTheCap`.
3. **The warning rate.** One warning for each elapsed hour, not one
   for each dropped batch, and not tied to the UTC clock hour (round
   1 correction; see the round-1 comment).
4. **A failed count fails open, then fails closed (round 1).** The
   guard keeps the last good estimate on a count error. After 3
   errors in a row it fails closed: the store drops each batch until
   a count succeeds, with one warning each hour that holds the
   MongoDB error code. Tests:
   `EventCapGuardTest.aCountErrorKeepsTheLastGoodEstimateInsteadOfResettingItToZero`
   and
   `EventCapGuardTest.aThirdCountErrorInARowFailsClosedAndASuccessReopensTheGuard`.
5. **`CapturingLoggerFinder` of `kit/jvm-core`.** `messages()` and
   `clear()` are `public`, so a test of `octometer.kit.core.store`
   can read the one registered `System.LoggerFinder` of the module.
   Round 1: `messages()` now returns a copy of the recorded messages
   (MongoDB MINOR 5).

## Notes for the reviewer

- `origin/main` is merged into this branch with a merge commit; see
  "The merge with `origin/main`" above for the commit and the
  conflicts. This pull request no longer claims that it stays out of
  the way of #157 (MongoDB MINOR 1 of the round-0 review); the two
  branches touched the same files, and the merge is now in the
  history.
- `kit/jvm-ktor` gets a new test only; `IngestRoute.kt` needs no
  production line change, because a store that silently drops a
  batch already reaches the existing `call.respond(HttpStatusCode.NoContent)`
  path.
- For each rule of this pull request, round 0 and round 1 each broke
  the implementation for a moment, ran the test, saw the failure, and
  reverted the change before the commit. The round-1 comment lists
  each break for round 1.

Closes #34

Generated with Claude Code


xamcross:

I made my own clone, I ran `gh pr checkout 165` in it, and I ran each command myself with
JDK 21. I did not trust the pull request text.

## The acceptance criteria

|---|---|---|---|
| 1 | Above `OCTOMETER_MAX_EVENTS` the store drops the batch, the route gives 204, the store writes one warning | PASS | `:kit:jvm-mongo:test` — `MongoEventLogStoreEventCapTest` 4 tests, 0 failures. `:kit:jvm-ktor:test` — 42 tests, 0 failures. |
| 2 | Fixed clock: a maximum of one count in 60 seconds | PASS | `:kit:jvm-core:test` — 238 tests, 0 failures. A change of `EVENT_CAP_REFRESH_INTERVAL` to 3600 seconds makes 2 tests fail (see below). |
| 3 | A delete of each document, as the TTL does, lets the store accept a batch again | PASS | `MongoEventLogStoreEventCapTest.aDeleteOfEveryDocumentAsTheTtlDoesLetsTheStoreAcceptABatchAgainAfterTheNextCount`. |
| 4 | `:kit:jvm-mongo:test` and `:kit:jvm-ktor:test` pass | PASS | mongo 43/0/0, ktor 42/0/0, core 238/0/0. Docker ran the container tests; 0 tests were skipped. |
| 5 | A comment on the issue records the test output | PASS | The comment of 2026-09-22 on issue #34. |

Each criterion passes. The findings below name risks that the criteria do not cover.

## The tests are real

I broke the code four times and I saw the correct test fail each time.

1. `EventCapGuard:106`, `>=` to `>`: `anEstimateAtTheCapIsOverCap` and
   `aLowerEstimateAfterADeleteLetsTheGuardAcceptEventsAgain` fail.
2. `EventCapGuard:117`, `return 0L` to `return Long.MAX_VALUE`:
   `anEstimatorThatThrowsFailsOpenAndTheBatchIsNotOverCap` fails.
3. `EventCapGuard:145-153`, the throttle removed:
   `warnDropOncePerHourWritesOneLogLineForTheHourWithNoUserData` fails.
4. `MongoEventLogStore:103`, 60 seconds to 3600 seconds:
   `aBatchAboveTheCapIsDroppedAndTheStoreWritesOneWarning` and
   `aDeleteOfEveryDocumentAsTheTtlDoesLetsTheStoreAcceptABatchAgainAfterTheNextCount` fail.

I restored each file. My clone is clean. The diff holds 10 source files, no secret, no database
file, and no build output.

## BLOCKER 1 — the fail-open state has no bound, and an attacker can hold the store in it

`kit/jvm-core/src/main/java/octometer/kit/core/store/EventCapGuard.java:109-119`

`readEstimateOrFailOpen` catches each `RuntimeException`, it caches the estimate 0, and it
returns 0. Nothing counts the failures in a row. Nothing bounds the time in this state. While
the count query fails, the cap is off.

**My evidence.** I added a temporary probe test to my clone. It gives a guard the cap 1 and an
estimator that always throws. It then calls `isOverCap()` one time for each refresh interval
across 24 simulated hours:

```
PROBE failOpen: estimatorCalls=1440 overCapCount=0 warnings=24
```

1440 refreshes, 0 of them over the cap, 24 warnings. The cap never starts again. I deleted the
probe file.

**Each way to force a count error from outside.**

1. **A role without the right.** `estimatedDocumentCount()` needs a read privilege.
   `MongoEventLogStoreInsertOnlyUserTest` already proves that an app user with only the
   `insert` action cannot run it. In that deployment the cap is off from the first request, for
   ever, and the app fills the M0 cluster. This is not an attack. It is a deployment that this
   repository tests and supports today.
2. **A slow cluster or a network fault.** The driver throws `MongoTimeoutException` or
   `MongoSocketReadException`. Both extend `RuntimeException`, so both fail open. An M0 cluster
   accepts 500 connections. A flood saturates them, the count query times out, and each next
   refresh times out too. The attacker holds the store in the fail-open state for the whole
   flood. The cap is off at the exact moment that design decision D21 needs it.
3. **A throttle of the Atlas project.** Atlas throttles an M0 cluster above its operation rate.
   The count query is one more operation, thus it fails with the flood.

Decision 4 of the pull request text says that a count failure must never turn into a full
ingest outage. For this issue that trade is the wrong way round. Issue #34 states the reason
itself: "A full Atlas M0 cluster refuses each write of the app, not only the metrics." A
dropped batch costs a metric. A full cluster costs the whole app.

**The exact fix.** Bound the fail-open state in `EventCapGuard`:

- Add a field `private final AtomicInteger failuresInARow = new AtomicInteger();` and a
  constant `MAX_FAILURES_IN_A_ROW = 3`.
- On a success, call `failuresInARow.set(0)`.
- On a failure, keep the estimate of the previous snapshot, not 0, and return `maxEvents` (fail
  closed) once `failuresInARow.incrementAndGet() >= MAX_FAILURES_IN_A_ROW`.
- Keep the warning each hour, and state the new state in its text.

The result: the store accepts a batch for a maximum of 3 refresh intervals (3 minutes) after
the first count failure, then it drops each batch until one count succeeds. Add one test for
the change to the closed state, and one test for the return to the open state after a count
succeeds.

## BLOCKER 2 — a fail-closed guard needs a documented privilege, and no document states it

`kit/jvm-mongo/README.md`, `docs/superpowers/specs/2026-09-21-octometer-design.md:364`

This pull request adds a new requirement for the app database user: the `find` privilege on
`octometer_events`. No file states it. `kit/jvm-mongo/README.md` names no environment variable
and no privilege. Design section 4.4 names only the reader user of the monitor. An operator who
follows the least-privilege guidance gets the silent, permanent fail-open of BLOCKER 1. The fix
of BLOCKER 1 turns the same deployment into a silent, permanent ingest stop, thus the two
findings must ship together.

**The exact fix.** Add to `kit/jvm-mongo/README.md` a section with:

- the variable `OCTOMETER_MAX_EVENTS`, the default 200000, and the behaviour of a bad value;
- the sentence: "The app database user needs the `find` privilege on `octometer_events`. The
  event cap reads `estimatedDocumentCount()`. Without that privilege the cap cannot work."

Add the same sentence to design decision D21.

## MAJOR 1 — the overshoot inside one refresh window has no bound

`kit/jvm-core/src/main/java/octometer/kit/core/store/EventCapGuard.java:94-107`,
`kit/jvm-mongo/src/main/java/octometer/kit/mongo/store/MongoEventLogStore.java:175-178`

The guard compares the cached estimate only. It never counts the events that the store writes
after that estimate. The overshoot is therefore the whole write volume of one 60-second window.
`theStoreReadsTheCountAtMostOneTimeInsideOneRefreshIntervalAndCanOvershootTheCapOnce` shows the
behaviour with 2 events; the production number is much larger.

**The pull request text states a wrong bound.** Decision 2 says: "in practice the rate limiter
of D20 and the anonymous caps of D43 bound the write volume of 60 seconds." Neither one bounds
it today.

- D43 is not implemented. Issue #117 adds it.
- `IngestRateLimiter.check` takes the client address text as the key with no change
  (`IngestRateLimiter.java:119`). The /64 rule of D43 is absent. One IPv6 /64 therefore gives
  2^64 keys, not one key.
- `IngestRateLimiter.KeyMap` evicts the oldest key when the map holds 20000 keys
  (`IngestRateLimiter.java:245-250`). A new key always starts at count 1. An attacker who uses
  a new address for each request is never limited.

**The number.** One batch holds 50 clicks (rule C17) in 16 KB (rule C18). One stored event
costs about 350 bytes (design section 2: about 70 MB for 200000 events).

- One IPv4 address: 120 requests each minute × 50 clicks = 6000 events, about 2.1 MB.
- One IPv6 /64 with a new address for each request: the limiter gives no bound. The bound comes
  from the app and the cluster. The route runs 8 store calls at the same time
  (`defaultStoreDispatcher`). With a round trip of 30 ms for one `insertMany` of 50 documents,
  the app writes about 13000 events each second, thus about 800000 events in 60 seconds, about
  **280 MB**. The operation ceiling of an M0 cluster (about 100 operations each second) lowers
  the number to about 300000 events, about **105 MB**.

Both numbers are above the 70 MB that the full cap of 200000 events uses. One overshoot window
can therefore write more than the whole cap allows. A TTL delete that drops the count under the
cap opens a new window each time.

**The exact fix.** Count the accepted events in `EventCapGuard`:

- Add `private final AtomicLong appendedSinceRefresh = new AtomicLong();`.
- Add `public void countAppended(long eventCount)`, which adds to that field.
- Reset the field to 0 inside each refresh.
- Change line 106 to `return current.estimate() + appendedSinceRefresh.get() >= maxEvents;`.
- Call `eventCapGuard.countAppended(documents.size())` in `MongoEventLogStore.append` after the
  `insertMany` call succeeds.

The overshoot then falls to one batch, a maximum of 50 events. Add a test: with the cap 1, two
appends inside one refresh interval must store 1 document, not 2.

Also correct decision 2 of the pull request text. Name issue #117 as the open work, and state
that the limiter of #33 gives no bound for an attacker with many addresses.

## MAJOR 2 — a slow count query blocks each ingest thread

`kit/jvm-core/src/main/java/octometer/kit/core/store/EventCapGuard.java:97-105`

The guard calls the estimator inside `synchronized (refreshLock)`. Each other caller waits for
the whole network call.

**My evidence.** The second probe test gives the guard an estimator that sleeps 2000 ms. A
second thread calls `isOverCap()` while the first thread is inside the estimator:

```
PROBE lockHold: a second caller waited 2000 ms
```

The Javadoc of `MongoEventLogStore` (lines 62-70) states that the store sets no timeout, and
that the driver has no socket read timeout by default. The Ktor route runs a maximum of 8 store
calls at the same time. One blocked count query therefore stops the whole ingest route, and
each waiting request holds its connection. Before this change the request path held no read
operation at all.

**The exact fix.** Move the estimator call out of the lock. Use an
`AtomicBoolean refreshInFlight`. The caller that wins `compareAndSet(false, true)` reads the
estimator, writes the new snapshot, then sets the flag to false. Each other caller uses the
current snapshot and returns at once. Add the timeout requirement for the count path to the
Javadoc of `MongoEventLogStore`.

## The other checks that I ran

**The answer above the cap.** PASS. `MongoEventLogStore.append` returns with no exception, thus
the route reaches `call.respond(HttpStatusCode.NoContent)`. `IngestRouteTest` proves the status
204 and the empty body. The two warning texts hold no user id, no client address, and no
session id. The route writes no log line for a dropped batch.

**The warning throttle.** PASS. `warnOncePerHour` writes a maximum of one line for each
60-minute span of the clock, for each of the two counters. A tight loop of dropped requests
cannot flood the log. My probe confirms 24 lines across 24 hours.

**The resume after a TTL delete.** PASS. The Testcontainers test deletes each document, it
moves the injected clock forward by 60 seconds, and the store then accepts a batch.

**`OCTOMETER_MAX_EVENTS`.** The value 0, a negative value, and a text each give the default
200000 with one WARNING that names the variable and never repeats the value
(`MongoEventLogStoreMaxEventsTest`, 7 tests). A value with more digits than a `long` also gives
the default with the warning. A value above the D21 default gives no error and no warning; see
MINOR 1.

## MINOR

1. `MongoEventLogStore.java:351-370`: `maxEventsFromValue` accepts each positive `long`. A value
   of 100000000000 passes with no warning, and the collection then grows past the 512 MB of an
   M0 cluster. `clampRetentionDays` (line 372) already clamps its own variable and warns. Add
   the same clamp, for example at 1000000 events, with one warning.
2. `EventCapGuard.java:146`: the throttle keys on the hour of the clock, not on the time since
   the last line. Two lines can therefore land one second apart across an hour boundary. The
   count stays bounded. Keep the last warn instant instead, and compare the time that passed.
3. `EventCapGuard.java:121-124`: a backward step of the clock freezes the snapshot.
   `now - checkedAtMillis` is then negative, thus the guard refreshes no more until the clock
   passes the old instant. `IngestRateLimiter.windowFor` handles the same case on purpose. Add
   `|| now < current.checkedAtMillis()` to `needsRefresh`.
4. A dropped batch returns much faster than a stored batch, because the store makes no network
   call. A client can measure the difference and learn that the cap is active. It learns no user
   data. No fix is needed; a note in the Javadoc is enough.
5. `EventCapGuard.java:18-23` and `21-23`: two Javadoc sentences hold about 30 words each. The
   ASD-STE100 limit for a description is 25 words. Split each one.

## The verdict

CHANGES REQUESTED. 2 BLOCKER findings, 2 MAJOR findings, 5 MINOR findings. The code is careful
and the tests are real, but the cap does not hold against the threat that issue #34 names: a
flood that an attacker sends from many addresses turns the cap off (BLOCKER 1), and one refresh
window accepts more events than the whole cap allows (MAJOR 1).

xamcross:

I made my own clone, I ran `gh pr checkout 165`, and I ran each command on JDK 21
with Docker. Each number below comes from my own run.

## The acceptance criteria of issue #34

|---|---|---|---|
| 1 | Above the cap the store drops the batch, the route returns 204, the store writes one warning | PASS | `MongoEventLogStoreEventCapTest.aBatchAboveTheCapIsDroppedAndTheStoreWritesOneWarning` passes against `mongo:7.0`. I removed the cap check at `MongoEventLogStore.java:175`; the test failed. I removed `warnDropOncePerHour()` at `MongoEventLogStore.java:176`; the same test failed on the warning count only. The 204 answer: `IngestRouteTest`, the test "a batch that the store drops above the event cap still gives 204 and an empty body". |
| 2 | A fixed clock: the store calls `estimatedDocumentCount()` a maximum of one time in 60 seconds | PASS | I changed `needsRefresh` at `EventCapGuard.java:129` to a refresh on each call. `MongoEventLogStoreEventCapTest.theStoreReadsTheCountAtMostOneTimeInsideOneRefreshIntervalAndCanOvershootTheCapOnce` failed with "Inside one refresh interval, the store accepts a batch above the cap once". The same change failed 3 tests of `EventCapGuardTest`. |
| 3 | The test deletes each document, as the TTL does. After the next count the store accepts a batch again | PASS | `MongoEventLogStoreEventCapTest.aDeleteOfEveryDocumentAsTheTtlDoesLetsTheStoreAcceptABatchAgainAfterTheNextCount` passes. Two separate mutations broke it. |
| 4 | `./gradlew :kit:jvm-mongo:test` and `./gradlew :kit:jvm-ktor:test` pass | PASS | See the counts below. |
| 5 | A comment on the issue records the test output | PASS | The comment of 2026-09-22 on issue #34. Its numbers agree with mine. |

## The test counts, from my own runs (JDK 21, Docker)

| Task | Tests | Skipped | Failures | Errors |
|---|---|---|---|---|
| `:kit:jvm-core:test` | 238 | 0 | 0 | 0 |
| `:kit:jvm-ktor:test` | 42 | 0 | 0 | 0 |
| `:kit:jvm-mongo:test` | 43 | 0 | 0 | 0 |
| `:kit:jvm-mongo:testMongoSyncNewest` | 43 | 0 | 0 | 0 |

Each number agrees with the pull request text.

## The strength of the tests

I broke the code 12 times. Each time the right test failed, and no time did the
suite stay green.

| # | The change | The test that failed |
|---|---|---|
| M1 | `EventCapGuard.java:112`, `>=` becomes `>` | `anEstimateAtTheCapIsOverCap`, `aLowerEstimateAfterADeleteLetsTheGuardAcceptEventsAgain` |
| M2 | `EventCapGuard.java:129`, a refresh on each call | `theGuardCallsTheEstimatorAtMostOneTimeInsideOneRefreshInterval` and 2 more |
| M3 | `EventCapGuard.java:123`, a failed read rethrows | `anEstimatorThatThrowsFailsOpenAndTheBatchIsNotOverCap`, and `MongoEventLogStoreInsertOnlyUserTest` with the MongoDB error 13 |
| M4 | `EventCapGuard.java:154`, no hour throttle | `warnDropOncePerHourWritesOneLogLineForTheHourWithNoUserData` |
| M5 | `InMemoryEventLogStore.java:62`, the estimate is always 0 | `estimatedEventCountReflectsTheStoredEventCount` |
| M6 | `EventCapGuard.java:129`, no refresh after the first read | 7 tests of `EventCapGuardTest` |
| M7 | `MongoEventLogStore.java:175`, no cap check | 2 tests of `MongoEventLogStoreEventCapTest` |
| M8 | `MongoEventLogStore.java:176`, no warning | `aBatchAboveTheCapIsDroppedAndTheStoreWritesOneWarning` |
| M9 | `MongoEventLogStore.java:95`, the default cap becomes 100 | 4 tests of `MongoEventLogStoreMaxEventsTest` |
| M10 | `MongoEventLogStore.java:166`, the estimate is always 0 | 2 tests of `MongoEventLogStoreEventCapTest` |
| M11 | the same as M3, against the insert-only container | `MongoEventLogStoreInsertOnlyUserTest` |
| M12 | the same as M2, against the container | `theStoreReadsTheCountAtMostOneTimeInsideOneRefreshIntervalAndCanOvershootTheCapOnce` |

I found no empty test and no always-green test. The three `assertFalse(message.contains(...))`
lines in `EventCapGuardTest.java:459-461` cannot fail today, because the warning text is a
constant with no interpolation. They still guard a later change, so I accept them.

## MAJOR 1 — the cap never applies under the minimum role of issue #70

`kit/jvm-mongo/src/main/java/octometer/kit/mongo/store/MongoEventLogStore.java:166`
and `kit/jvm-mongo/README.md`.

**The evidence.** I ran a probe against `mongo:7.0` with `mongod --auth` and the
compile driver 5.0.1. I built four custom roles on `octometer_events` and called
`estimatedDocumentCount()` with each one.

| The role | The result of `estimatedDocumentCount()` |
|---|---|
| `insert`, `createIndex`, `collMod` (the minimum role of the note on issue #70) | error 13 `Unauthorized`, on the command `count` |
| `insert`, `createIndex`, `collMod`, `find` | 3 (it works) |
| `insert`, `createIndex`, `collMod`, `collStats` | error 13 `Unauthorized`, on the command `count` |
| `find` | 4 (it works) |

Driver 5.x sends the `count` command for `estimatedDocumentCount()`. That command
needs the `find` action on the collection. The action `collStats` does not help, and
`dbStats` does not help.

**Why this matters.** In the same probe I built the store with `maxEvents = 1` under
the minimum role of issue #70. The count threw, `EventCapGuard` failed open, and
`append` stored the event. The document count went from 2 to 3. So an app that
follows the documented minimum role never stops the ingest above the cap. Issue #34
is a `production-blocker`, and for that app the blocker stays open. The one warning
for each hour is the only signal, and it names no MongoDB error code, so an operator
cannot tell a permission gap from a network fault.

I accept the fail-open rule itself. A count failure must never stop each write.
The defect is the silence: the kit documents a role that turns the new feature off.

**The exact fix.**
1. Add a section to `kit/jvm-mongo/README.md` with this text: "The minimum role of
   the app database user is a custom role on `octometer_events` with the actions
   `insert`, `createIndex`, `collMod`, and `find`. The action `find` lets the store
   run `estimatedDocumentCount()` for the event cap of design decision D21. Without
   `find` the store fails open: it accepts each batch, and the cap never stops the
   ingest. The built-in role `readWrite` holds `find` and `insert`, and it holds no
   `collMod`."
2. Add the same rule to the class comment of `MongoEventLogStore`, in the paragraph
   that describes the cap.
3. Put the numeric MongoDB error code in the fail-open warning of
   `EventCapGuard.readEstimateOrFailOpen` (`EventCapGuard.java:119`), the same way as
   `MongoEventLogStore.warnIndexError` does for the index step. A store can pass the
   code, or the guard can read it from a `MongoCommandException` through a small
   callback. The code stays a number, so design decision D15 stays true.
4. Add a test with the role `insert`, `createIndex`, `collMod`, `find` that proves
   the cap does stop a batch under a restricted user. Today only an unrestricted user
   proves the cap.

Please also tell the maintainer of issue #70, because the note of 2026-09-22 on that
issue states the old role.

## The answer to the overshoot question

The guard has no bound of its own on the overshoot. The overshoot of one window is
the whole write volume of 60 seconds. The numbers of the code on `main` today:

- Contract rule C17: 50 clicks in one batch.
- `IngestRateLimiter.USER_LIMIT_PER_WINDOW` = 30 requests for each user key in 60
  seconds. One user key gives 30 x 50 = **1 500 events**.
- `IngestRateLimiter.CLIENT_ADDRESS_LIMIT_PER_WINDOW` = 120 requests for each client
  address in 60 seconds. One address gives 120 x 50 = **6 000 events**.
- The maps hold 5 000 user keys and 20 000 address keys, and the limiter sets no
  global ceiling. Many keys therefore give 7 500 000 and 120 000 000 events in one
  window on paper.
- With decision D43 of issue #117: 300 requests x 50 clicks = 15 000 events for each
  key, and the counter of 900 click entries lowers it to 900 events for each key.

**My judgement against D21.** D21 asks for exactly this trade, thus the pull request
follows the decision. Section 2 of the design measures about 350 bytes for one event
with the field `path`, so 200 000 events use about 70 MB. The step from 70 MB to the
512 MB of an M0 cluster is about 1 260 000 events. One busy key overshoots by 6 000
events, which is 0.5 percent above the cap, and the pilot app has fewer than 500
visits each day. About 210 client addresses at the full limit of D20 fill the whole
step inside one window. So D21 protects the pilot app, and it does not protect an M0
cluster against a flood from many addresses. That residual risk sits with issue #117,
which the issue already names as Related. I raise no blocker for it. Please add the
two numbers (1 500 and 6 000 events for one key) to the class comment of
`EventCapGuard`, so a later reader sees the size of the accepted overshoot.

## The cost of the refresh

- No count query runs on each request. `theGuardCallsTheEstimatorAtMostOneTimeInsideOneRefreshInterval`
  makes 1 000 calls and reads the estimate 1 time.
- Under contention the refresh runs one time only.
  `oneThousandCallsFrom8ThreadsReadTheEstimateOneTimeOnly` proves it, and my own
  probe with 8 threads and an estimator of 600 ms gave 1 estimator call.
- See MINOR 2 for the block time.

## MINOR findings

**MINOR 1 — the pull request text states a wrong merge result.**
The "Notes for the reviewer" section says the change "stays out of the way of pull
request #157". I fetched `pull/157/head` and I ran a local merge. Git reported four
conflicts:

- `kit/jvm-core/src/main/java/octometer/kit/core/store/InMemoryEventLogStore.java`
- `kit/jvm-core/src/test/java/octometer/kit/core/ingest/CapturingLoggerFinder.java`
- `kit/jvm-core/src/test/java/octometer/kit/core/store/InMemoryEventLogStoreTest.java`
- `kit/jvm-mongo/src/main/java/octometer/kit/mongo/store/MongoEventLogStore.java`

I aborted the merge at once. `EventLogStore.java` gives no conflict, so that part of
the claim holds. Pull request #157 also edits `kit/jvm-mongo/README.md`, thus the fix
of MAJOR 1 touches a file of #157 too. Please correct the pull request text, and
please agree an order with #157.

**MINOR 2 — the refresh blocks each request that arrives at the same moment.**
`EventCapGuard.java:104`. The refresh runs inside `synchronized (refreshLock)`. I ran
a probe with 8 threads at the refresh moment and an estimator of 600 ms. The
estimator ran 1 time, and each of the 8 calls of `isOverCap()` took 600 ms. The Ktor
route uses `Dispatchers.IO.limitedParallelism(8)`, so one count round trip holds the
whole store dispatcher for its length, one time in each 60 seconds. On a healthy M0
cluster that is a few milliseconds. **The fix:** hold an `AtomicBoolean` for a refresh
in flight. The thread that wins the flag reads the estimate. Each other thread returns
the previous cached value at once. The rule "one read in each 60 seconds" stays true,
and no second request waits.

**MINOR 3 — "one warning for each hour" is one warning for each UTC clock hour.**
`EventCapGuard.java:152`. The method rounds `now` down to a UTC hour boundary. I ran
a probe with two drops 2 seconds apart, across the boundary 10:59:59Z to 11:00:01Z.
The guard wrote 2 warnings. The bound of 24 lines for each day still holds, so I raise
no defect. **The fix:** change the words "for each 60-minute span" to "for each UTC
clock hour" in the class comment and in the comment of `warnDropOncePerHour`.

**MINOR 4 — the package-private constructor reads the process environment.**
`MongoEventLogStore.java:138`. `MongoEventLogStore(MongoDatabase, int)` now calls
`maxEventsFromEnvironment()`. A developer with `OCTOMETER_MAX_EVENTS=5` in the shell
breaks `MongoEventLogStoreTest` and `MongoEventLogStoreInsertOnlyUserTest`, with no
change of the code. **The fix:** let each container test call the four-argument
constructor with an explicit cap.

**MINOR 5 — `CapturingLoggerFinder.messages()` hands out the live deque.**
`kit/jvm-core/src/test/java/octometer/kit/core/ingest/CapturingLoggerFinder.java:29`.
The class sits in `src/test`, and the published source jar holds the main source set
only, thus I see no API risk. The new `public` modifier lets a test of any package
change the recorded messages. **The fix:** return `List.copyOf(MESSAGES)`, or keep the
deque and state in the comment that a caller must only read it. The `public` change
itself is correct, and the module still registers one `System.LoggerFinder`.

**MINOR 6 — the fast count can drift after an unclean shutdown.**
`MongoEventLogStore.java:165`. `estimatedDocumentCount()` reads the count of the
storage engine metadata. After an unclean shutdown that number can differ from the
true count until an administrator runs `validate`. The cap then works on a wrong
number. **The fix:** add one sentence with this fact to the comment of
`estimatedEventCount()`.

**MINOR 7 — the cap check runs before the per-event null check.**
`MongoEventLogStore.java:175`. Above the cap, `append` returns before the loop that
calls `Objects.requireNonNull(event, ...)`. A batch with a null event therefore throws
below the cap, and it returns silently above the cap. `IngestPipeline` never passes a
null event, so no user hits this today. **The fix:** state the order in the comment of
`append`, or move the cap check after the loop.

## The other checks

- **The seam.** `EventCountEstimator` and `EventCapGuard` sit in `kit/jvm-core`, and
  that module has no run-time dependency at all. `InMemoryEventLogStore` implements
  the interface. Correct.
- **A consumer with its own store (rule C11).** It gets neither a default method nor a
  compile error. `EventLogStore` keeps its two methods, so a custom store builds with
  no change, and it simply has no cap. I judge this the right choice: a new method on
  a published interface would break each consumer. The comment of `EventLogStore`
  already states that a store can drop a batch above the cap. Please name
  `EventCapGuard` there, so a consumer finds the ready class.
- **The regression of `MongoEventLogStoreInsertOnlyUserTest`.** I confirmed both
  halves. I removed the fail-open at `EventCapGuard.java:123`, and the test failed
  with the MongoDB error 13 `Unauthorized` on the command `count`. With the fail-open
  in place the test passes. The fix is correct. The test asserts the log records
  before `append` only, so it does not prove the fail-open warning. Please add that
  assertion.
- **The contract.** Rule C19 of version 1.2 already names "a batch that the event cap
  drops". The pull request needs no contract change. Correct.
- **Design decision D15.** The drop warning and the failure warning hold no user id,
  no client address, and no session id. I read both constants. Correct.
- **The scope and the diff.** 10 files, and each one belongs to issue #34. I found no
  secret, no database file, and no build output.
- **The language.** Each new comment and the commit body follow ASD-STE100. I found no
  defect.

## What I need before an approval

Fix MAJOR 1. The MINOR findings are for you to judge.

xamcross:
## Correction round 1

I made my own clone at a new path, I ran `gh pr checkout 165` in it, and I ran
each command myself with JDK 21 and Docker. This comment lists each finding of
the two round-0 reviews, the fix, the file and line, and the evidence.

### The merge with `origin/main`

Merge commit `fd01be1` merges `origin/main` (which now holds pull request #157,
commit `1f7f436`) into this branch. Git reported the four conflicts that the
MongoDB review found:

- `kit/jvm-core/.../store/InMemoryEventLogStore.java` — a class-comment
  conflict; both paragraphs kept, in sequence.
- `kit/jvm-core/.../ingest/CapturingLoggerFinder.java` — a class-comment
  conflict; both paragraphs kept, and I added the "returns a copy" behaviour
  in the same edit (MINOR 5 below).
- `kit/jvm-core/.../store/InMemoryEventLogStoreTest.java` — two branches added
  a test after the same line; both tests kept, one renamed to match the final
  name that pull request #157 gave it on `main`.
- `kit/jvm-mongo/.../store/MongoEventLogStore.java` — an import conflict and a
  class-comment conflict; both features' imports and paragraphs kept.

I also fixed one merge fallout: `kit/jvm-ktor/src/test/kotlin/octometer/kit/ktor/IngestRouteTest.kt:132`,
a fake `EventLogStore` whose `deleteByUserId` still returned `Unit` after the
merge; it now returns `DeletionResult(0, 0, true)`.

`./gradlew :kit:jvm-core:test` passed after the merge (248 tests, before
round-1's new tests), and `./gradlew build` passed for the whole workspace.

### BLOCKER 1 (security) — the fail-open state had no bound

**Fix.** `kit/jvm-core/.../store/EventCapGuard.java`. A new `AtomicInteger
failuresInARow` (line 86) and `MAX_FAILURES_IN_A_ROW = 3` (line 71). On a
count success, `readEstimateOrKeepLastGood` (line 194) resets the counter to
zero. On a failure, it keeps the last good cached estimate instead of
resetting it to zero (line 214); once the failure count reaches 3, it returns
`maxEvents` (line 208), so `isOverCap()` reports over-cap until a read
succeeds.

**Test.** `EventCapGuardTest.aCountErrorKeepsTheLastGoodEstimateInsteadOfResettingItToZero`
and `aThirdCountErrorInARowFailsClosedAndASuccessReopensTheGuard` (an injected
`MutableClock` and a `FlakyEstimator`, proving all three states: a normal
read, a kept estimate on failures 1–2, and fail-closed from failure 3, with a
later success reopening the guard).

**Break/restore.** I changed `failures >= MAX_FAILURES_IN_A_ROW` to
`failures >= 999999` at `EventCapGuard.java:202`; the fail-closed test failed.
I restored the file and reran; it passed.

### BLOCKER 2 (security) / MAJOR 1 (MongoDB) — the `find` right was undocumented

**Fix.** `kit/jvm-mongo/README.md` (new section "The event cap") and the
class comment of `MongoEventLogStore.java:87–94` now state: the app database
user needs `find`, next to `insert`, `createIndex`, and `collMod`, and that
`estimatedDocumentCount()` sends the MongoDB `count` command, which needs
`find`. `MongoEventLogStore.java:239–250` adds `mongoErrorCodeOf`, passed into
`EventCapGuard`'s new 5-argument constructor (`EventCapGuard.java:137`), so a
fail-open or fail-closed warning now names the numeric MongoDB error code, not
just "unknown".

**Test (new).** `MongoEventLogStoreRestrictedRoleEventCapTest` (new file): a
Testcontainers user with exactly `insert`, `createIndex`, `collMod`, `find`
proves the cap stops a batch. `MongoEventLogStoreInsertOnlyUserTest.java:108-123`:
the existing insert-only test (no `find`) now also asserts the guard's
fail-open warning, with `MongoDB error code 13`.

**Break/restore.** I commented out the cap check in `append()`
(`if (false && eventCapGuard.isOverCap())`); the new restricted-role test
failed. Restored, reran, passed. I also changed `mongoErrorCodeOf` to always
return `"unknown"`; the insert-only test's error-code assertion failed.
Restored, reran, passed.

### MAJOR 1 (security) — no bound inside one refresh window

**Fix.** `EventCapGuard.java`: `AtomicLong appendedSinceRefresh` (line 84),
`countAppended(long)` (line 188, public), added to the cap check at line 179
(`current.estimate() + appendedSinceRefresh.get() >= maxEvents`), reset to
zero inside each refresh (line 174). `MongoEventLogStore.append` calls
`eventCapGuard.countAppended(documents.size())` after a successful
`insertMany` (line 299).

**Test.** `EventCapGuardTest.countAppendedAddsToTheCachedEstimateInsideOneRefreshInterval`,
`aRefreshResetsTheAppendedCounter`, and
`MongoEventLogStoreEventCapTest.aFloodInsideOneRefreshIntervalStopsAtTheCap`
(5 appends inside one window, cap 1, store ends with 1 document, not 5). This
replaces the old `...CanOvershootTheCapOnce` test, whose old assertion (2
documents) is now the wrong, pre-fix behaviour.

**Break/restore.** I changed the cap-check line to `current.estimate() >=
maxEvents` (dropping the appended term); both new `EventCapGuardTest` tests
failed. Restored, reran, passed.

### MAJOR 2 (security) — the estimator ran inside the lock

**Fix.** `EventCapGuard.java:79`, `AtomicBoolean refreshInFlight`. `isOverCap()`
(line 165) only the caller that wins `compareAndSet(false, true)` (line 168)
calls the estimator; every other caller returns the previous cached estimate
at once, with no lock and no wait.

**Test.** `EventCapGuardTest.aSecondCallerDuringARefreshReturnsAtOnceWithNoWaitForTheEstimator`:
a `GatedEstimator` signals a latch the moment it is entered; a second call,
strictly ordered after that signal, must return in under 500 ms while the
first call is still blocked inside a 2-second estimator wait.

**Break/restore.** I reverted `isOverCap()` to a `synchronized (refreshInFlight)`
block around the whole method (the pre-fix shape); the new test failed
(the second call waited for the full estimator call). Restored, reran, passed.
I first tried breaking this with an 8-thread call-count test, and found it did
not distinguish the two implementations (the fixed simulated clock makes
`needsRefresh` false for every caller after the first, in both the broken and
the fixed code, so the blocking difference did not show up in that test's
metric); I replaced it with the latch-gated test above, which does.

### MINOR findings

| # | Source | Fix | File : line |
|---|---|---|---|
| 1 | MongoDB MINOR 1 | The pull request text no longer claims "stays out of the way of #157"; it states the merge commit and the four conflicts. | PR text |
| 2 | Security MINOR 2 / MongoDB MINOR 3 | The hour throttle now tracks the elapsed time since the last warning, not the UTC clock hour. | `EventCapGuard.java:250-258` |
| 3 | MongoDB MINOR 4 | `MongoEventLogStore(MongoDatabase, int)` no longer reads `OCTOMETER_MAX_EVENTS`; it uses `DEFAULT_MAX_EVENTS` explicitly. I chose this over threading an explicit cap through every other test file that calls this constructor (write-error, index-error, delete-error, delete-by-user-id, insert-only tests are owned by issues #11/#35/#126, not #34); this removes the actual flakiness risk with a much smaller diff. | `MongoEventLogStore.java:207-219` |
| 4 | MongoDB MINOR 5 | `CapturingLoggerFinder.messages()` (jvm-core) returns `new ArrayDeque<>(MESSAGES)`, a copy. | `CapturingLoggerFinder.java` (jvm-core) |
| 5 | MongoDB MINOR 7 | `append()` now builds the document list (with the per-event null check) before it checks the cap. | `MongoEventLogStore.java:276-289` |
| 6 | MongoDB MINOR 6 | One sentence on the drift of `estimatedDocumentCount()` after an unclean shutdown, in the Javadoc of `estimatedEventCount()`. | `MongoEventLogStore.java:252-259` |
| 7 | Maintainer decision (round 1) | `OCTOMETER_MAX_EVENTS` of zero, negative, or non-numeric text now stops the app start (`IllegalStateException`), instead of silently defaulting. | `MongoEventLogStore.java:544-575` |
| 8 | Security MINOR 1 | A value of `OCTOMETER_MAX_EVENTS` above 1,000,000 is clamped at start, with a warning (mirrors the existing retention-days clamp). | `MongoEventLogStore.java:184-190, 579-586` |
| 9 | Security MINOR 3 | A backward clock step no longer freezes the snapshot: `needsRefresh` also returns `true` when `now < current.checkedAtMillis()`. | `EventCapGuard.java:221` |
| 10 | Security MINOR 4 | Not a code fix, as the reviewer noted. One sentence added to the Javadoc of `append()`: a dropped batch returns faster, with no user data revealed by that. | `MongoEventLogStore.java:265-274` |
| 11 | Security MINOR 5 | The two over-length Javadoc sentences of `EventCapGuard`'s class comment are split into shorter sentences (also applied more broadly across the round-1 diff for ASD-STE100 compliance). | `EventCapGuard.java:13-63` |

**Break/restore for the MINOR fixes with a dedicated new test.** I reverted the
hour-throttle to the UTC-bucket form; `EventCapGuardTest.theHourThrottleUsesElapsedTimeNotTheUtcClockHour`
failed. Restored, reran, passed. I reverted `maxEventsFromValue`'s error throw
to the old default-with-warning; `MongoEventLogStoreMaxEventsTest`'s three new
"stops the app start" tests failed. Restored, reran, passed. I reverted
`mongoErrorCodeOf` (see BLOCKER 2 above; the same change also covers MINOR 6's
sibling assertion). The MINOR-7 reorder is covered by a new test,
`MongoEventLogStoreEventCapTest.aNullEventThrowsEvenWhenTheStoreIsAlreadyOverTheCap`;
I reverted the ordering, the test failed, I restored, reran, passed.

### Test output (round 1, JDK 21)

- `./gradlew :kit:jvm-core:test`: 254 tests, 0 failures, 0 errors.
- `./gradlew :kit:jvm-ktor:test`: 44 tests, 0 failures, 0 errors.
- `./gradlew :kit:jvm-mongo:test` (Docker): 63 tests, 0 failures, 0 errors.
- `./gradlew :kit:jvm-mongo:testMongoSyncNewest` (the newest 5.x driver, Docker,
  run alone): 63 tests, 0 failures, 0 errors.
- `./gradlew build`: BUILD SUCCESSFUL, every module.

Commit `9cd86f3` holds the round-1 fix, on top of merge commit `fd01be1`.

### CI

`gh pr checks 165` (2026-09-22): Tracker pass, JVM modules pass, Monitor
backend on Windows pass, Frontend pass, Secret scan pass. All 5 checks green.

xamcross:

I made a second clone at a new path, I ran `gh pr checkout 165` in it, and I ran each command
myself with JDK 21 and Docker. I checked `9cd86f3` on top of `fd01be1`.

## Each round-0 finding of my review

- **BLOCKER 1 — the fail-open state.** FIXED. `EventCapGuard.java:195-217` keeps the last good
  estimate (line 215), it resets the failure count on a success (line 198), and it returns
  `maxEvents` after 3 failures in a row (line 203-209). `needsRefresh` (line 219-223) also
  handles a backward clock step. Tests:
  `aCountErrorKeepsTheLastGoodEstimateInsteadOfResettingItToZero` and
  `aThirdCountErrorInARowFailsClosedAndASuccessReopensTheGuard`. The state opens again after one
  good count, thus a short network fault never stops the ingest for long.
- **BLOCKER 2 — the undocumented privilege.** FIXED. `kit/jvm-mongo/README.md:28-51` holds a
  section "The event cap". It names the minimum role: `insert`, `createIndex`, `collMod`, and
  `find`. The class comment of `MongoEventLogStore.java:88-94` states the same four actions and
  the reason. `MongoEventLogStoreRestrictedRoleEventCapTest` builds a user with exactly those
  four actions (line 66) and proves that the cap stops the second batch (line 82-107).
  `MongoEventLogStoreInsertOnlyUserTest.java:109-124` asserts the fail-open warning with the
  text "MongoDB error code 13", and it asserts that the server text stays out of the line.
- **MAJOR 1 — the overshoot.** FIXED. `countAppended(long)` (line 189) adds each written batch,
  `isOverCap()` (line 179) adds that total to the cached estimate, and each refresh resets the
  total (line 174). `MongoEventLogStore.append` (line 299) calls it after a successful insert.
  `MongoEventLogStoreEventCapTest.aFloodInsideOneRefreshIntervalStopsAtTheCap` proves that 5
  appends with the cap 1 store 1 document.
- **MAJOR 2 — the lock.** FIXED. `EventCapGuard.java:169` uses
  `refreshInFlight.compareAndSet(false, true)`. A second caller reads the cached snapshot and
  returns at once. Test: `aSecondCallerDuringARefreshReturnsAtOnceWithNoWaitForTheEstimator`.
- **MINOR 1 to 5.** Each one is fixed: the clamp at 1000000 with a WARNING
  (`MongoEventLogStore.java:578-586`), the elapsed-time throttle
  (`EventCapGuard.java:252-259`), the backward clock step (line 222), the timing note
  (`MongoEventLogStore.java:265-273`), and the split Javadoc sentences.

## `OCTOMETER_MAX_EVENTS`

`maxEventsFromValue` (`MongoEventLogStore.java:559-576`) throws an `IllegalStateException` for a
zero, a negative value, and a text, thus the app start stops. The message names the variable,
never the value. A value above 1000000 is clamped with one WARNING. A `null` value still gives
the default 200000 with no warning. This matches the decision of the maintainer.

## The merge with `origin/main`

I compared each of the three files with `origin/main` (`git diff origin/main HEAD`). The three
files together delete 18 lines. Each deleted line is a class declaration that gains an
interface, or a comment that this work replaces. The erasure of #157 stays whole:
`deleteByUserId` returns `DeletionResult`, the pass loop and the bound stay, and the `COLLSCAN`
sentence stays in a new form. The only removed statement of fact is "Issue #34 adds this cap to
the store; the kit does not enforce it yet", which this pull request makes false.

## No user data in a log line

I read each `LOGGER.log` call and each exception message of the two main files. Each one holds a
fixed text plus one of: a numeric MongoDB error code, a count of failures, a configured number,
or the name of a step. No line holds a connection string, a host, a port, a user id, or a
session id. The message of `maxEventsFromValue` names the variable and never repeats the value.

## The test run

One run of `./gradlew :kit:jvm-core:test :kit:jvm-ktor:test :kit:jvm-mongo:test` with JDK 21:
BUILD SUCCESSFUL. jvm-core 254/0/0, jvm-ktor 44/0/0, jvm-mongo 63/0/0, 0 tests skipped. The
numbers match the text of the author.

## One nit, which does not block

`MongoEventLogStore.java:84`: the class comment still says "one warning for each clock hour".
The guard now throttles on the elapsed time. Change the words at the next touch of the file.

No BLOCKER and no MAJOR finding stays. I approve.

xamcross:
**Review (second review, Kotlin and MongoDB engineer): APPROVE**

I made a new clone, I ran `gh pr checkout 165`, and I checked commit `9cd86f3` on
top of the merge `fd01be1`. I ran each command myself on JDK 21 with Docker.

## The count query (my MAJOR 1, round 0)

`kit/jvm-mongo/README.md:38-46` now states the minimum role: a custom role on
`octometer_events` with `insert`, `createIndex`, `collMod`, and `find`. It states
that `find` lets the store run `estimatedDocumentCount()`, and that the cap never
stops the ingest without it. The class comment of `MongoEventLogStore` says the
same. `mongoErrorCodeOf` (`MongoEventLogStore.java:246`) puts the numeric MongoDB
error code in each warning. `MongoEventLogStoreInsertOnlyUserTest.java:113-123`
now asserts the second warning with the text `MongoDB error code 13`, and
`MongoEventLogStoreRestrictedRoleEventCapTest.java:66` proves the cap with a user
that holds exactly `insert`, `createIndex`, `collMod`, `find`. My MAJOR 1 is
closed.

## The refresh runs outside a lock

`EventCapGuard.java:165-180`. Only the caller that wins
`refreshInFlight.compareAndSet(false, true)` calls the estimator, and that call
sits outside every lock. A caller that loses the flag skips the block, and it
returns the cached estimate at once.

**My break and restore.** I wrapped the whole body of `isOverCap()` in
`synchronized (refreshInFlight)`, the shape before the fix.
`EventCapGuardTest.aSecondCallerDuringARefreshReturnsAtOnceWithNoWaitForTheEstimator`
failed with "The second call must not wait for the estimator; it took 2003 ms".
I restored the file and the suite went green again. The test is real, and my
MINOR 2 is closed.

## The bound inside the 60-second window

`EventCapGuard.java:84, 174, 179, 188`. `countAppended(long)` adds each accepted
batch to `appendedSinceRefresh`. `isOverCap()` adds that total to the cached
estimate. Each refresh sets the total back to zero.

- `MongoEventLogStoreEventCapTest.aFloodInsideOneRefreshIntervalStopsAtTheCap`:
  5 appends inside one window with a cap of 1 leave 1 document. It passed in my
  own run of the 63 tests.
- **My break and restore.** I dropped the `appendedSinceRefresh.get()` term from
  the cap check. `countAppendedAddsToTheCachedEstimateInsideOneRefreshInterval`
  and `aRefreshResetsTheAppendedCounter` both failed. I restored the file and both
  passed.

## The order inside `append`

`MongoEventLogStore.java:276-299`. The method builds the document list first, and
that loop calls `Objects.requireNonNull(event, ...)` for each event. The cap check
runs second. `countAppended(documents.size())` sits after the whole `try`/`catch`
block, thus a failed `insertMany` throws `EventLogWriteException` and never reaches
that line. Both points are correct.
`MongoEventLogStoreEventCapTest.aNullEventThrowsEvenWhenTheStoreIsAlreadyOverTheCap`
covers the order.

## My MINOR findings of round 0

|---|---|---|---|
| 1 | The pull request text claimed no conflict with #157 | FIXED | The section "The merge with `origin/main`" names the merge `fd01be1` and the four conflicts. |
| 2 | The refresh blocked each concurrent caller | FIXED | `EventCapGuard.java:168`, and my break and restore above (2003 ms). |
| 3 | The throttle used the UTC clock hour | FIXED | `EventCapGuard.java:250-258` uses the elapsed time. Test `theHourThrottleUsesElapsedTimeNotTheUtcClockHour`. |
| 4 | The two-argument constructor read the process environment | FIXED | `MongoEventLogStore.java:218` passes `DEFAULT_MAX_EVENTS`. I accept this choice over an explicit cap in each test file; it removes the same risk with a smaller diff. |
| 5 | `CapturingLoggerFinder.messages()` handed out the live deque | FIXED | It returns a copy. The class comment states it. |
| 6 | The fast count can drift after an unclean shutdown | FIXED | One sentence in the comment of `estimatedEventCount()`, `MongoEventLogStore.java:252-258`. |
| 7 | The cap check ran before the per-event null check | FIXED | `MongoEventLogStore.java:281-289`, plus the new test. |

## The test counts, from my own runs (JDK 21, Docker)

| Task | Tests | Skipped | Failures | Errors |
|---|---|---|---|---|
| `:kit:jvm-mongo:test` (driver 5.0.1) | 63 | 0 | 0 | 0 |
| `:kit:jvm-mongo:testMongoSyncNewest` (the newest 5.x driver, run alone) | 63 | 0 | 0 | 0 |

Both numbers agree with the pull request text.

## One observation, no new finding

The new rule fails closed after three count errors in a row. This reverses the
fail-open rule of round 0. An app with the old role, and thus with no `find`,
now stops the whole ingest about two minutes after the start. The README states
this plainly, and each warning names the MongoDB error code, thus an operator
sees the cause. The security review asked for the rule. I accept it.

I have no BLOCKER finding and no MAJOR finding left. I approve.

Issue #150, opened 2026-09-22
Stop the TRACE line of StatusPages that holds the query string


xamcross:
## Goal

No log line of the monitor holds a user id, a session id, or a query string, at each log level.

## Observed on 2026-09-22

Found in the correction round and the second privacy review of #147 (issue #61): https://github.com/xamcross/octometer/pull/147#issuecomment-5770065012

- The plugin `io.ktor.server.plugins.statuspages.StatusPages` writes the line "No handler found for status code 404 Not Found for call: /api/apps/777777/events?userId=<value>" at the TRACE level for each unhandled status. The line holds the full path and the query string.
- The default config `monitor/backend/src/main/resources/logback.xml` sets the root level to INFO, thus the line does not print today. A person who sets the level to TRACE for a diagnosis prints each user id of the routes of level 2, level 3, and the erasure (design decision D13 puts the user id in the query string or in the path).
- The log sentinel test of #147 filters that line out. Each sentinel test asserts on each level (lesson 9 of the backend brief).

## Implementation steps

1. Set the level of the logger `io.ktor.server.plugins.statuspages.StatusPages` to OFF in `logback.xml`, or install a handler for each status that the routes give (400, 403, 404, 415), thus the plugin finds a handler and writes no line. Confirm the logger name with a real run at TRACE before you use it.
2. Remove the filter of the sentinel test in `monitor/backend/src/test/kotlin/octometer/monitor/erasure/` (the test of #147), thus the sentinel reads each line at each level.
3. Add one test that sets the root level to TRACE for its scope, sends a request with a marker in the query string to an unknown route and to a route with a wrong parameter, and asserts that no line holds the marker.
4. Search each other logger of Ktor at TRACE for a line with the path or the query string (the routing trace of `io.ktor.server.routing`), and treat it in the same way.

## Acceptance criteria

- [x] At the TRACE level, a 404 and a 400 with a marker in the query string write no line with the marker.
- [x] The sentinel test of the erasure route has no filter.
- [x] `./gradlew :monitor:backend:test` passes with JDK 21, two runs in sequence.
- [x] A comment on this issue records the test output.

## Related

- #61 found it (pull request #147). #57 (the file log) must keep this rule.
- #148 adds a handler for `BadRequestException`; do #148 first, because both change the `StatusPages` install in `Application.kt`.

xamcross:
## Test output

[...]

PR #166, opened 2026-09-22
fix(monitor): give 400 for a bad percent escape in a query string

### LINKED ISSUE #148: Answer 400, not 500, for a bad percent escape in a query string
## Goal

A request with a bad percent escape in its query string gets 400 with a fixed sentence, on each route of the monitor.

## Current behaviour

Found on 2026-09-22 in the security review of #144 (https://github.com/xamcross/octometer/pull/144#issuecomment-5769526769), on `main` at commit 3186569:

- `GET /api/health?x=%zz` on a real Netty server answers `500 {"error":"The server had an internal error."}` and writes one ERROR log line. The cause is an `io.ktor.server.plugins.BadRequestException` from the query parser of Ktor.
- `monitor/backend/src/main/kotlin/octometer/monitor/Application.kt:87-88` installs `StatusPages` with one handler, `exception<Throwable>`. That handler gives 500 to each throw, also to a throw that means "the request is invalid".
- The registry routes catch `BadRequestException` and `ContentTransformationException` for a request body (`registry/AppRegistryRoutes.kt:106-110`). No code catches the same exception for a query string, thus each route with a query parameter (#50, #51, #61, #113, #114) has the defect.

## Implementation steps

1. In `Application.kt`, add `exception<BadRequestException>` before the catch-all handler. It answers 400 with the fixed body of `Errors.kt` for an invalid request, and it writes no ERROR line.
2. Check that the handler holds no text of the request: the message of `BadRequestException` can hold the raw query string. Log the class name only.
3. Read `registry/AppRegistryRoutes.kt:95-112`. Keep the route catches when they give a different sentence; else remove the duplicate.
4. Add a test with a real server (the pattern of `RequestGuardRawSocketTest`): `GET /api/health?x=%zz` gives 400, the body is the fixed sentence, and the log holds no ERROR line and no `%zz`.
5. Add the same test for one route with a query parameter, when one is on `main` at that time.

## Acceptance criteria

- [x] `GET /api/health?x=%zz` gives 400 with the fixed body, and the log holds no ERROR line.
- [x] No body and no log line holds a part of the query string.
- [x] A throw of the app code still gives the fixed 500 body (the catch-all stays).
- [x] `./gradlew :monitor:backend:test` passes with JDK 21, two runs in sequence.

## Related

- #51 (pull request #144) found it. The route of #51 does not change `Application.kt`; this issue owns the fix.
- #15 closed on 2026-09-22. It added the body catches of the registry routes.
- Start this issue after the open backend pull requests merge (#127, #144, #145, #147, and the ones of #38, #55, #59), because each one changes `monitor/backend`.

xamcross:
## Summary

A bad percent escape in a query string, for example `?x=%zz`, threw `BadRequestException` from the Ktor query parser. `RequestGuard` checks only the path, so the exception reached the catch-all `StatusPages` handler and gave 500. This change adds a handler for `BadRequestException` before the catch-all. It gives 400 with the fixed error body, and it logs the exception class name only, never the message.

## Acceptance criteria

- [x] `GET /api/health?x=%zz` gives 400 with the fixed body, and the log holds no ERROR line.
  Evidence: `BadQueryEscapeTest.a bad percent escape in the query string of health gives 400, with the fixed body and no ERROR line` sends the request over a raw socket and asserts the status line, the body `{"error":"The request is not valid."}`, and the absence of an ERROR event.
- [x] No body and no log line holds a part of the query string.
  Evidence: the same test asserts the response and every captured log event (each level, `captureLogEvents`) do not contain `%zz`. The handler logs `cause.javaClass.simpleName` only.
- [x] A throw of the app code still gives the fixed 500 body (the catch-all stays).
  Evidence: `BadQueryEscapeTest.a plain app failure still gives the fixed 500 body` throws `IllegalStateException` from a probe route and asserts 500 with `{"error":"The server had an internal error."}`. `ApplicationErrorHandlingTest` (unchanged) still passes too.
- [x] `./gradlew :monitor:backend:test` passes with JDK 21, two runs in sequence.
  Evidence: both runs gave `BUILD SUCCESSFUL`, 319 tests, 0 failures, 0 errors.
  Evidence: posted after this pull request opens.

## Extra evidence

- `BadQueryEscapeTest.a bad percent escape in the query string of a users route gives 400 too` proves the same fix on `GET /api/apps/1/users?q=%zz`, a route beyond health with its own query parameter.
- I broke the code for each rule, in turn, and saw the matching test fail, then restored the code: an ERROR log line, a leak of the raw query string in the log, a 500 status instead of 400, and a removed catch-all handler. Each break made the matching assertion fail.

## Decisions where the issue text left a choice

- The fixed 400 sentence is `The request is not valid.` The issue names no exact wording, only "a fixed sentence".
- `AppRegistryRoutes.kt:95-112` keeps its three catches of `receiveOrNull`. They give the sentence `The body is not a valid app request.`, for a malformed request body. My new handler answers a different case, a bad query string, with a different sentence, so the two are not duplicates.
- I put the fix in `Application.kt` only, with no route-level change, because `RequestGuard` checks the path, not the query string, and the query parser runs inside the route pipeline on the first read of `queryParameters`.

Closes #148

Generated with Claude Code


xamcross:
**Review (Ktor and security engineer): CHANGES REQUESTED**

I made my own clone. I ran `gh pr checkout 166` in it, at commit `fb449be`. I used JDK 21
(`ms-21.0.9`). Each result below comes from a real Netty server on a free loopback port. A raw
socket sends each request. A Logback `ListAppender` on the root logger at TRACE reads each log
line. I deleted each probe after the run. The tree is clean again.

## The acceptance criteria

- [x] **`GET /api/health?x=%zz` gives 400 with the fixed body, and the log holds no ERROR line.**
  PASS. The status line is `HTTP/1.1 400 Bad Request`. The body is
  `{"error":"The request is not valid."}`. The ERROR count is 0. The answer also keeps
  `Cache-Control: no-store`, `X-Content-Type-Options: nosniff`, and the CSP header of D12.
- [x] **No body and no log line holds a part of the query string.**
  PASS at the production log level. I sent 14 requests: a bad escape in the query string of
  health, of `/api/apps/1/users`, of `/api/apps/1/elements`, and of `DELETE
  /api/apps/1/events`; a bad escape in the path (`/api/%zz` and `/api/apps/%zz/users`); a bad
  escape in a header value; a POST body with a bad escape; a malformed POST body; a POST with
  no body. No answer body holds `%zz`. The monitor writes one line only:
  `INFO [octometer.monitor.Application] The query string is not valid. BadRequestException`.
  **One fact for the record:** at TRACE the plugin logger
  `io.ktor.server.plugins.statuspages.StatusPages` writes
  `Call /api/health?x=%zz failed with cause ...`. The root level of `logback.xml` is INFO, so
  that line never runs in production. Issue #150 owns that exact line, and it says to do #148
  first. This does not block #166.
- [x] **A throw of the app code still gives the fixed 500 body (the catch-all stays).**
  PASS. A probe route that throws `IllegalStateException("probe-marker-app-72f1")` gives
  `HTTP/1.1 500`, the body `{"error":"The server had an internal error."}`, and one ERROR line
  `An unhandled exception reached the server. IllegalStateException`. No answer holds the
  marker text.
- [x] **`./gradlew :monitor:backend:test` passes with JDK 21, two runs in sequence.**
  PASS. Run 1: `BUILD SUCCESSFUL in 1m 18s`. Run 2 (`--rerun-tasks`): `BUILD SUCCESSFUL in
  1m 42s`. The JUnit XML files give 319 tests, 0 failures, 0 errors, 0 skipped. That matches
  your count.
- [x] **A comment on the issue records the test output.** PASS. Comment 5774499216 on #148.

## Findings

### MAJOR 1 — the log assertion cannot fail for the regression that the issue names

`monitor/backend/src/test/kotlin/octometer/monitor/BadQueryEscapeTest.kt:39-47` and `:70-78`.

Step 2 of issue #148 says: the message of `BadRequestException` can hold the raw request text,
so log the class name only. The new test claims to lock that rule. It cannot.

Evidence. I changed `Application.kt:105` from `cause.javaClass.simpleName` to `cause.message`.
I ran `./gradlew :monitor:backend:test --tests "octometer.monitor.BadQueryEscapeTest"`. The
result was `BUILD SUCCESSFUL`. No assertion failed.

The cause: for a **query** decode failure the message of Ktor is the fixed text `Unable to
decode query parameters`. It holds no part of the query string. For a **path** decode failure
the message is `Url decode failed for /api/apps/1/users/%zz`. It holds the raw request target.
The test sends a bad escape in the query string only, so the leak stays invisible.

I proved the leak on a real server, with `cause.message` in the handler and the root level at
INFO:

```
status=HTTP/1.1 400 Bad Request
INFO  The query string is not valid. Url decode failed for /api/apps/1/users/%zz
```

That line goes to the production log today, because the root level is INFO. Lesson 1 and
lesson 3 of the backend brief both apply.

**The fix.** Add a third case to `BadQueryEscapeTest` that sends `GET /api/apps/1/users/%zz`,
with the bad escape in the path and not in the query string. Assert 400, the fixed body, no
ERROR line, and `events.none { it.formattedMessage.contains(BAD_ESCAPE) }`, the same four
assertions as the two cases above. Then change `cause.javaClass.simpleName` to `cause.message`
one time, see the new case fail, and restore the code.

### MINOR 1 — the log sentence names the wrong part of the request

`monitor/backend/src/main/kotlin/octometer/monitor/Application.kt:105`.

The handler answers every `BadRequestException`, not a query-string failure only. Evidence:
`GET /api/%zz` and `GET /api/apps/%zz/users` each give the line `The query string is not valid.
BadRequestException`, while the Ktor cause is `Url decode failed for ...`. The sentence sends a
person who reads the log to the wrong place.

Fix: write `log.info("The request is not valid. {}", cause.javaClass.simpleName)`.

### MINOR 2 — two sentences of the comment block are not correct

`monitor/backend/src/main/kotlin/octometer/monitor/Application.kt:100-103`.

(a) The comment says "This handler goes before the catch-all, so a request mistake gives 400,
not 500." The position is not the reason. Ktor 3.6.0 picks a handler with
`selectNearestParentClass` in
`io.ktor.server.plugins.statuspages.StatusPagesUtilsJvmKt`. Evidence: I put the catch-all
first and ran `BadQueryEscapeTest`. The result was `BUILD SUCCESSFUL`. Keep the order, because
it helps a reader, but correct the sentence.

(b) The comment says "The message of `BadRequestException` can hold the raw query string". The
message holds the raw **target** of a path failure, and a fixed text for a query failure. See
MAJOR 1.

Fix: write "Ktor picks the nearest parent class, so this handler wins over the catch-all. The
message of BadRequestException can hold the raw request target. The log line names the
exception class only."

### MINOR 3 — the raw-socket helper is a copy

`monitor/backend/src/test/kotlin/octometer/monitor/BadQueryEscapeTest.kt:115-150` repeats
`withRawSocketServer` and `sendRawRequest` of
`monitor/backend/src/test/kotlin/octometer/monitor/security/RequestGuardRawSocketTest.kt:88-116`.
Two private copies with the same names now exist.

Fix: put the two functions in one test-support file beside `TestConfigs.kt`. Give the probe
routes as a parameter with an empty default.

### MINOR 4 — the two query tests repeat five assertions

`monitor/backend/src/test/kotlin/octometer/monitor/BadQueryEscapeTest.kt:26-96`.

Fix: extract one private function `assertFixedBadRequest(response, events)`. Each case then
holds the request and one call.

### MINOR 5 — two comment sentences are too long

`BadQueryEscapeTest.kt:16-19` holds a sentence of 30 words. `BadQueryEscapeTest.kt:114-118`
holds a sentence of 36 words. The brief sets a maximum of 25 words for a comment sentence, and
one instruction for one sentence.

Fix: split each sentence.

### MINOR 6 — the test builds its own config

`BadQueryEscapeTest.kt:120-126` sets `retentionDays = 395`, and not the large value of
`TestConfigs.kt`. The purge of `MonitorServices.open` runs on the real clock. The new test
holds no row, so nothing fails today.

Fix (optional): write `prodConfig(dataDir = testDataDir()).copy(port = port)`.

## The handler for BadRequestException (the review question)

I read `ktor-server-core-jvm-3.6.0.jar` with `javap`. Three classes of
`io.ktor.server.plugins` reach the new handler:

| The class | The source | Is 400 right? |
| --- | --- | --- |
| `BadRequestException` | the URL decoder: `Unable to decode query parameters`, `Url decode failed for <target>` | Yes. The client sent a target the server cannot decode. |
| `ParameterConversionException` | `call.parameters.getOrFail<T>()` with a value of the wrong type | Yes. The client sent a value of the wrong type. |
| `MissingRequestParameterException` | `getOrFail` with a missing name | Yes. The client left out a required name. |

The last two do not run today. Each route reads `call.parameters["id"]?.toLongOrNull()` and
answers with its own sentence.

A `receive` of a wrong type does **not** reach the handler.
`io.ktor.server.plugins.ContentTransformationException` extends `java.io.IOException`, and
`io.ktor.serialization.ContentConvertException` extends `java.lang.Exception`. Evidence: a
probe route that throws `CannotTransformContentToTypeException` still gives 500 and one ERROR
line. This is correct for #166, because `receiveOrNull`
(`registry/AppRegistryRoutes.kt:103-112`) catches both at the one `call.receive` site of the
module. My probe confirms it: a POST with no body gives 400
`{"error":"The body is not a valid app request."}`. A later route that calls `call.receive`
without that helper gets 500, against lesson 5 of the brief. That belongs in a new issue, not
in #166.

**Did the change hide a real server defect?** No. No file under `monitor/` or `kit/` throws
`BadRequestException`, so no true 500 becomes a 400 today. One risk stays for later: a probe
route that throws `BadRequestException` now gives 400, one INFO line, and no stack trace. App
code that throws that class in the future hides a server defect. Please add one sentence to the
comment block that tells a later author to throw a different class for a server failure.

## The order of the handlers

The order in the file is right: the specific handler stands before the catch-all. The rule of
lesson 2 stays as it is at `Application.kt:109`:
`if (cause is CancellationException && !call.isActive) throw cause`. The diff does not touch
it. `ApplicationErrorHandlingTest` still covers it; I removed the catch-all and both of its
tests failed.

## The body catches of the registry routes

Keep them, and keep the two sentences apart. `The body is not a valid app request.` tells the
client which part of the request is wrong. `The request is not valid.` does not. A merge would
lose that help, and it would change the answer of three routes that #148 does not own.

`Errors.kt` holds the `ErrorBody` class only. It holds no fixed sentence. The 500 sentence sits
inline at `Application.kt:111`, and `SECRET_STORE_UNAVAILABLE_MESSAGE` sits at
`AppRegistryRoutes.kt:20`. The repository keeps each sentence beside its use, so the new
constant is in the right file. Do not move it. You can make the 500 sentence a named constant
too, for symmetry.

## The broken-code runs

| The break | The result |
| --- | --- |
| The `BadRequestException` handler is removed | 2 of 3 tests of `BadQueryEscapeTest` FAILED |
| The handler logs `cause.message` | **BUILD SUCCESSFUL — see MAJOR 1** |
| The handler logs at ERROR | 2 of 3 tests of `BadQueryEscapeTest` FAILED |
| The catch-all is removed | 3 tests FAILED (1 of `BadQueryEscapeTest`, 2 of `ApplicationErrorHandlingTest`) |
| The two handlers change places | BUILD SUCCESSFUL — the order does not select the handler |

I restored the code after each break. The final tree matches `fb449be` exactly.

## The merge test

- `git merge --no-commit --no-ff origin/main` (`1f7f436`): `Automatic merge went well`. I
  aborted it.
- `git fetch origin pull/160/head:pr160`, then `git merge --no-commit --no-ff pr160`:
  `Automatic merge went well`. I aborted it.
- Pull request #160 changes no file of #166. It adds `monitor/backend/.../mongo/MongoAppReader.kt`
  and two test files, and it changes `.github/workflows/ci.yml`, `gradle/libs.versions.toml`,
  `monitor/backend/build.gradle.kts`, and `monitor/backend/src/main/resources/logback.xml`
  (a new `org.mongodb` logger at WARN). It changes neither `Application.kt` nor
  `MonitorServices.kt`.

## The scope and the diff

The diff holds two files, 165 added lines, 0 deleted lines. It holds no secret, no database
file, and no build output. The change does what #148 asks, and no more.

Close MAJOR 1. The MINOR findings are your choice.

xamcross:
Correction round 1, commit f39d7e8, pushed after a merge of `origin/main` (fb449be).

- MAJOR 1: added a test case for a bad percent escape in the path of `/api/apps/1/users/%zz`. I changed `cause.javaClass.simpleName` to `cause.message`, ran the test, and saw only the new case fail (the two query-string cases still passed, as the review found). I restored the code. Closed.
- MINOR 1: the log sentence now reads `The request is not valid.`. Closed.
- MINOR 2: the comment block now says Ktor picks the nearest parent class, and names the raw request target, not the raw query string. Closed.
- MINOR 3: `withRawSocketServer` and `sendRawRequest` now live in one shared file, `RawSocketServer.kt`. `RequestGuardRawSocketTest` and `BadQueryEscapeTest` both use it. Closed.
- MINOR 4: extracted `assertFixedBadRequest`, one function for the five assertions of a 400 case. Closed.
- MINOR 5: split each comment sentence above 25 words. Closed.
- MINOR 6: the raw-socket helper now builds its config with `prodConfig(dataDir = testDataDir()).copy(port = port)`. Closed.

`./gradlew :monitor:backend:test`, JDK 21, two runs in sequence: run 1 `BUILD SUCCESSFUL in 1m 20s`, run 2 (`--rerun`) `BUILD SUCCESSFUL in 1m 38s`. 320 tests (319 + the new MAJOR 1 case), 0 failures, 0 errors, in each run.

CI on this push: all 5 checks pass (Frontend, JVM modules, Monitor backend on Windows, Secret scan, Tracker).

xamcross:
**Review (second review, Ktor and security engineer): APPROVE**

I made a new clone. I ran `gh pr checkout 166` in it, at commit `f39d7e8`. I used JDK 21
(`ms-21.0.9`). I deleted the clone after the run.

## MAJOR 1 — closed, with my own evidence

I changed `Application.kt:109` from `cause.javaClass.simpleName` to `cause.message`. I ran
`./gradlew :monitor:backend:test --tests "octometer.monitor.BadQueryEscapeTest"`:

```
BadQueryEscapeTest > a bad percent escape in the path of a users route gives 400 too, with no leak of the escape() FAILED
4 tests completed, 1 failed
BUILD FAILED
```

The new case fails. The two query-string cases still pass, so the new case is the one that
catches the leak. I restored the code, and the tree matches `f39d7e8` exactly.

## The 6 MINOR findings

| The finding | The state | The evidence |
| --- | --- | --- |
| MINOR 1 — the log sentence names the wrong part | Closed | `Application.kt:109` now reads `log.info("The request is not valid. {}", cause.javaClass.simpleName)`. |
| MINOR 2 — two wrong comment sentences | Closed | `Application.kt:97-107` now says Ktor picks the nearest parent class, and it names the raw request target. It also adds the sentence for a later author: app code must throw a different class for a server failure. |
| MINOR 3 — the raw-socket helper is a copy | Closed | The new file `monitor/backend/src/test/kotlin/octometer/monitor/RawSocketServer.kt` holds one copy. `RequestGuardRawSocketTest.kt:3` imports it. No private copy of `withRawSocketServer` stays in the module. |
| MINOR 4 — the repeated assertions | Closed | `BadQueryEscapeTest.kt:104-121` holds `assertFixedBadRequest`. Each of the three 400 cases makes one call. |
| MINOR 5 — the long comment sentences | Closed | The longest sentence of `BadQueryEscapeTest.kt` and of `RawSocketServer.kt` now holds 24 words. |
| MINOR 6 — the hand-built test config | Closed | `RawSocketServer.kt:32` reads `prodConfig(dataDir = testDataDir()).copy(port = port)`. |

## The test run

`./gradlew :monitor:backend:test --rerun-tasks`, JDK 21: `BUILD SUCCESSFUL in 1m 34s`. The
JUnit XML files give 320 tests, 0 failures, 0 errors, 0 skipped. That matches your count.

## The result

No BLOCKER and no MAJOR finding stays. The change is ready to merge.

PR #167, opened 2026-09-22
fix(monitor): remove a leftover apps-*.json.tmp file at the start

### LINKED ISSUE #141: Remove a leftover apps-*.json.tmp file at the start of the monitor
## Goal

No leftover temporary file with connection strings stays in the `secrets` folder after a kill of the monitor.

## Current behaviour

Found on 2026-09-22 in the third and the fourth security review of #125 (https://github.com/xamcross/octometer/pull/125#issuecomment-5768598361 and https://github.com/xamcross/octometer/pull/125#issuecomment-5769037345):

- `monitor/backend/.../SecretStore.kt:143` writes each connection string of the registry to a temporary file `apps-<random>.json.tmp`, then it moves that file over `apps.json`.
- A `finally` block removes the temporary file on a normal path. A kill of the process between the write and the move leaves the file in the `secrets` folder, and no code removes it later.
- The severity is MINOR: the leftover file holds the same plaintext as `apps.json`, in the same folder, under the same rights. It does not appear on a normal path.

## Implementation steps

1. In the start sweep of `MonitorServices.open`, list each file of the `secrets` folder whose name matches `apps-*.json.tmp`.
2. Delete each such file, and write one log line with the count. The line holds no file content.
3. Keep the rule of #15: a failed delete gives a warning, and it does not stop the start.
4. Add the tests of the acceptance criteria.

## Acceptance criteria

- [x] A file `apps-123.json.tmp` in the `secrets` folder is gone after the start, and `apps.json` stays unchanged byte for byte.
- [x] A file with a different name in the `secrets` folder stays.
- [x] A locked temporary file gives one warning, and the monitor starts.
- [x] No log line holds a part of a connection string.
- [x] `./gradlew :monitor:backend:test` passes with JDK 21.

## Related

- #15 made the secret store (#125). This issue starts after #125 merges.
- #16 reads the connection strings. It does not change this file.

xamcross:
## Summary

A kill of the monitor process between the temporary write and the atomic
move can leave a leftover `apps-<random>.json.tmp` file in the secrets
folder. `SecretStore.removeLeftoverTempFiles()` now runs at the start of
`MonitorServices.open`, before the first write, and removes each such
file. A failed delete gives one warning with the count only, and it does
not stop the start.

Closes #141

## Acceptance criteria

- [x] A file `apps-123.json.tmp` in the `secrets` folder is gone after the start, and `apps.json` stays unchanged byte for byte.
  Evidence: `SecretStoreTest."removeLeftoverTempFiles removes a matching leftover file, and apps json stays unchanged byte for byte"`, and `MonitorServicesTest."open removes a leftover apps-123_json_tmp file from the secrets folder, and leaves apps json unchanged"`.
- [x] A file with a different name in the `secrets` folder stays.
  Evidence: `SecretStoreTest."removeLeftoverTempFiles keeps a file with a different name"`, and the same `MonitorServicesTest` test (it keeps an `apps-backup.json` file).
- [x] A locked temporary file gives one warning, and the monitor starts.
  Evidence: `SecretStoreTest."removeLeftoverTempFiles gives one warning and leaves the file, when a lock blocks the delete"`, and `MonitorServicesTest."open starts, and gives one warning with no connection string, when a leftover temporary file is locked"`. Each test locks the file with a real `FileChannel` lock, guarded with `Assumptions.assumeTrue` for Windows, and it ran for real on this Windows machine.
- [x] No log line holds a part of a connection string.
  Evidence: the two lock tests above capture every log level with `captureLogEvents`, and assert that no event text holds the test connection string.
- [x] `./gradlew :monitor:backend:test` passes with JDK 21.
  Evidence: two full runs in sequence, each `BUILD SUCCESSFUL`, 322 tests total, 0 failures, 0 errors.
  Evidence: added as a separate issue comment.

## Test count

322 tests, 0 failures, 0 errors, across the whole `monitor/backend` module, on two runs in sequence.

## Decisions where the issue was not clear

1. The delete logic and its own warning line live in `SecretStore.removeLeftoverTempFiles()`, not in `MonitorServices.kt`, to keep the change small and inside the secret store, the same pattern as `removeOrphans()`. `MonitorServices.open` calls it once and logs the removed count, the same pattern as the existing orphan sweep.
2. The sweep runs right after `SecretStore` construction, before the retention purge job starts and before the orphan sweep, so it runs before the first possible write of the secret store.

Generated with Claude Code


xamcross:
**Review (Kotlin engineer): APPROVE**

I made my own clone. I ran `gh pr checkout 167`. I ran each check with JDK 21 on Windows 11.
No BLOCKER and no MAJOR finding stays. Each of the 6 findings below is MINOR.

## The acceptance criteria

1. **PASS** — A file `apps-123.json.tmp` goes at the start. `apps.json` stays unchanged byte for
   byte. My own probe wrote the test connection string of `allowlistedSrvUri()` into
   `secrets/apps.json` and the same text into `secrets/apps-987.json.tmp`. After
   `MonitorServices.open` the leftover file was gone. The two tests of the author also pass in my
   clone.
2. **PASS** — A file with a different name stays. My own probe put five other names in the folder.
   `apps.json.tmp`, `apps-1.json.tmpx` and `Apps-2.json.tmp` each stayed. See MINOR 2 and MINOR 3.
3. **PASS** — A locked temporary file gives one warning. The monitor starts. The two lock tests ran
   for real on this Windows machine with a real `FileChannel` lock. My two runs report
   `skipped 0`, thus `Assumptions.assumeTrue` never skipped them.
4. **PASS** — No log line at any level holds a part of a connection string. My own probe put the
   marker in `apps.json` and in the leftover file, then it captured 49 events at TRACE level. No
   event holds the URI, the user, the host, the app id, the temporary file name, or the folder
   name. The orphan sweep also ran in that probe, and it rewrote `apps.json`.
5. **PASS** — `./gradlew :monitor:backend:test` passes with JDK 21. Two runs in sequence give
   **322 tests, 0 failures, 0 errors, 0 skipped**. Note for a later run: a plain second run reports
   `UP-TO-DATE` and runs no test. I used `cleanTest` before the second run, thus both runs are real.
6. **PASS** — Issue #141 holds a comment with the test output.

`./gradlew build` also passes one time in my clone.

## The race between the sweep and a write

**PASS.** `MonitorServices.kt:80` calls the sweep. `MonitorServices.kt:85` starts the orphan sweep,
which is the first possible write. `Application.kt:77` calls `open` before the module installs a
route, thus no request can write earlier.

`SecretStore.removeLeftoverTempFiles()` takes the same `Mutex` as `put`, `remove` and
`removeOrphans`, and `writeAll` runs only inside that lock. A write of this process can never hold
a live temporary file while the sweep lists the folder.

My own probe ran 8 `put` calls and 8 sweeps at the same time in one coroutine scope. Each sweep
returned 0. Each of the 8 entries survived. The folder held `apps.json` only.

## The pattern

The write path is `File.createTempFile("apps-", ".json.tmp", secretsDir)` at `SecretStore.kt:187`.
It always inserts a non-empty random number. My own probe gives the behaviour of the filter:

| The name in the `secrets` folder | The result |
| --- | --- |
| `apps-123.json.tmp` | removed |
| `apps.json.tmp` (a different program) | stays |
| `apps-1.json.tmpx` | stays |
| `Apps-2.json.tmp` (a different case) | stays |
| `apps-.json.tmp` | **removed**, see MINOR 3 |
| a folder `apps-emptydir.json.tmp`, empty | **removed**, and counted, see MINOR 1 |
| a folder `apps-fulldir.json.tmp`, not empty | stays, and gives the warning, see MINOR 1 |

The sweep makes no `secrets` folder when none is there. A test asserts it, and I saw it pass.

## The failure path

A failed delete gives one warning with the count only. The line holds no path and no file name. The
start goes on. I saw this on a real Windows lock, in the two lock tests and in my own probe:
`The start could not remove 1 leftover temporary secret file(s).`

## The break tests

I broke the code 5 times in my clone, and I ran the 6 new tests each time. I restored each file
after each break. Each break makes the right test fail, thus no new test is always green.

| The break | The tests that failed |
| --- | --- |
| B1 the filter never matches | the 2 remove tests, and the 2 lock tests |
| B2 `open` does not call the sweep | the 2 `MonitorServicesTest` tests only |
| B3 no warning on a failed delete | the 2 lock tests |
| B4 the filter matches every file | the 2 remove tests, and `keeps a file with a different name` |
| B5 a log line holds the test connection string | the 2 lock tests |

B3 also proves that `assertTrue(warnings.isNotEmpty())` at `MonitorServicesTest.kt:109` can fail.
No other component writes a WARN line during that `open` call.

## The merges

- `git merge --no-commit --no-ff origin/main` (`1f7f436`): clean, no conflict. I aborted it.
- `git merge --no-commit --no-ff pr160` (`a8657fb`): clean, no conflict. I aborted it.

A fact for the maintainer: the head of pull request #160 does **not** change `MonitorServices.kt`.
It adds `MongoAppReader.kt`, and it changes `build.gradle.kts`, `ci.yml`, `libs.versions.toml` and
`logback.xml`. The two branches share no file.

## The scope

The pull request changes 4 files: 2 source files and 2 test files. It holds no secret, no database
file, and no build output.

## The findings

**MINOR 1 — `SecretStore.kt:120`: the filter accepts a folder.**
`listFiles { file -> ... }` binds to `FileFilter`, thus it gives a folder too. My probe shows two
results. An empty folder `apps-emptydir.json.tmp` goes, and the count reports it as a removed file.
A folder `apps-fulldir.json.tmp` with a file inside stays, and it gives the warning at each start,
for ever, with no path in the line. An operator cannot find the cause from the log.
*The fix:* add the file test to the filter:
`file.isFile && file.name.startsWith(...) && file.name.endsWith(...)`.

**MINOR 2 — `SecretStore.kt:126`: the delete has no retry.**
The class retries a read (`readTextWithRetry`, line 159) and a move (`moveWithRetry`, line 202)
`MAX_MOVE_ATTEMPTS` times with a pause of `MOVE_RETRY_PAUSE_MILLIS`. The comment at line 196 names
the reason: a backup tool or an antivirus scan holds the file for a short time on Windows. The
sweep tries the delete one time. A lock of 50 ms then leaves a plaintext secret file on the disk
until the next start.
*The fix:* wrap `leftoverFile.delete()` in the same retry loop, with the two constants that the
class already has.

**MINOR 3 — `SecretStore.kt:121`: the filter is broader than the write path.**
`createTempFile` never writes the name `apps-.json.tmp`, because the random part is never empty. My
probe shows that the sweep deletes a file with that exact name. A file of a different program in
the `secrets` folder thus goes. The risk is small, because D34 gives the folder to the monitor.
*The fix (optional):* require one character between the prefix and the suffix, for example
`file.name.length > LEFTOVER_TEMP_FILE_PREFIX.length + LEFTOVER_TEMP_FILE_SUFFIX.length`.

**MINOR 4 — `MonitorServices.kt:80`: the call has no `catch`.**
The orphan sweep two lines below catches `SecretStoreUnavailableException` and `IOException`, and
the doc comment states that rule. The new call has no such guard. `listFiles` and `delete` throw
only `SecurityException`, thus the real risk is small. The two lines still read as two rules.
*The fix (optional):* move the call inside the same `try`, or add one comment sentence that says
why this call needs no catch.

**MINOR 5 — `SecretStore.kt:41`: the issue number of the cross-process lock looks wrong.**
The class comment says "issue #58 adds a lock file for two monitor processes on the same folder".
Issue #58 is "Test the store against a corrupt, a locked, and a full database". The `Mutex` guards
one process only, thus a second monitor process can delete the live temporary file of the first.
The first process then fails the atomic move, retries 5 times, and throws
`SecretStoreUnavailableException`; the caller answers 503 and `apps.json` stays correct. No data
goes. This line is older than this pull request, so it is a note, not a demand.
*The fix (optional):* correct the number, or file the lock-file issue and name it.

**MINOR 6 — the language of the commit body and of `SecretStore.kt:110`.**
Two sentences join two statements with "and": "... before the first write, and removes each such
file." and "... with the count only, and it does not stop the start." Rule 4 of ASD-STE100 asks for
one instruction in one sentence.
*The fix:* split each sentence into two.

## A note, with no action in this pull request

`MonitorServicesTest.kt:32` and `SecretStoreTest.kt:23` call `Files.createTempDirectory` without
`registerTempRoot`. Issue #142 names `MonitorServicesTest` already. The new tests follow the old
pattern of the class, thus they add no new leak. Issue #142 covers the correction.

PR #169, opened 2026-09-22
fix(monitor): register the temp root of MonitorServicesTest for cleanup

### LINKED ISSUE #142: Remove each temporary test folder of the monitor backend after a run

xamcross:
## Summary

`MonitorServicesTest` made its own temporary folder, but it did not call
`registerTempRoot`. `MonitorServices.open()` starts a purge thread and a
writer thread. A native SQLite file handle can outlive `close()` for a
short time on Windows, so the eager `@AfterTest` delete could lose that
race and leave the folder behind. This change adds `registerTempRoot`
to `MonitorServicesTest`, the same pattern that pull request #146 gave
the other test classes. `ApplicationLifecycleTest` already had this
pattern from #146, so it needed no change. Each other test class of
`monitor/backend` that calls `createTempDirectory` already deletes its
own folder, in a try/finally block or in an `@AfterTest` method, on
each code path.

## Acceptance criteria

- [x] The count of folders with the prefix `octometer-` in `%TEMP%` is the same before and after a green test run.
  Evidence: two forced runs of `./gradlew :monitor:backend:test` after the fix and the merge of `main`, each `BUILD SUCCESSFUL`:
  - Run 1: 43 folders before, 43 after.
  - Run 2: 43 folders before, 45 after (other agents on this machine run the same suite at the same time and add their own folders; the set settled back to the same 43 names once their runs finished).
  - The folder set of the first snapshot of this task and the folder set after the last run are byte-identical (43 names, `diff` shows no line).
  - The count of folders with the exact prefix `octometer-monitor-services-test-` (the class this issue fixes) stayed at 2 through every run in this task, with no new name added.
- [x] No test class of `monitor/backend` calls `createTempDirectory` without a cleanup.
  Evidence: a search of `monitor/backend/src/test` for `createTempDirectory` and `createTempFile` found every call site; each one now has `registerTempRoot`, `@TempDir`, or a manual delete on every code path (a try/finally block or an `@AfterTest` method).
- [x] `./gradlew :monitor:backend:test` passes with JDK 21.
  Evidence: `java -version` reports `openjdk version "21.0.9"` (from `C:\Users\xamcr\.jdks\ms-21.0.9`). Two runs, each `BUILD SUCCESSFUL`, 316 tests, 0 failures, 0 errors (sum of `tests`/`failures`/`errors` attributes of the JUnit XML reports).
- [x] A comment on this issue records the two counts.
  Evidence: a comment on issue #142 with the run 1 and run 2 counts.

## Confirm results

- JDK 21 path: `C:\Users\xamcr\.jdks\ms-21.0.9`, confirmed with `java -version` (implementer-rules.md, and the folder listing of `C:\Users\xamcr\.jdks`).
- The seven test classes of pull request #146 and the `registerTempRoot`/`testDataDir` helpers: confirmed by reading `monitor/backend/src/test/kotlin/octometer/monitor/TestConfigs.kt` and `git log` on the changed test files.

## Notes for the maintainer

- Other agents run `./gradlew :monitor:backend:test` on their own clones of this repository at the same time as this task. Their runs add and remove their own `octometer-` folders in the shared `%TEMP%` folder of this machine, so a raw total count between two points in time can differ even when no test leaks a folder. The name-set diff and the class-specific count above give the clean evidence for this fix.
- This change deletes each leftover `octometer-` folder from 2026-09-21 and 2026-09-22 in `%TEMP%`, from earlier runs of this task and of other agents, as the task asked.

Closes #142

Generated with Claude Code


xamcross:
**Review (test engineer): APPROVE**

I made my own clone at `C:/Users/xamcr/octo-wt/review-169-test`. I ran `gh pr checkout 169` in it. I ran each command in that clone. The module sets the toolchain JDK 21 (`monitor/backend/build.gradle.kts:12`, `jvmToolchain(21)`).

## The acceptance criteria

**1. The count of folders with the prefix `octometer-` in `%TEMP%` stays the same after a green run. PASS.**

A raw count in the shared `%TEMP%` folder is not a valid measure on this machine. Other agents run the same suite at the same time. My first count gave 111 entries, and no run of mine had started. Two minutes later the same command gave 3 entries.

I made the measure valid. I pointed the test worker of my clone at a private temp folder. For the experiment I added one line to `monitor/backend/build.gradle.kts`, and I removed the line after: `systemProperty("java.io.tmpdir", "<a private folder>")`. Only my own run writes into that folder.

The result of one green run of the whole suite:

- `BUILD SUCCESSFUL in 2m 8s`.
- The count of entries with the prefix `octometer-` in the private folder after the run: **0**.
- One file stays: `sqlite-3.53.4.0-...-sqlitejdbc.dll`. The SQLite JDBC driver writes it. No test writes it, and its name has no `octometer-` prefix.

So a green run of the suite leaves no folder with the prefix. This holds for each test class of the module, not only for the class that this pull request changes.

The two runs in the shared `%TEMP%` folder, for the record: run 1 `BUILD SUCCESSFUL in 1m 44s`, run 2 `BUILD SUCCESSFUL in 1m 38s`. The counts went 111, then 52, then 1, because the runs of other agents ended in that period. The name sets are not comparable for the same reason. One folder with the name `octometer-monitor-services-test-2807486389455515260` was present right after run 2, and it was gone a few minutes later. I cannot say which clone made it, because another agent runs the same class on this machine. The private folder gives the clean answer, thus I use it and not the shared count.

**2. No test class of `monitor/backend` calls `createTempDirectory` without a cleanup. PASS.**

I searched each test source of the module for `createTempDirectory`, `Files.createTempDirectory`, `createTempFile`, and `File.createTempFile`. I found 24 folder sites and 2 file sites. No site uses `@TempDir`. Each site has a cleanup:

- 17 folder sites call `registerTempRoot`: `ApplicationLifecycleTest.kt:41` and `:89`, `AppTotalsRouteTest.kt:37`, `BackupPathsTest.kt:85`, `BackupSecretSafetyTest.kt:33`, `DailyBackupJobTest.kt:30`, `DatabaseBackupTest.kt:21`, `EventStoreTest.kt:22`, `MigrationRunnerBackupTest.kt:24`, `MonitorServicesTest.kt:34`, `PragmaTest.kt:21`, `RestoreTest.kt:26`, `RetentionPurgeTest.kt:34`, `SqliteDatabaseTest.kt:56`, `TestConfigs.kt:51`, `UserErasureServiceTest.kt:31`, `UserTotalsRouteTest.kt:37`.
- 7 folder sites have an eager removal only, with no call to `registerTempRoot`. Each one removes the folder in a `try/finally` block or in an `@AfterTest` method, on each code path. See MINOR 1.
- 2 file sites in `config/ConfigLoaderTest.kt`: line 306 calls `deleteOnExit()`, and line 313 removes the file at once.

**3. `./gradlew :monitor:backend:test` passes with JDK 21. PASS.**

Two runs in sequence in my clone, each `BUILD SUCCESSFUL`. The sum of the attributes of the JUnit XML reports: **316 tests, 0 failures, 0 errors, 0 skipped**. The count matches the pull request text.

**4. A comment on issue #142 records the two counts. PASS.**

Issue #142 holds one comment with the count of run 1 and the count of run 2.

## The checks of my role

### The root cause claim

`registerTempRoot` (`monitor/backend/src/test/kotlin/octometer/monitor/TestConfigs.kt:21-39`) adds the root to a list. One shutdown hook calls `deleteRecursively()` on each root when the test JVM ends.

The helper gives no retry, no `File.deleteOnExit`, and no log line. It calls `deleteRecursively()` one time, and it discards the Boolean result. So:

- The helper covers a handle that closes a short time after `close()`. The hook runs seconds later, thus the second attempt passes.
- The helper does not cover a handle that lives to the end of the JVM.
- The helper does not cover a JVM that a signal stops, because a shutdown hook does not run then.
- The helper gives no signal when the removal fails.

The claim itself holds. Windows puts a file with an open handle into a pending-delete state, and the removal of the parent folder then fails. `main` already had the eager `@AfterTest` removal (`git show origin/main:monitor/backend/src/test/kotlin/octometer/monitor/MonitorServicesTest.kt`, line 33), and the folders of the class still stayed. So the eager removal alone was not enough at least one time.

### The break test

I broke the change for a moment, and I restored it after. To remove the noise of the other agents, I also gave the class a unique folder prefix, `octometer-rev169mst-`. No other clone can make that name.

| Case | Runs | Result | Leftover folders |
| --- | --- | --- | --- |
| The change as it is | 3 | each `BUILD SUCCESSFUL` | 0 |
| The `registerTempRoot` call removed | 3 | each `BUILD SUCCESSFUL` | 0 |

I also ran the class one time with the call removed and with a private `java.io.tmpdir`: `BUILD SUCCESSFUL`, 0 leftover folders.

So **no test fails** when the change is gone. In four attempts **no folder stayed** either. The race is real, but it is rare, and I could not reproduce it. The change is a safety net, and it costs nothing.

A test guard is **not worth it**, and the issue is right to ask for a count. An assertion on the result of `deleteRecursively()` in `@AfterTest` would assert the very operation that loses the race, thus the test would become flaky. A guard at JVM exit has no place in a JUnit method. The helper plus a review of each new site is the correct control.

### The earlier leftover folders

- No entry with the prefix `octometer-` in `%TEMP%` has a time before today. The command `find <%TEMP%> -maxdepth 1 -name 'octometer-*' ! -newermt '2026-09-22 00:00'` gives 0 lines. So no folder from 2026-09-21 stays.
- The author removed nothing else. 1722 entries in `%TEMP%` with a time before today stay. Each clone under `C:/Users/xamcr/octo-wt/` stays.

### The scope, the diff, and the language

- The diff changes one test file and adds 6 lines. It holds no secret, no database file, and no build output.
- `ApplicationLifecycleTest` needs no change. Pull request #146 already gave it the call at line 41 and at line 89.
- The new comment uses the active voice. Its longest sentence has 16 words. It holds no `-ing` form as a noun or as a verb. The commit body follows the same rule.
- The pull request text holds no Claude session link.
- Each CI check passes: Frontend, JVM modules, Monitor backend on Windows, Secret scan, Tracker.

## The findings

No BLOCKER and no MAJOR finding stays, thus I approve. Three MINOR findings follow.

**MINOR 1** — Seven folder sites keep an eager removal only. Each one can lose the same race that this pull request names.

- `monitor/backend/src/test/kotlin/octometer/monitor/store/Migration002Test.kt:20`
- `monitor/backend/src/test/kotlin/octometer/monitor/store/MigrationRunnerTest.kt:21`, `:37`, `:64`
- `monitor/backend/src/test/kotlin/octometer/monitor/store/SqliteDatabaseTest.kt:24`
- `monitor/backend/src/test/kotlin/octometer/monitor/registry/AppRegistryServiceTest.kt:20`
- `monitor/backend/src/test/kotlin/octometer/monitor/registry/SecretStoreTest.kt:21`

Six of the seven sites open a `SqliteDatabase` in the folder, then remove the folder at once. `AppRegistryServiceTest` and `SecretStoreTest` also write a test secret file through `SecretStore`, and the goal of issue #142 names a test secret file. The level is MINOR and not MAJOR, because my green run of the whole suite left 0 entries in a private temp folder, thus each of these seven sites removed its folder.

The fix: add `.also { registerTempRoot(it) }` to each of the seven sites, and add `import octometer.monitor.registerTempRoot` to `Migration002Test.kt`, `MigrationRunnerTest.kt`, `AppRegistryServiceTest.kt`, and `SecretStoreTest.kt`. Keep each eager removal. `SqliteDatabaseTest.kt:40` asserts that the removal passes; keep that assertion, and add the call at line 24. A follow-up issue is enough, because this pull request meets each acceptance criterion.

**MINOR 2** — `monitor/backend/src/test/kotlin/octometer/monitor/TestConfigs.kt:28` discards the Boolean result of `deleteRecursively()`. A failed removal in the shutdown hook stays silent, thus a leftover folder gives no evidence in the build log.

The fix: count each root that the hook cannot remove, and write one line with that count to `System.err`. Give no path in the line.

**MINOR 3** — The pull request text marks acceptance criterion 1 as met, and it gives 43 folders before run 2 and 45 after run 2. That pair does not meet the criterion as the issue writes it. The explanation below the checkbox is correct, but the evidence line states a result that the measure does not give.

The fix: name the measure in the evidence line. A private `java.io.tmpdir` for the test worker gives a count that no other agent can change, and it gives the exact number 0.

Issue #28, opened 2026-09-21
Set the poll status and the backoff after a failure


xamcross:
**Goal.** A failed poll cycle sets a stored status and a longer delay.

**Implementation steps.**
1. In `monitor/backend`, add the stored status values of D8.
2. Add one function that maps an exception to a status with the rules of D8.
3. Count the failed cycles in `app.consecutive_failures`. A good cycle sets the value to 0.
4. On a failed cycle, set `next_poll_at` to `now + min(interval * 2^failures, 300 s)` with a jitter of 10 percent. `failures` is the count before this cycle.
5. Set `UNREACHABLE` only after 2 failed cycles in sequence. The first failed cycle keeps the old status.
6. Write only the command error code or the exception class into `last_error`.

**Acceptance criteria.**
- [ ] Virtual time: the delays are the interval, 2x, 4x, and never more than 300 s. The jitter stays inside 10 percent.
- [ ] A success after failures gives `consecutive_failures` 0 and the status `OK` of #27.
- [ ] Testcontainers with `@Testcontainers(disabledWithoutDocker = true)`: a wrong password gives `UNAUTHORIZED`. A stopped container gives `UNREACHABLE` after 2 failed cycles.
- [ ] An unknown exception gives `ERROR`.
- [ ] An exception message with a URI gives a `last_error` without the URI.
- [ ] `./gradlew :monitor:backend:test` passes.
- [ ] A comment on this issue records the test output.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, D6, D8, section 6.

xamcross:
Body edit on 2026-09-22: #27 closed. The first line now names only the open blockers: #17.

xamcross:
Note from the review of #173 on 2026-09-22: step 3 (a good cycle sets `consecutive_failures` to 0) belongs in `EventStore.recordCycleSuccess` (`monitor/backend/src/main/kotlin/octometer/monitor/store/EventStore.kt`), the one write that #27 added for the end of a good cycle. The scheduler of #17 (`monitor/backend/src/main/kotlin/octometer/monitor/poll/PollScheduler.kt`, pull request #174) writes `next_poll_at` after each cycle; the backoff of step 4 replaces its fixed `now + interval` for a failed cycle.

xamcross:
Body edit on 2026-09-22: #17 closed. No open blocker remains, thus the first line and the label `blocked` are removed. The work can start.

PR #173, opened 2026-09-22
feat(monitor): skip an invalid document and set the poll status

### LINKED ISSUE #27: Skip an invalid document and set the status of the poll cycle

xamcross:
Closes #27

## Summary

The reader checks each document of one page against section 4.1. It returns one fixed reason, or `null` for a valid document. An invalid document goes to `skipped_event` with its reason. The cursor still moves past it. `EventStore.commitPage` writes the event rows, the `skipped_event` rows, and the cursor in one transaction. A good poll cycle now sets `last_poll_at`, `last_success_at`, and the status `OK`. A cycle with one skip or more sets the status `INVALID_DATA` instead. `MongoAppReader` takes an injected `Clock` for this update, with the default `Clock.systemUTC()`.

## Acceptance criteria

- [x] A document without `element`, without `ts`, without `sessionId`, with `userId: ""`, or with a wrong BSON type goes to `skipped_event`, and the cursor moves on. Evidence: `MongoAppReaderContainerTest.five invalid documents skip with fixed reasons, and the valid documents of the page arrive`; `MongoAppReaderUnitTest.invalidReason names each missing field`, `invalidReason names a wrong BSON type`, `invalidReason names an empty userId`.
- [x] The valid documents of the same page arrive in SQLite. Evidence: the same container test asserts `eventsStored == 2` and `countEvents == 2` for the two valid documents of the seven-document page.
- [x] After a good cycle the `app` row holds `status = OK`, `last_poll_at`, and `last_success_at`. Evidence: `MongoAppReaderContainerTest.a good cycle with no skip sets last_poll_at, last_success_at, and the status OK from the injected clock` (a `Clock.fixed` reader, so the test asserts the exact epoch-millisecond value).
- [x] A cycle that skips a document sets the status `INVALID_DATA`. Evidence: `MongoAppReaderContainerTest.a cycle that skips one document sets the status INVALID_DATA`.
- [x] `./gradlew :monitor:backend:test` passes. Evidence: run two times in sequence, both green, 372 tests, 0 failures. See the issue comment for the full output.
- [x] A comment on this issue records the test output. Evidence: https://github.com/xamcross/octometer/issues/27#issuecomment-5775671522

## The no-value guarantee

- [x] A reason row and every captured log line hold no field value of an invalid document. Evidence: `MongoAppReaderContainerTest.the skipped_event reason and every captured log line hold no field value of an invalid document` (a marker string in each of the four fields, asserted absent from the `skipped_event.reason` value and from every `captureLogEvents` line); `MongoAppReaderUnitTest.the skipped_event reason holds no field value of the document` (the no-Docker version of the same check).

## Decision: an absent `userId` is anonymous

Design section 4.1 and contract rule C6 give `userId` the type "string or null". `null` means "not signed in". The rule names no separate case for an absent field. This monitor already reads `userId` with `Document.getString`. That call returns the same `null` value for an absent key and for an explicit `null` value. This pull request keeps that reading. `invalidReason` treats an absent `userId` the same as an explicit `null`: both are a valid anonymous event. Only `userId: ""` is invalid (D5).

## Test count

- `MongoAppReaderContainerTest`: 14 tests (Testcontainers MongoDB, `@Testcontainers(disabledWithoutDocker = true)`).
- `MongoAppReaderUnitTest`: 26 tests (no Docker).
- `EventStoreTest`: 11 tests (no Docker).
- Full `monitor/backend` module: 372 tests, 0 failures, run twice in sequence.

Four checks were each proven by a temporary break, then a restore: `invalidReason`'s field checks, `EventStore.commitPage`'s `skipped_event` insert, `EventStore.recordCycleSuccess`'s three columns, and `MongoAppReader.pollOnce`'s status choice. Each one failed its test with the code broken. Each one passed again once restored.

## Files changed

- `monitor/backend/src/main/kotlin/octometer/monitor/mongo/MongoAppReader.kt`
- `monitor/backend/src/main/kotlin/octometer/monitor/store/EventStore.kt`
- `monitor/backend/src/test/kotlin/octometer/monitor/mongo/MongoAppReaderContainerTest.kt`
- `monitor/backend/src/test/kotlin/octometer/monitor/mongo/MongoAppReaderUnitTest.kt`
- `monitor/backend/src/test/kotlin/octometer/monitor/store/EventStoreTest.kt`

No schema migration was needed: `db/001_init.sql` on `main` already holds the `skipped_event` table and the `app.status`, `app.last_poll_at`, and `app.last_success_at` columns of section 6. The monitor README lists no status value table, so this pull request leaves it as it is.

Generated with Claude Code


xamcross:
**Review (Kotlin, MongoDB, and SQLite engineer): APPROVE**

I made my own clone, and I ran `gh pr checkout 173` in it. Each result below comes from my own run
on JDK 21, with Docker present.

## The acceptance criteria

1. **PASS.** A document without `element`, without `ts`, without `sessionId`, with `userId: ""`, or
   with a wrong BSON type goes to `skipped_event`, and the cursor moves on.
   `MongoAppReaderContainerTest.kt:319` runs the five cases in one page of seven documents. I also
   added my own probe with a page of three invalid documents and no valid one. The result:
   `eventsStored == 0`, `eventsSkipped == 3`, and the `app.cursor` column held the `_id` of the last
   document of the page.
2. **PASS.** The valid documents of the same page arrive. `MongoAppReaderContainerTest.kt:338`:
   `eventsStored == 2` and `countEvents == 2`.
3. **PASS.** After a good cycle the `app` row holds `status = OK`, `last_poll_at`, and
   `last_success_at`. `MongoAppReaderContainerTest.kt:386`. The unit is epoch milliseconds UTC. It is
   the same unit as `created_at`, because `AppRegistryService.kt:224` writes
   `System.currentTimeMillis()`.
4. **PASS.** A cycle that skips a document sets `INVALID_DATA`.
   `MongoAppReaderContainerTest.kt:403`. My probe confirmed the same for a page with no valid
   document.
5. **PASS.** `./gradlew :monitor:backend:test`: `BUILD SUCCESSFUL in 3m 37s`. The XML results give
   372 tests, 0 failures, 0 errors, 0 skipped, over 34 classes. `MongoAppReaderContainerTest` 14,
   `MongoAppReaderUnitTest` 26, `EventStoreTest` 11. The test count of the pull request text is
   correct.
6. **PASS.** The issue comment holds the test output.

## Step 1: `invalidReason` against the field list of section 4.1 and of the contract

| Field | Type in 4.1 and in the contract | The check of `invalidReason` |
|---|---|---|
| `_id` | ObjectId (C2) | **Not checked.** See MINOR 1. |
| `ts` | Date (C3) | Presence and `!is Date`. The right type. |
| `element` | String, 1 to 100 characters, pattern `[A-Za-z0-9_.:-]+` (C4) | Presence and `!is String`. The right type. The length and the pattern stay unchecked. See MINOR 2. |
| `sessionId` | String, a UUID (C5) | Presence and `!is String`. The right type. The UUID form stays unchecked. See MINOR 2. |
| `userId` | String or null, 1 to 254 characters (C6) | An absent field and an explicit `null` pass. A non-String gives `userId wrong type`. An empty string gives `userId empty`. The length limit stays unchecked. See MINOR 2. |
| `path` | String, optional (C39) | **Not checked.** The reader stores no `path` column yet, thus a wrong type causes no fault today. |
| `referrerHost` | String, optional (C40) | **Not checked.** The same case as `path`. |
| An unknown extra field | C9: a reader ignores it | **Correct.** `invalidReason` reads only the named keys. `parseEvent` reads only the five named fields. Two tests prove it. |

I agree with the decision on an absent `userId`. Rule C6 gives the type "string or null".
`Document.getString` returns the same `null` value for an absent key and for an explicit `null`.
The pull request keeps that reading.

## Step 3: the status after each kind of cycle

- A good cycle: `OK`, plus `last_poll_at` and `last_success_at` from the injected clock, in epoch
  milliseconds UTC.
- A cycle with a skip: `INVALID_DATA`.
- A failed cycle: no change. I proved it with a probe. A reader with a throwing `serverTimeSource`
  gave `MongoReadFailedException`, and the three columns kept the values of the earlier good cycle.
- A cycle with zero pages, on an empty collection: the code writes `OK`, `last_poll_at`, and
  `last_success_at`, and it leaves `cursor` at `NULL`. My probe printed
  `PollOutcome(eventsStored=0, pagesRead=0, cursor=null, eventsSkipped=0)` and the row
  `(OK, 1790071200000, 1790071200000)`. **This is right under D5.** D5 says "at the end of a good
  cycle". A cycle that reads no new event is a good cycle. It also removes the "–" of level 1 that
  the Evidence of the issue names. `AppTotalsRoute.kt:107` derives `NEVER_POLLED` from
  `last_poll_at IS NULL`, thus the derived value stays correct.

## Step 4: the transaction and the `write {}` rule

- **No blocking call runs inside `write {}`.** `commitPage` and `recordCycleSuccess` each run JDBC
  only. `MongoAppReader.kt:165` reads `clock.millis()` before the call.
- **The page commit runs in one transaction.** `BEGIN IMMEDIATE`, the event insert, the
  `skipped_event` insert, the cursor update, `COMMIT` (`EventStore.kt:54` to `EventStore.kt:63`). I
  proved the rollback. I moved `insertSkippedEvents` before `BEGIN IMMEDIATE`, and the test
  `a failed insert rolls back a skipped_event row too` failed.
- **The status write is one further `write {}`** (`EventStore.kt:82`). This form is right. Three
  reasons: a cycle with zero pages commits no page, thus the same transaction would never write the
  status; D4 fixes the shape of the page transaction; the choice between `OK` and `INVALID_DATA`
  needs the whole cycle and not one page.
- The risk of the present form: the monitor can stop between the last page commit and the status
  write. The cursor then holds the new value, and the status holds the old value. The next cycle
  repairs it. No event is lost.
- The risk of the other form: a zero-page cycle would write no status at all. A cycle of ten pages
  would write the status ten times, and each early value could be wrong.

## Step 5: the two default parameters and issue #17

Issue #17 has no branch and no pull request today. Both call forms stay valid, because each new
parameter sits last and holds a default:

- `PollOutcome(eventsStored, pagesRead, cursor)` compiles. The named form compiles too.
- `commitPage(appId, events, cursor)` compiles. `EventStoreTest.kt:116` guards it.
- `MongoAppReader(eventStore, settleLagSeconds, clientFactory, serverTimeSource, cycleTimeoutMillis)`
  compiles, because `clock` sits last.

One note for issue #17: `pollOnce` now writes to SQLite at the end of each cycle. A scheduler test
needs a real `app` row for each poll target. Without the row, `recordCycleSuccess` throws
`IllegalStateException` (`EventStore.kt:89`).

## Step 6: can each new test fail?

I broke the code seven times, I saw the failure each time, and I restored the code each time. Each
run used `./gradlew :monitor:backend:test --tests <class>`.

1. Remove `ON CONFLICT(app_id, event_id) DO NOTHING` of `INSERT_SKIPPED_EVENT_SQL` →
   `a replay with the same skipped events adds 0 new skipped_event rows` FAILED.
2. Move `insertSkippedEvents` before `BEGIN IMMEDIATE` → `a failed insert rolls back a skipped_event
   row too` FAILED.
3. Add the field value to the reason text → `the skipped_event reason holds no field value of the
   document` and `invalidReason names a wrong BSON type` FAILED.
4. Reject an absent `userId` → `invalidReason accepts an absent userId field as an anonymous event`
   and `invalidReason accepts an explicit null userId as an anonymous event` FAILED.
5. Pass `emptyList()` to `commitPage` inside `runCycle` → `runCycle writes a skipped_event row with
   the fixed reason` FAILED.
6. Reject each String `userId` → `invalidReason accepts a valid document` FAILED, plus five page-loop
   tests.
7. Write `0` and the literal `"OK"` in `recordCycleSuccess` → both `recordCycleSuccess` tests FAILED.

One new test cannot fail at run time: `commitPage with no skippedEvents argument keeps the old caller
of issue 16 valid` (`EventStoreTest.kt:116`). It guards the default at compile time only. That is its
purpose, thus I keep it.

No new test uses `Thread.sleep`.

## Step 8: the language of each new comment

I scanned each comment block of the five changed files with a script. No sentence holds more than 25
words. The split of the long sentences of #16 is complete in the three reader files. Each `-ing` word
is a noun of the domain (`string`, `finding`, `nothing`, `sibling`) or an adjective (`the calling
coroutine`). The active voice holds throughout.

## The findings

No BLOCKER. No MAJOR. Four MINOR findings follow.

**MINOR 1 — `invalidReason` does not check `_id`.**
`monitor/backend/src/main/kotlin/octometer/monitor/mongo/MongoAppReader.kt:367`.
Section 4.1 and rule C2 give `_id` the type ObjectId. `runCycle` at line 290 and `parseEvent` at line
343 both call `document.getObjectId("_id")`. That call throws `ClassCastException` for another type.
I tested the real path against MongoDB 7.0. I wrote a document with a String `_id` beside a good
document, and I ran one cycle. The cycle passed with `eventsStored == 1`. MongoDB brackets `$lt` by
BSON type, thus the filter of `idFilter` (line 402) never returns a document with a non-ObjectId
`_id`. **The reader is therefore safe today.** The gap that stays: such a document is lost with no
record. No `skipped_event` row names it, and the status stays `OK`.
The fix: add a `$type` term with the value `objectId` to the `_id` range of `idFilter`, and name the
type bracket in the KDoc of `invalidReason`. A later change that adds a projection, or that reads by
a different filter, then cannot reach the `ClassCastException`.

**MINOR 2 — `invalidReason` checks the presence and the type only, not the value rules of 4.1.**
`monitor/backend/src/main/kotlin/octometer/monitor/mongo/MongoAppReader.kt:367` to line 381.
`element` has no length check (1 to 100) and no pattern check (`[A-Za-z0-9_.:-]+`). `sessionId` has
no UUID check. `userId` has no length check (1 to 254). D5 names only "an absent field, a wrong BSON
type, or an empty `userId`", thus this pull request meets D5. One case still reaches SQLite: an empty
`element` string. The column `event.element` is `TEXT NOT NULL` with no `CHECK`, thus an empty value
lands in the level 3 table.
The fix: add `if (element.isEmpty()) return "element empty"` after the type check of `element`. File
a follow-up issue for the full value rules of C4, C5, and C6.

**MINOR 3 — two behaviours of the KDoc hold no test.**
`monitor/backend/src/main/kotlin/octometer/monitor/mongo/MongoAppReader.kt:157`.
The KDoc states that a failed cycle changes none of the three columns. No test asserts it. A cycle
with zero pages also writes `OK`, and no test asserts that either. I proved both with a probe class
in my own clone, thus the code is right.
The fix: add two tests to `MongoAppReaderContainerTest`. Test one: a reader with a throwing
`serverTimeSource` after a good cycle, then assert the three columns keep their values. Test two: a
poll of an empty collection, then assert `OK` and the two times from the injected clock.

**MINOR 4 — a good cycle does not reset `consecutive_failures`.**
`monitor/backend/src/main/kotlin/octometer/monitor/store/EventStore.kt:32`.
D8 shows `UNREACHABLE` after 2 failed cycles in sequence. The column `app.consecutive_failures`
exists in `db/001_init.sql`, and no code writes it yet. Issue #28 adds the failure count. A count
that never resets after a good cycle gives a wrong status later.
The fix: add `consecutive_failures = 0` to `RECORD_CYCLE_SUCCESS_SQL` now, or name the reset in the
text of issue #28.

## Other checks

- The diff holds five files only. It holds no secret, no database file, and no build output.
- The scope matches the issue. The pull request adds no route, no config key, and no migration.
- `db/001_init.sql` on `main` already holds the `skipped_event` table and the three `app` columns.
  The claim of the pull request text is correct.
- The frontend already knows the value `INVALID_DATA` (`monitor/frontend/src/app/apps/app-row.ts:12`),
  and it marks each status other than `OK` and `NEVER_POLLED` as a warning. The new value needs no
  frontend change.
- I delete my clone and each scratch file after this comment.

xamcross:
**Review (application security engineer): APPROVE**

I made my own clone, I ran `gh pr checkout 173`, and I ran each check myself. I write each result below with my own evidence.

## The acceptance criteria

|---|---|---|
| An invalid document goes to `skipped_event`, and the cursor moves on | PASS | My own container test inserted five invalid documents in one page. The table held exactly five rows: `sessionId missing`, `userId empty`, `userId wrong type`, `sessionId wrong type`, `ts wrong type`. The cursor moved to the last `_id` of the page. |
| The valid documents of the same page arrive | PASS | The same cycle stored the valid documents of the page. A second cycle read 0 documents. |
| A good cycle sets `status = OK`, `last_poll_at`, `last_success_at` | PASS | I broke `EventStore.recordCycleSuccess` (`setLong(2, 0L)`). The container test and `EventStoreTest.recordCycleSuccess sets last_poll_at, last_success_at, and the status` both failed. I restored the line. |
| A cycle with a skip sets `INVALID_DATA` | PASS | I broke the status choice of `pollOnce` to a constant `OK`. The container test `a cycle that skips one document sets the status INVALID_DATA` failed. I restored the line. |
| `./gradlew :monitor:backend:test` passes | PASS | One run in my clone: `BUILD SUCCESSFUL in 3m 8s`. The XML results give 372 tests, 0 failures, 0 skipped. `MongoAppReaderContainerTest` 14, `MongoAppReaderUnitTest` 26, `EventStoreTest` 11. |
| A comment records the test output | PASS | The comment is on issue #27. |

## The security checks

**1. No field value leaves the document. PASS.**
I built five invalid documents. Each one held a marker: in `element`, in `path`, in `userId` with a wrong BSON type, in `sessionId` as a number, and in an unknown field. I ran one cycle against a real MongoDB, and one cycle with a fake page source. I then read the whole `skipped_event` table and each captured log event at every level. I checked the formatted message, the argument array, and the whole chain of the log throwable. No marker appeared. Each `event_id` value held the hex `_id` and nothing else.

**2. The WARN line of the cycle. PASS.**
The line is `The reader skipped 5 invalid document(s) of one page.` It holds the count only. It holds no document value, no app name, and no part of a connection string.

**3. The status write. PASS.**
A good cycle writes the three columns from the injected clock. A failed cycle changes none of them. I proved this two ways. First, a page source that threw after the first page left `status`, `last_poll_at`, and `last_success_at` at `null`, and the committed first page stayed. Second, a `pollOnce` with a failed server-time read kept the three old values (`OK`, `1700000111000`, `1700000111000`).

**4. One transaction for one page. PASS.**
I made the `skipped_event` insert bind a `NULL` reason for a moment. `commitPage` threw `SQLITE_CONSTRAINT_NOTNULL`. The event rows of that page were absent (0 rows), the cursor stayed `null`, and the `skipped_event` table was empty. I restored the line.

**5. The bound parameters. PASS.**
Each new statement binds each value with `setLong` or `setString`. I sent a single quote and a `DROP TABLE` text through a cycle as an `element`, a `sessionId`, a `userId`, and a `ts`. The output of `sqlite_master` was the same before and after. The quote text arrived as a plain value in the `element` column.

**6. The anonymous case. PASS.**
An absent `userId` and an explicit `null` both count as anonymous. I agree with the decision, and the pull request text states it. Contract rule C6 gives the field the type "string or `null`", and it names no separate absent case. The kit store writes `userId` on each event, so a real app always sets the field. The kit also deletes an anonymous event with the filter `{userId: null}`, and that filter matches an absent field too. A strict reading of D5 ("an absent field") points the other way, so please add one sentence to D5 or to rule C6. A later agent must not change this behaviour back.

**7. Each new test can fail. PASS.**
Four temporary breaks gave seven failures, then each restore gave a green run again:
- A reason text with the value in it (`"ts wrong type: $ts"`) failed `the skipped_event reason holds no field value of the document` and `invalidReason names a wrong BSON type`. This is the important one. The no-value test is real.
- A removed empty-`userId` check failed `invalidReason names an empty userId`.
- A `NULL` reason failed `commitPage inserts each skipped event with its reason...` and `a replay with the same skipped events adds 0 new skipped_event rows`.
- A broken status write and a broken status choice failed the three status tests.

**8. No secret and no build output in the diff. PASS.** The diff holds five source files only.

## The findings

All six are MINOR. None of them blocks a merge.

**MINOR 1. `monitor/backend/src/main/kotlin/octometer/monitor/mongo/MongoAppReader.kt:358`.**
The comment says "Checks one event document against section 4.1". The commit body says "The reader checks each document field of section 4.1". The function checks the presence and the BSON type only. Section 4.1 also gives `element` 1 to 100 characters with the pattern `[A-Za-z0-9_.:-]+` (rule C4), `sessionId` the form of a UUID (rule C5), and `userId` 1 to 254 characters (rule C6). No check runs for those three rules. The code agrees with D5, because D5 names only an absent field, a wrong BSON type, and an empty `userId`. The comment claims more than the code does.
Fix: change the first sentence to "Checks the presence and the BSON type of each field of section 4.1 (design decision D5)." Then open a follow-up issue for the length rule and the pattern rule of C4, C5, and C6.

**MINOR 2. `monitor/backend/src/main/kotlin/octometer/monitor/store/EventStore.kt:34` to `:40`.**
This pull request adds one sentence to the comment. The paragraph now holds seven sentences. The language policy allows six.
Fix: put a `*` line before "All three steps sit inside one `BEGIN IMMEDIATE` transaction." at line 38.

**MINOR 3. `monitor/backend/src/test/kotlin/octometer/monitor/mongo/MongoAppReaderContainerTest.kt:39`.**
The split work of issue #16 left one semicolon: "Each test needs Docker; a machine with no Docker skips the whole class." I read ten other split sentences in the three reader files. Each other one is correct.
Fix: "Each test needs Docker. A machine with no Docker skips the whole class."

**MINOR 4. The pull request text and the issue comment.**
Two sentences use the passive voice: "Four checks were each proven with a temporary break, a failing run, then a restore" and "No schema migration was needed". The text "a failing run" also uses an `-ing` form.
Fix: "A temporary break proved each of four checks. Each broken check failed its test. Each restore gave a green run." And: "This change needs no schema migration."

**MINOR 5. `monitor/backend/src/main/kotlin/octometer/monitor/retention/RetentionPurge.kt:30`.**
This pull request is the first code that writes a row into `skipped_event`. The retention purge deletes from `event` only. The erasure service of `UserErasureService.kt:60` to `:79` deletes from `event` only. `AppRegistryService.kt:262` deletes a `skipped_event` row only when the owner deletes the whole app. Thus the table grows with no bound, from data that the monitored app controls. One broken app deployment adds a permanent row for each invalid document.
Fix: open a follow-up issue. Add `skipped_event` to the retention purge, with the same retention as `event`.

**MINOR 6. `monitor/backend/src/main/kotlin/octometer/monitor/mongo/MongoAppReader.kt:290` and `:296`.**
A document whose `_id` is not an ObjectId never reaches `skipped_event`. I put such a document in the collection with five valid fields. The cycle gave `eventsStored=1, eventsSkipped=0`, and it stored the good document only. MongoDB type bracketing is the cause: the filter `{_id: {$lt: <ObjectId>}}` of section 4.3 matches an ObjectId value only. The reader therefore never sees the document, and it never writes a row for it. Contract rule C2 gives `_id` the type ObjectId, so this is an invalid document under D5. Both lines also call `getObjectId("_id")` with no type check. The filter keeps them safe today, and a reader of the code cannot see why.
Fix: add one sentence to the comment of `runCycle`: "The filter of section 4.3 returns an ObjectId `_id` only, thus `getObjectId` is safe here. A document with a different `_id` type stays out of each page."

## One suggestion

`MongoAppReaderContainerTest.kt:381` reads `event.formattedMessage` only. A log line can also carry a marker inside its throwable. My own run checked the formatted message, the argument array, and the whole throwable chain, and it found no marker. Please add the throwable to the assertion, so a later change cannot leak a value through an exception message.

Issue #171, opened 2026-09-22
Remove the dead secret-scan allow-list entries


xamcross:
**Goal.** The secret-scan configuration names only findings that exist on `main`.

**Evidence.** Observed on 2026-09-22, after the squash merges of pull requests #160 and #161:
- `.gitleaksignore` (added by #160, commit 293b526 on `main`) holds two fingerprints. Each names the commit `a8657fbde1d0613c503022741f5542e6f5eb94de` of the branch `issue-16-mongo-reader`. The squash merge dropped that commit; `git merge-base --is-ancestor a8657fb origin/main` gives exit code 1, and no remote branch holds it. The identifier `markerPassword` that the rule `generic-api-key` matched is now `markerSecondValue` (`monitor/backend/src/test/kotlin/octometer/monitor/mongo/MongoAppReaderContainerTest.kt:197`, `MongoAppReaderUnitTest.kt:274`).
- `.gitleaks.toml:34-39` holds two regexes under `[[rules.allowlists]]`. The second one (line 38) covers the commit `a92504b` of the branch of pull request #161. That branch is deleted, and the commit is absent from the repository.
- The first regex (line 37) stays. The fixture is in use: `monitor/frontend/src/app/manage/add-app-form.spec.ts:54,105` and `monitor/backend/src/test/kotlin/octometer/monitor/registry/MongoTestUris.kt:9`.
- The second security review of #160 (https://github.com/xamcross/octometer/pull/160#issuecomment-5775093541) states that a fingerprint in `.gitleaksignore` dies on a squash, and that the right tool for a test value is a top-level `[[allowlists]]` with `paths` and `targetRules`. No such value exists on `main` today, thus no new allow-list entry is needed.

**Implementation steps.**
1. Delete `.gitleaksignore`.
2. Remove the second regex (line 38) from `.gitleaks.toml`. Shorten the `description` of line 35 to the first fixture only, and remove the sentence on pull request #161.
3. Open a pull request. The job `Secret scan` runs `gitleaks detect` over the full history (`.github/workflows/ci.yml:141-151`, `fetch-depth: 0`).
4. After the merge, read the run of the job `Secret scan` on `main`.

**Acceptance criteria.**
- [ ] `.gitleaksignore` is absent from `main`.
- [ ] `.gitleaks.toml` holds one allow-list regex, the test fixture of line 37.
- [ ] The job `Secret scan` passes on the pull request and on `main` after the merge.
[...]

PR #185, opened 2026-09-22
test(monitor): add the end-to-end test from a click to the level 1 totals

### LINKED ISSUE #21: Add the end-to-end test from a click to the level 1 totals
**Goal.** One JVM test in `monitor/backend` proves the path from an ingest request to the level 1 totals.

**Implementation steps.**
1. Add one JVM test in `monitor/backend` with a Testcontainers MongoDB and `@Testcontainers(disabledWithoutDocker = true)`.
2. Start the ingest route of the demo app in the same process, with the container as its database.
3. Send synthetic clicks of several user ids and sessions through the HTTP ingest route.
4. Start the monitor in dev mode with the real reader and the real SQLite store.
5. Register the app through `POST /api/apps` with the URI of the container.
6. Assert the totals of `GET /api/apps` after one poll cycle.

**Acceptance criteria.**
- [x] The totals equal the sent clicks, users, and sessions.
- [x] The row of the app holds `status: OK` and a `lastSuccessAt` value.
- [x] After a PATCH of the connection string the next cycle uses the new string.
- [x] The test needs less than 60 seconds.
- [x] The test runs in the `jvm` CI job.
- [x] A comment on this issue records the CI run URL and the test time.

**Source.** `docs/superpowers/specs/2026-09-21-octometer-design.md`, sections 4.2 and 4.3, D13, D39.

xamcross:
## Summary

This pull request adds `EndToEndContainerTest` to `monitor/backend`. The test proves the
full path of section 4: an HTTP click travels through the ingest route of `kit/jvm-ktor`,
into two real MongoDB containers, through one poll cycle of the real monitor, and into the
level 1 totals of `GET /api/apps`.

The test found a real gap: `MongoAppReader` kept the MongoDB client of one app id for
ever, keyed on the app id alone, with no check of the connection string. Design decision
D10 asks for a close "after a PATCH of the connection string", but no code did that close.
A PATCH then had no effect on the next poll cycle. This pull request fixes that one gap in
`MongoAppReader.kt`, so the PATCH acceptance criterion of this issue is true.

The first form of this fix kept the connection string itself in a field. Design decision
D11 forbids that (BLOCKER 1 of the security review of issue #16). The fix now keeps the
SHA-256 hex of the connection string instead, the same form as the long-key hash of the
ingest rate limiter (issue #33). Two new Docker-free unit tests in `MongoAppReaderUnitTest`
prove the cache (reuse, close-and-replace) and prove, through reflection, that no field of
the reader or of a cached client entry holds the connection string after the call.

### Corrections after the two reviews of this pull request

The security review and the Kotlin review each read commit `cf29561`. This section
records the fix of each MAJOR finding. The pull request comment holds the full table of
every finding.

- **Security MAJOR 1.** `clientFor` now builds the fresh client first. It then puts it in
  the map. It closes the old client last. A failed build now leaves the old, working
  client in place. The monitor then needs no restart. A new unit test proves this. A
  factory throws for the second connection string. A third call, with the first string,
  reuses the first client. A fourth call, with a working second string, closes the first
  client.
- **Security MAJOR 2.** The log sentinel of `EndToEndContainerTest` now reads
  `throwableProxy` of each log event, and each `cause`, for the class name and the message
  of each one. It also marks the host and the port of each container, and a marker user
  name and password from a short auth probe. A `log.warn` with a marker string inside an
  attached exception now fails the test (proved, then reverted).
- **Kotlin MINOR 3.** `CachedClient` is no longer a data class. It prints the fixed text
  `"CachedClient"`. A future log line of the map then holds no hash and no client text.
- **Kotlin MINOR 8.** `monitor/backend/build.gradle.kts` drops the Mockito test
  dependency. `MongoAppReaderUnitTest` now doubles `MongoClient` through the driver's own
  public constructor, with Kotlin interface delegation over a real reactive-streams
  client.
- **Kotlin MINOR 4.** The class time budget of `EndToEndContainerTest` now starts before
  the two containers start, in a companion object field. It no longer excludes the
  container start from the measured time.
- **Kotlin MINOR 1.** The class comment of `EndToEndContainerTest` names issue #110 as the
  owner of the click-count gap (the `kind` mark of `octo:session-start`).

## Acceptance criteria

- [x] The totals equal the sent clicks, users, and sessions. Evidence: the test asserts
  `clicks`, `uniqueUsers`, and `uniqueSessions` of `GET /api/apps` against the named
  constants `SENT_CLICKS`, `SENT_USERS`, and `SENT_SESSIONS` of the test class. One
  `octo:session-start` entry of each session counts as a click today (see the class
  comment, and issue #110). The test adds that count in `EXPECTED_FIRST_CLICKS`.
- [x] The row of the app holds `status: OK` and a `lastSuccessAt` value. Evidence: the
  test asserts both fields after the first poll cycle and again after the PATCH.
- [x] After a PATCH of the connection string the next cycle uses the new string.
  Evidence: the test sends a second batch to a second MongoDB container, then a PATCH,
  then it waits for the totals to grow by that second batch. This criterion needed the
  fix of `MongoAppReader.kt`; the unpatched code left the totals frozen, and the test
  failed with a clear timeout message until the fix landed.
- [x] The test needs less than 60 seconds. Evidence: the test measures its own run time,
  from before the two containers start, and prints it. Each local run of this correction
  finished in a range of 10.6 s to 11.5 s, well under the budget. Gradle shows a passing
  test's stdout only on failure. This number therefore does not appear in the CI console;
  the CI job passes the same code path (see the next criterion).
- [x] The test runs in the `jvm` CI job. Evidence: the class name ends with
  `ContainerTest`, so the existing `OCTOMETER_REQUIRE_DOCKER` guard of
  `monitor/backend/build.gradle.kts` already covers it (it matches by name suffix, not
  by a fixed list; no change to that guard was needed). CI run
  (`jvm` job, "JVM modules", commit `ed51755`):
  https://github.com/xamcross/octometer/actions/runs/35763607627/job/106867704829. The
  Windows job ("Monitor backend on Windows") also runs on this commit:
  https://github.com/xamcross/octometer/actions/runs/35763607627/job/106867704737. Both
  jobs were still running when this text was last updated; the pull request comment
  below records their final result.
- [x] `./gradlew :monitor:backend:test` passes. Evidence: a full local run of the module,
  with `OCTOMETER_REQUIRE_DOCKER=true`, gave 428 tests, 0 failures, 2 skipped
  (`FrontendCspTest` and `WindowsStartScriptTest`, each gated on `installDist`, the same
  known skip of a plain test run). The count grew from 426 to 428 with the two new unit
  tests of the security MAJOR 1 and Kotlin MINOR 3 fixes above.
- [x] A comment on this issue records the CI run URL and the test time. Evidence: posted
  on issue #21 (https://github.com/xamcross/octometer/issues/21#issuecomment-5780538036),
  with the CI run URL and the test time, updated after this correction.

## Design decisions followed

- Decision 1: one class, `EndToEndContainerTest`, suffix `ContainerTest`.
- Decision 2: `monitor/backend` gets no test dependency on `tools:demo-app`. That module
  builds its tracker page with `npm` (`npmCiTracker`, `buildTracker`), and the
  "Monitor backend on Windows" CI job sets up no Node for that build; a project
  dependency would have pulled that build into this module's Windows test task. The
  test starts the ingest route of `kit/jvm-ktor` directly, with the same
  `IngestSettings` and `MongoEventLogStore` as the demo app.
- Decision 3: the sent clicks, users, and sessions sit in the named constants
  `SENT_CLICKS`, `SENT_USERS`, `SENT_SESSIONS` (and the second-source equivalents). Each
  session sends `octo:session-start` first (contract rule C38), through the cookie form
  of the demo app's user id resolver.
- Decision 4: `MonitorConfig` uses `pollIntervalSeconds = 1` and `settleLagSeconds = 1`,
  a temporary data directory through `testDataDir`, and the real `module(config)` (the
  real reader, the real scheduler, both through `MonitorServices.open`). The test polls
  `GET /api/apps` with a 30 s bound; it never sleeps a fixed time.
- Decision 5: the PATCH test uses a second MongoDB container (`SECOND_SOURCE`), because
  the `PATCH` route accepts only `name` and `connectionString`
  (`UpdateAppRequest`), never `database`; the second container keeps the same database
  name `exampledb`.
- Decision 6: `status: OK` and `lastSuccessAt` are asserted after both poll cycles.
- Decision 7: `pollIntervalSeconds = 1` and `settleLagSeconds = 1` keep the wait short;
  the test prints its own run time.
- Decision 8: the `OCTOMETER_REQUIRE_DOCKER` guard of `monitor/backend/build.gradle.kts`
  matches any class name ending with `ContainerTest`, so it already covers the new
  class; no edit to that guard was needed. Confirmed by reading the guard.

## The break-and-restore proofs

- The reader's collection name changed from `octometer_events` to `octometer_events_v2`
  for one local run (a name the registry still accepts, but with no data in it). The
  test failed with a clear message (the totals never reached the expected values within
  the 30 s bound). The name was restored, and the test passed again.
- The PATCH fix itself: with the old `MongoAppReader.clientFor` (keyed on the app id
  alone), the end-to-end test failed with the totals frozen at the pre-PATCH values.
  With the fix, the same run passes.
- The hash-based cache, after the first review correction: `clientFor` reverted for a
  moment to a form that reused any kept client regardless of a changed connection
  string. The new unit test `clientFor closes the old client on a changed connection
  string, and builds one client for each distinct string` failed with a clear assertion
  message. Restored, the same test passes.
- Security MAJOR 1, this correction: `clientFor` reverted for a moment to the old form
  (close the old client, then call the factory). The new unit test `a failed client
  build leaves the old client open, and a later working string closes it` failed with a
  clear assertion message. Restored, the same test passes.
- Security MAJOR 2, this correction: `PollScheduler.runPollCycle` gained one temporary
  `log.warn` line with the real connection string inside an attached `RuntimeException`.
  `EndToEndContainerTest` then failed at its sentinel assertion. The line was removed,
  and the same test passed again.

## Connection strings, never in a field or in text

- The test builds each MongoDB URI from a live Testcontainers container
  (`MongoDBContainer.getConnectionString()`), with no user and no password. No text of
  this pull request, this file included, holds a credential-shaped string; each mention
  above uses the placeholder form `mongodb://<host>:<port>/exampledb`.
- `MongoAppReader`'s client cache now keeps the SHA-256 hex of the connection string,
  never the string itself (design decision D11). The unit test now walks the whole
  graph of the reader, of a cached client entry, and of the client's own settings, with
  a marker user name, password, host, and port. It found the whole connection string
  absent everywhere, the same finding as the security review's own probe.

Closes #21

🤖 Generated with [Claude Code](https://claude.com/claude-code)


xamcross:

I made my own clone. I ran `gh pr checkout 185` in it, on the commit `cf29561`. I ran each check
myself. I restored each file after each experiment, and `git status` is clean now. The module
suite gave 426 tests, 0 failures, 0 errors, and 2 skipped. The two skips are `FrontendCspTest`
and `WindowsStartScriptTest`. Each one needs a build artifact that a plain test run does not
make. No container test skipped, thus Docker ran.

## Acceptance criteria

| --- | --- | --- |
| The totals equal the sent clicks, users, and sessions | PASS | `EndToEndContainerTest`, 1 test, 0 skipped, in my clone |
| The row holds `status: OK` and a `lastSuccessAt` value | PASS | The same class asserts both fields two times |
| A PATCH of the connection string moves the next cycle | PASS | The same class; see MAJOR 1 for the failure path |
| The test needs less than 60 seconds | PASS | 11.4 s in the result XML of my run |
| The test runs in the `jvm` CI job | PASS | The run 35756776353 sets `OCTOMETER_REQUIRE_DOCKER: true`. `Task :monitor:backend:test` ran, and not from the cache |
| `./gradlew :monitor:backend:test` passes | PASS | 426 tests, 0 failures, 0 errors |
| A comment on the issue holds the CI run URL and the test time | PASS | I read it with the GitHub API |

## The client cache and the hash form

I ran my own marker probe. I gave `clientFor` a loopback URI with a marker user name and a
marker password. I used the real client factory, not a double. I then walked the object graph
from the reader. The walk read each field, each map entry, each array, and the `toString()` of
each object. Here is the result:

- The whole connection string is **absent** everywhere. No field holds it. No `toString()`
  prints it. The hash form of this pull request is correct on that point.
- The marker user name is a `String` at
  `reader.clients{}.client.wrapped.settings.credential.userName`.
- The marker password is a `char[]` at `reader.clients{}.client.wrapped.settings.credential.password`.
- The host and the port are at `reader.clients{}.client.wrapped.cluster.settings.hosts[]`.
- `MongoClientSettings.toString()` and `MongoCredential.toString()` both print the user name.
  Both print `password=<hidden>`.

**My verdict on the hash.** Keep it. It removes a second copy of the secret from the heap, and
it costs almost nothing. A heap dump gives the hash and not the string. It is not a secrecy
boundary, because the driver keeps the user name and the password beside the hash, in the same
map entry. The driver needs them for each reconnect, thus no design of this class can remove
them. A salt or an HMAC key adds **no** value here. The hash stays inside one process. It never
reaches a file, a log line, or the network. An attacker who reads the hash already reads the
credential of the driver next to it. A rainbow attack on a random password is not feasible
anyway.

Please correct one word of the pull request text. The text says that the unit test proves the
absence of the connection string "through reflection over the fields of the reader and of the
cached client entries". The test proves a much smaller claim. See MINOR 1.

## MAJOR 1. A failed client build leaves a closed client in the cache

`monitor/backend/src/main/kotlin/octometer/monitor/mongo/MongoAppReader.kt:267` to `:277`.

`clientFor` closes the old client at line 273. It then calls the factory at line 274. Both calls
sit inside `ConcurrentHashMap.compute`. A failure of the factory leaves the map entry unchanged;
that is the contract of `compute`. The entry then holds the hash of the old string and a client
that is already closed.

The path of a real operator:

1. The app polls with the string A.
2. The operator sends a PATCH with the string B. B passes `ConnectionStringValidator`, and the
   driver still rejects it. An invalid port, an empty host, or a bad percent escape does this.
3. The old client closes. The factory throws. The stale entry stays.
4. The operator sees the failure. The operator sends a PATCH back to A.
5. `clientFor` finds the hash of A. It gives the closed client back.

Each later cycle of that app then fails on a closed client. Only a restart of the monitor clears
this state. The old form, `computeIfAbsent`, left no entry after a failure, thus this defect is
new in this pull request.

My proof (a scratch test in my clone, now deleted): a factory that throws for the second string
only. The first call built the client. The second call closed it and threw. The third call, with
the first string again, called no factory and gave the same closed client back.

**The fix.** Build the client before any change of the map:

```kotlin
internal fun clientFor(appId: Long, connectionString: String): MongoClient {
    val hash = sha256Hex(connectionString)
    val cached = clients[appId]
    if (cached != null && cached.connectionStringHash == hash) return cached.client
    val fresh = clientFactory(connectionString)
    clients.put(appId, CachedClient(hash, fresh))?.client?.close()
    return fresh
}
```

A failed build then keeps the old client open and usable. The `activePolls` guard of
`PollScheduler` still gives one call for each app id, thus `put` is safe here. Add one unit test
of this path. The test asserts that the third call gives an open client.

## MAJOR 2. The log sentinel reads no exception of a log event

`monitor/backend/src/test/kotlin/octometer/monitor/EndToEndContainerTest.kt:413` to `:422`.

`assertNoSensitiveText` reads `event.formattedMessage` only. A connection string inside an
attached exception passes the check.

My break-and-restore proof, in `PollScheduler.runPollCycle`
(`monitor/backend/src/main/kotlin/octometer/monitor/poll/PollScheduler.kt:231`):

- With `log.info("REVIEW PROBE A. {}", connectionString)` the class **failed**, at
  `EndToEndContainerTest.kt:417`. The assertion target works for a message.
- With `log.warn("REVIEW PROBE B.", RuntimeException(connectionString))` the class **passed**.
- I restored the file. The full suite then passed.

The logback `PatternLayout` prints the stack trace of an event even with the pattern `%msg%n`,
thus such a line reaches the console. BLOCKER 1 of the security review of pull request #160 was
an exception message with a user name and a password. This sentinel cannot catch a return of
that defect.

**The fix.** Read `event.throwableProxy` too. Check its `className`, its `message`, each
`cause`, and each `suppressed` entry. Assert the same markers on that text. Add the host and the
port of each container to the marker list. The test URI holds no user name and no password, thus
a match of the full string alone is a weak check.

## MINOR 1. The new reflection test is much smaller than its name

`monitor/backend/src/test/kotlin/octometer/monitor/mongo/MongoAppReaderUnitTest.kt:528` to `:537`,
and the helper at `:540` to `:559`.

The test gives a Mockito double as the client, thus the real `MongoClient` never enters the map.
It reads the declared fields of the reader, and one level of the map values. It compares only a
`String` value. The marker sits in the host, and not in a user name or a password. The test
therefore proves that `CachedClient` holds no string field of the URI. It proves nothing about
an object that the map reaches.

**The fix.** Give the test a name that states that smaller claim. Or use the real factory with a
loopback URI, a marker user name, and a marker password, then walk the graph as my probe did.
Then state the known driver fields in the comment.

## MINOR 2. The close after a PATCH is lazy

`monitor/backend/src/main/kotlin/octometer/monitor/mongo/MongoAppReader.kt:267`.

D10 says that the monitor closes the client "after a PATCH of the connection string". The close
now waits for the next cycle of that app. The old client keeps an authenticated pool to the old
deployment until then, with `maxIdleTimeMS=120000`. A rotated password therefore stays in use
for up to one poll interval. The delete path is better: the scheduler calls `closeClient` at the
next tick.

**The fix.** Name this window in the issue of the backoff work (#28). Or call `closeClient(appId)`
from the PATCH path, under a guard against a poll in flight.

## MINOR 3. The pull request text names a review thread that GitHub does not hold

The summary says "A review correction of this fix (thread on this pull request)". The pull
request holds no comment and no review today. I read the issue comments, the pull request
comments, and the review comments with the GitHub API. Please name the real source of that
correction.

## MINOR 4. About 18 new comment sentences hold more than 25 words

Rule 8 of the backend brief sets 25 words. The longest new sentence holds 41 words, at
`EndToEndContainerTest.kt:77` to `:81`. The class comment of that file, lines 40 to 90, holds
seven more. The other long sentences sit in these comment blocks:
`EndToEndContainerTest.kt:127, 268, 287`, `MongoAppReader.kt:100, 112, 244`,
`MongoAppReaderUnitTest.kt:463, 521`, and `monitor/backend/build.gradle.kts:61`.
`EndToEndContainerTest.kt:86` also uses two `-ing` forms as a verb ("failing", "passing").
Rule 7 of the language policy forbids that form.

- **The PATCH, the good path.** A changed string closes the old client and builds one fresh
  client. A repeat of the same string builds no client. The two new unit tests prove both, and I
  ran them.
- **The secret.** `AppRegistryService.kt:177` writes the new string with
  `SecretStore.put`. `SecretStore.kt:57` replaces the map entry, and `writeAll` writes the whole
  map to a temporary file, then it moves that file over `apps.json`. The old string leaves the
  file. D11 accepts the plain form of that file, and it names BitLocker for the disk.
- **No secret in a message.** `ConnectionStringValidator` has one fixed sentence for each rule.
  `defaultMongoClient` runs inside `withMongoFailure`. `logback.xml` holds `org.mongodb` at WARN.
- **The ingest side of the test.** The test gives `IngestSettings(false, ...)`, thus
  `recordAnonymousClicks` is `false`. The demo app gives the same value
  (`tools/demo-app/.../Application.kt:41`). `IngestRoute.kt:304` drops each request with no user
  id, and it answers 204. A session start with no user id therefore stores nothing, and it counts
  nothing. This is the D19 rule of production. The test sends a cookie on each request, thus it
  never walks the anonymous path. The pilot app turns that path on, and no test of this pull
  request covers it.
- **The test output.** The class prints its own run time only. The failure text of `waitForRow`
  prints an `AppRow`. That row holds no user id, no session id, and no connection string.
- **The temporary folders.** The class nests its data folder, its secrets folder, and its backup
  folder inside one registered root. `tearDown` removes the root. My machine holds no
  `octometer-e2e-container-test-*` folder after my runs.
- **The containers.** `docker ps -a` shows no container of this class and no reaper container
  after my runs. The Testcontainers reaper did its work.
- **The diff.** It holds no secret, no database file, and no build output. The pull request text,
  the issue comment, and the diff hold no URI with a credential, and no name of the pilot app.
  The `Secret scan` job of the run 35756776353 passed.
- **The CI guard.** The listener of `monitor/backend/build.gradle.kts:159` reads the class name
  suffix `ContainerTest`, thus it already covers the new class. A skip on the Ubuntu job fails
  the build.

## Summary

I found 2 MAJOR findings and 4 MINOR findings. I found no BLOCKER. The hash form of the client
cache is the right choice, and I ask you to keep it. MAJOR 1 is a new defect of `clientFor`, and
a restart of the monitor is its only cure today. MAJOR 2 is a hole in the new sentinel, on the
exact path of BLOCKER 1 of pull request #160. Please fix both, then I will review again.

xamcross:
**Review (Kotlin, Ktor, and MongoDB engineer): APPROVE**

I made my own clone. I ran `gh pr checkout 185` in it. I ran each command below myself, with
JDK 21 and Docker. I do not trust the pull request text; each result below is my own evidence.

## The acceptance criteria of issue #21

### 1. The totals equal the sent clicks, users, and sessions — PASS, with one open gap

`EndToEndContainerTest.kt:218-220` asserts the three totals against the named constants.

- The users total equals `SENT_USERS` (3). PASS.
- The sessions total equals `SENT_SESSIONS` (4). PASS.
- The clicks total equals `EXPECTED_FIRST_CLICKS`, and that value is `SENT_CLICKS + SENT_SESSIONS`
  (16, not 12).

The reason is real, and the class comment names it. `EventStore.kt:17-21` inserts no `kind`
column, so each row keeps the column default 0. `AppTotalsRoute.kt:39` counts a row with
`kind = 0`. One `octo:session-start` row of each session therefore counts as a click today.
Contract rule C38 and design decision D41 say that this element is not a click. The open issue
#110 owns that gap. The constants of the test state today's behaviour correctly. See MINOR 1.

**How the test counts a session.** `UNIQUE_SESSIONS_SQL` (`AppTotalsRoute.kt:44`) has no `kind`
filter. A session start alone is a visit (design decision D44). The test gives three clicks to
each session, so it never exercises that rule. See MINOR 2.

**The ingest path against the demo app's path.** The test starts `octometerIngestRoute` with
`IngestSettings(false, PathPatternMatcher.of(listOf("/")))` and `MongoEventLogStore(database)`
(`EndToEndContainerTest.kt:279-287`). The demo app uses the same two values
(`tools/demo-app/src/main/kotlin/octometer/demo/Application.kt:35-40` and `:151`). Three
differences stay:

- The demo app sets `requestReadTimeoutSeconds = 10` on Netty (`Application.kt:63`). The test
  does not.
- The demo app sets a socket read timeout of 5 s on its MongoDB client (`Application.kt:76`).
  The test does not.
- `demoUserId` checks the length and each control character of the cookie
  (`DemoUserIdResolver.kt:20-25`). The test resolver returns the raw cookie value
  (`EndToEndContainerTest.kt:285`).

No difference changes the event path. The tracker JavaScript stays outside the test; the test
posts the JSON body itself. That matches decision 2 of the design brief.

**Contract rule C38.** `sendSessionBatches` posts the session-start body first, then the click
batch (`EndToEndContainerTest.kt:310-312`). PASS.

**The wait.** `waitForRow` polls `GET /api/apps` with a 30 s bound and a 300 ms step
(`EndToEndContainerTest.kt:399-410`). There is no fixed sleep. PASS.

**The real services.** The test calls the real `module(config)`. That call opens the real reader
and the real scheduler (`Application.kt:88`, `MonitorServices.kt:126-135`). The config sets
`pollIntervalSeconds = 1` and `settleLagSeconds = 1` (`EndToEndContainerTest.kt:189-191`). The
test registers the app through `POST /api/apps` (`:345-358`). PASS.

### 2. After a PATCH of the connection string the next cycle uses the new string — PASS

The test writes a second batch into a second container, then it sends the PATCH, then it waits
for the grown totals (`EndToEndContainerTest.kt:228-245`).

The totals of the first source stay in SQLite. `AppRegistryService.updateApp:142-183` writes only
the name and the secret. It touches no `event` row and it removes no app row. The test asserts
the sum of both sources (`EXPECTED_FINAL_*`, `:132-135`). PASS.

I proved that this assertion can fail. I keyed `clientFor` on the app id alone
(`computeIfAbsent`), and the test failed at `EndToEndContainerTest.kt:409` with the timeout
message. I restored the code.

### 3. `status: OK` and `lastSuccessAt` after each cycle — PASS

`EndToEndContainerTest.kt:222-223` and `:246-247` assert both fields, after the first cycle and
after the PATCH.

### 4. The reader fix of `MongoAppReader.kt` — PASS

- The map `clients` holds `Long -> CachedClient` (`:174`). The value holds the SHA-256 hex and
  the client (`:111`).
- A different hash closes the old client, then it builds a fresh one (`:269-276`).
- The connection string reaches no field. The unit test at `MongoAppReaderUnitTest.kt:527-538`
  reads each declared field of the reader, and each field of each map value, through reflection.
  I stored the raw string in `CachedClient` for one run, and that test failed. I restored the
  code. Design decision D11 holds.
- `MongoAppReaderUnitTest.kt:497-519` proves two factory calls and one `close()` for two strings,
  and no third client for the repeat. I broke the cache two ways, and I saw a failure each time:
  the app-id key gave one failure, and a cache that always rebuilds gave two failures.

**The hash form is acceptable.** The reader needs an equality test only, and a hash gives that
test. SHA-256 is a one-way function. A person who reads a heap dump gets no credential back,
because the password of a real URI has high entropy. The hash is not the connection string, so
design decision D11 stays true.

**`toString()` of the entry is not safe today.** See MINOR 3.

### 5. The class runs in less than 60 s — PASS

I ran the class three times in sequence. The JUnit XML `time` values were 12.245 s, 12.379 s, and
12.383 s. The slowest run was 12.4 s. The log line of the class printed 7.0 s, 7.0 s, and 7.2 s.
Both numbers stay under 60 s. The gap between the two numbers has a cause; see MINOR 4.

### 6. The test runs in the `jvm` CI job — PASS

The class name ends with `ContainerTest`, so the guard at `monitor/backend/build.gradle.kts:159`
covers it. The job "JVM modules" sets the variable, and the build passed:

```
JVM modules	Build each JVM module	2026-09-22T16:51:07.5152888Z   OCTOMETER_REQUIRE_DOCKER: true
JVM modules	Build each JVM module	2026-09-22T16:52:37.3556616Z > Task :monitor:backend:test
JVM modules	Build each JVM module	2026-09-22T16:53:52.3151435Z BUILD SUCCESSFUL in 2m 44s
```

The log holds no line that names the class and its skip count. The Docker guard writes no such
line; only the `OCTOMETER_REQUIRE_DIST` guard writes one (`build.gradle.kts:199`). The evidence
is therefore indirect: the guard throws for any `*ContainerTest` suite with one skip or more, and
the build passed. I proved that the guard works. My first local run met a short Docker outage,
and the build failed with this exact message:

```
The class "octometer.monitor.EndToEndContainerTest" of monitor:backend skipped 1 test(s) on CI.
```

The Windows job sets no such variable. It ran the same task and it passed (job 106844470341). The
class skips there by design. See MINOR 5.

### 7. The new test dependency on `kit/jvm-ktor` — PASS

`./gradlew :monitor:backend:dependencies --configuration runtimeClasspath` prints no `jvm-ktor`
entry and no `mockito` entry; my grep count of both names was 0. `kit/jvm-ktor/build.gradle.kts`
holds no npm task and no Node task. It depends on `kit:jvm-core` only, and that module is already
a test dependency of the backend.

### 8. `./gradlew :monitor:backend:test` passes — PASS, with a count correction

My full run gave **426 tests, 0 failures, 2 ignored**. The two ignored tests are `FrontendCspTest`
and `WindowsStartScriptTest`; each one needs the `installDist` artifact, which my run did not
build. The pull request text says 423 tests. That number is stale. The comment on issue #21 says
426, and that number is right. See MINOR 6.

**Each new test can fail.** I proved it, one break at a time:

| Test | The break | The result |
|---|---|---|
| `EndToEndContainerTest` | key on the app id alone | FAILED at line 409 |
| `clientFor closes the old client ...` | key on the app id alone | FAILED |
| `clientFor gives the same client back ...` | always build a fresh client | FAILED |
| `no field of the reader ... holds the connection string` | keep the raw string | FAILED |

### 9. The language of each new comment — CHANGES NEEDED (MINOR)

See MINOR 7.

### 10. The comment on issue #21 — PASS

The comment exists (id 5780538036). It records the CI run URL of the `jvm` job and the test time.
The checkbox in the pull request text is still empty; please tick it.

## The findings

All findings are MINOR. No BLOCKER and no MAJOR remains open, thus I approve.

**MINOR 1. Name issue #110 in the class comment.** `EndToEndContainerTest.kt:64-75` says "A later
issue that adds the `kind` mark ... changes this total". It names no issue. Fix: write "issue
#110" in that sentence, and add the same reference beside `EXPECTED_FIRST_CLICKS` at line 119.
Add one line to the pull request text too. The ticked criterion reads "The totals equal the sent
clicks". The assertion adds the session-start rows on top of that count.

**MINOR 2. Prove the visit rule.** `EndToEndContainerTest.kt:200-205` gives three clicks to each
session, so the sessions total never differs from the count of sessions with a click. Fix: send
one extra session with the session-start request only, then raise `SENT_SESSIONS` by 1 and
`EXPECTED_FIRST_CLICKS` by 1. The test then proves that `UNIQUE_SESSIONS_SQL` uses no `kind`
filter (design decision D44).

**MINOR 3. `CachedClient` is a data class, so `toString()` prints the hash and the client.**
`MongoAppReader.kt:111`. The Kotlin compiler writes `toString()` for a data class from each
property. A later edit that logs the map, or that prints one entry, therefore writes a 64
character hash of a credential into a log file. No code does that today, so the risk stays
latent. The class also needs no `equals` and no `hashCode`. Fix: drop the `data` keyword, or add
`override fun toString() = "CachedClient"`.

**MINOR 4. The self-measured time excludes the container start.**
`EndToEndContainerTest.kt:154` reads `System.nanoTime()` in the instance initializer. The
Testcontainers extension starts the two static containers before that, in `beforeAll`. My runs
show the gap: 12.4 s in the JUnit XML against 7.2 s in the log line. The assertion at `:264-267`
therefore does not bound the real class time. Fix: read the start time in a `@BeforeAll` function
of the companion object, before the container start, or drop the assertion and keep the print.

**MINOR 5. The CI log proves the skip count of no container class.**
`monitor/backend/build.gradle.kts:148-170`. The `OCTOMETER_REQUIRE_DIST` guard at `:190-205`
already writes one `logger.lifecycle` line for each class that it guards, and its own comment
gives the reason: "the default Gradle console prints no per-class line on a pass". Fix: add the
same one line to the Docker guard, for each suite whose name ends with `ContainerTest`. The CI
log then names the class, the test count, and the skip count. This change touches a shared file,
so the maintainer can also move it to its own issue.

**MINOR 6. The test count in the pull request text is stale.** The text says 423 tests. My run
gives 426. Fix: write 426 in the pull request text, the same number as the comment on issue #21.

**MINOR 7. Nine new comment sentences break the language policy.** The limit is 25 words for a
description. Each sentence below is longer:

- `MongoAppReader.kt:106-109` (33 words). It also joins three clauses with "and".
- `MongoAppReader.kt:114-116` (27 words).
- `MongoAppReader.kt:248-251` (37 words).
- `MongoAppReader.kt:253-256` (27 words).
- `MongoAppReader.kt:256-260` (46 words).
- `MongoAppReader.kt:262-265` (38 words).
- `EndToEndContainerTest.kt:68-75` (27 and 30 words).
- `EndToEndContainerTest.kt:77-89` (41, 39, and 40 words).
- `MongoAppReaderUnitTest.kt:521-526` (37 words).
- `monitor/backend/build.gradle.kts:64-68` (27 words).

`EndToEndContainerTest.kt:86-88` also uses two `-ing` forms as a verb: "with the assertion ...
failing with a clear message" and "(the same assertion passing)". Fix: cut each long sentence
into two. Write "the assertion failed with a clear message" and "the same assertion passed".

**MINOR 8. Mockito is not necessary here.** `monitor/backend/build.gradle.kts:52` and
`MongoAppReaderUnitTest.kt:477`, `:499-500`, `:531`. The catalog already holds `mockito-core`
5.23.0, and `kit/jvm-mongo` already uses it, so the version rule is correct. The dependency is
still avoidable. `com.mongodb.kotlin.client.coroutine.MongoClient` is final, but its constructor
is public and it takes `com.mongodb.reactivestreams.client.MongoClient`, an interface that the
runtime classpath already holds. Kotlin interface delegation gives a double in five lines:

```kotlin
// The argument below is a loopback URI of the placeholder form mongodb://<host>:<port>.
val real = com.mongodb.reactivestreams.client.MongoClients.create(loopbackUri)
var closeCount = 0
val counting = object : com.mongodb.reactivestreams.client.MongoClient by real {
    override fun close() { closeCount += 1; real.close() }
}
val client = MongoClient(counting)
```

Fix: use that form, then remove the line at `build.gradle.kts:52`. A working test with a new
dependency blocks nothing, so the maintainer can keep the present form.

**MINOR 9. `internal` on `clientFor` is the smallest seam.** No change needed. `pollOnce` needs a
real `MongoDatabase` for `readPage`, thus a test cannot reach the cache through `pollOnce` with a
double. The file already marks `runCycle`, `idFilter`, and `boundObjectId` as `internal` for the
same purpose. The choice matches the practice of the module.

**MINOR 10. A PATCH keeps the cursor of the old source.** `MongoAppReader.kt:267` and
`AppRegistryService.kt:175-181`. The app row keeps its cursor after a PATCH. The next cycle reads
`{_id: {$gt: cursor}}` against the new deployment. Every event of the new deployment with a
smaller `_id` therefore stays unread, and no log line names that loss. The test passes because
the driver of one JVM builds each `_id`, and those values rise in order across both containers.
Design decision D10 names no cursor reset. Issue #21 asks only that the next cycle uses the new
string. This gap therefore stays outside the scope of this pull request. Fix: add one comment line to
`EndToEndContainerTest.kt` that states the order rule of the two containers. Then file one issue
for the cursor after a PATCH.

## Scope, secrets, and the temporary files

- The diff holds four files. It adds no database file and no build output.
- The test builds each URI from a live container (`EndToEndContainerTest.kt:174-179`). No text of
  the diff holds a credential. The "Secret scan" job passed.
- The test root uses `registerTempRoot` and `testDataDir`, and `tearDown` removes it
  (`:152-159`). That follows lesson 7 of the brief.
- Both commit bodies use short sentences and the active voice.

xamcross:
This comment answers the security review and the Kotlin review of commit `cf29561`.
Commit `b24f33e` and commit `a035b1e` hold the fix of each finding below, or the reason
that a finding stays open for the maintainer.

| Reviewer | Finding | Commit | File and line | Proof |
| Security | MAJOR 1: a failed client build left a closed client in the cache | `b24f33e`, `a035b1e` | `MongoAppReader.kt:278-285` (fix); `MongoAppReaderUnitTest.kt:529` (test) | A break to the old close-before-build form failed the new test. Restored, the test passed. |
| Security | MAJOR 2: the log sentinel read no exception of a log event | `a035b1e` | `EndToEndContainerTest.kt:461-472` | A temporary `log.warn` with the connection string inside a `RuntimeException`, in `PollScheduler.kt`, failed the class. Restored, the class passed. |
| Security | MINOR 1: the reflection test read one level of the map, and marked the host only | `a035b1e` | `MongoAppReaderUnitTest.kt:592` | The test now walks the reader, each cached entry, and the client's own settings, with a marker user name, password, host, and port. |
| Security | MINOR 2: the close after a PATCH is lazy (it waits for the next poll cycle) | none | `MongoAppReader.kt:267` | Open. Design decision D10 names no deadline for the close. The maintainer can file this against issue #28 (the backoff work) or a new issue. |
| Security | MINOR 3: the pull request text named a review thread that GitHub holds no record of | `a035b1e` | pull request text, Summary section | Fixed. The text now names the real source: BLOCKER 1 of the security review of issue #16. |
| Security | MINOR 4: about 18 new comment sentences held more than 25 words | `a035b1e` | `EndToEndContainerTest.kt`, `MongoAppReader.kt` (doc comments) | Fixed. Each rewritten sentence now holds 25 words or fewer. |
| Kotlin | MINOR 1: the class comment named no issue for the click-count gap | `a035b1e` | `EndToEndContainerTest.kt:67`, `:124` | Fixed. Both lines now name issue #110. |
| Kotlin | MINOR 2: the sessions total never proves the visit rule (no click of its own) | none | `EndToEndContainerTest.kt:200-205` | Open. Out of scope of this pull request, by the maintainer's decision. The maintainer files a new issue. |
| Kotlin | MINOR 3: `CachedClient` is a data class, so `toString()` prints the hash and the client | `b24f33e`, `a035b1e` | `MongoAppReader.kt:116-118` (fix); `MongoAppReaderUnitTest.kt:563` (test) | The class now prints the fixed text `"CachedClient"`. A test asserts it holds neither the hash nor the client text. |
| Kotlin | MINOR 4: the self-measured class time excludes the container start | `a035b1e` | `EndToEndContainerTest.kt:101`, `:166` | Fixed. `classLoadNanos` is a companion object field; it initializes at class load, before the `@Container` fields start. |
| Kotlin | MINOR 5: no CI log line proves the skip count of a container-test class | none | `monitor/backend/build.gradle.kts:148-170` | Open. A shared-file change; the review itself says the maintainer can move it to its own issue. |
| Kotlin | MINOR 6: the pull request text held a stale test count (423) | already correct | pull request text | The text already held 426 when this correction began. This correction raised the true count to 428 with two new unit tests, and updated the text to match. |
| Kotlin | MINOR 7: nine new comment sentences held more than 25 words, two used an `-ing` form as a verb | `b24f33e`, `a035b1e` | `MongoAppReader.kt`, `EndToEndContainerTest.kt`, `MongoAppReaderUnitTest.kt`, `build.gradle.kts` (doc and line comments) | Fixed. Each rewritten sentence now holds 25 words or fewer, in the active voice. |
| Kotlin | MINOR 8: Mockito is not necessary for the `MongoClient` double | `b24f33e`, `a035b1e` | `build.gradle.kts:41-52` (dependency removed); `MongoAppReaderUnitTest.kt` (`StubMongoClient`) | Fixed. The double now uses the driver's public constructor, through Kotlin interface delegation over a real reactive-streams client. |
| Kotlin | MINOR 9: `internal` on `clientFor` is the smallest seam | no change needed | `MongoAppReader.kt:278` | The reviewer asked for no change. `clientFor` stays `internal`, the same rule as `runCycle`, `idFilter`, and `boundObjectId`. |
| Kotlin | MINOR 10: a PATCH keeps the cursor of the old source | none | `MongoAppReader.kt:267`, `AppRegistryService.kt:175-181` | Open. Out of scope of this pull request, by the maintainer's decision. The maintainer files a new issue. |

Two findings stay open by the maintainer's own decision: Kotlin MINOR 2 and Kotlin MINOR 10.
The maintainer files each one as its own issue. Security MINOR 2 and Kotlin MINOR 5 also
stay open; each review named a reason of its own to leave the choice to the maintainer.

xamcross:
Follow-up (coordinator note, after the findings comment above): the two new marker
identifiers named `password` beside a string literal, so gitleaks' `generic-api-key`
rule would have matched the squash commit on `main`, past the branch-only
`.gitleaksignore` entries.

| Reviewer | Finding | Commit | File and line | Proof |
| Coordinator follow-up | `markerPassword` would match `generic-api-key` on the post-squash commit of `main`; the `.gitleaksignore` entries name branch commits only | `36371f7` | `EndToEndContainerTest.kt:280` (was), `MongoAppReaderUnitTest.kt:594` (was) | Renamed both to `markerSecondValue` (the form of #160), same values. No other new identifier named `password`, `secret`, `token`, or `apikey` beside a string literal. Both test classes ran once (31 tests, 1 test, 0 failures). `.gitleaksignore` gained one line: each entry names a branch commit and dies at the squash merge (issue #171 removes it after). Secret scan passed on commit `36371f7`. |

xamcross:

I made a second clone. I ran `gh pr checkout 185` in it, on the head `36371f7`. I ran each
check myself. I restored each file after each experiment, and the tree is clean now.

## The two MAJOR findings

**MAJOR 1 — fixed.** `MongoAppReader.kt:278` to `:285` now builds the client first. It then
puts it in the map. It closes the old client last. My own scratch test (a factory that throws
for the second string, now deleted) gave this result:

- The second call threw `IllegalArgumentException`.
- The third call, with the first string, gave the same client object back.
- The factory count stayed at 2, thus the third call built nothing.
- The reused client was **open**. `getDatabase(...)` worked on it.

Before the fix, that third call gave a closed client. The new unit test
`a failed client build leaves the old client open, and a later working string closes it`
covers the same path, and it also proves the close of the fourth call.

**MAJOR 2 — fixed.** `EndToEndContainerTest.kt:465` to `:476` now reads `event.throwableProxy`,
its `className`, its `message`, and each `cause`. My break: a `log.warn` with the connection
string inside an attached `RuntimeException`, in `PollScheduler.kt:231`. The class **failed**,
at `EndToEndContainerTest.kt:480`. The same break passed on the old code. I restored the file.

The new marker list is stronger than my request. It holds the host and the port of each
container, in the plain `host:port` form. It also holds a marker user name and a marker second
value from a short failed-authentication probe against the first container. That probe gives
the sentinel of design decision D15 a real credential to search for.

## The four MINOR findings

| --- | --- | --- |
| MINOR 1. The reflection test read one level of the map, with a double, and marked the host only | Fixed | `MongoAppReaderUnitTest.kt:592` to `:611`, and the walk at `:640`. It reads each `String` and each `char[]` of the reader, of the entry, and of the driver settings. It marks the user name, the second value, the host, and the port. The comment at `:585` states the reason for the one combined marker. My own probe found the same split of the driver fields. |
| MINOR 2. The close after a PATCH waits for the next poll cycle | Open | `MongoAppReader.kt:278`. The maintainer keeps it open, and files it against issue #28 or a new issue. I accept that. D10 names no deadline for the close. |
| MINOR 3. The text named a review thread that GitHub holds no record of | Fixed | The Summary now names BLOCKER 1 of the security review of issue #16. |
| MINOR 4. About 18 new comment sentences held more than 25 words | Partly fixed | About 10 stay. The longest holds 40 words, at `MongoAppReaderUnitTest.kt:585`. The others sit at `MongoAppReader.kt:263`, `EndToEndContainerTest.kt:40, 137, 163, 183`, and `MongoAppReaderUnitTest.kt:466, 494, 578, 626`. The response says "Each rewritten sentence now holds 25 words or fewer". That claim is too wide. No `-ing` form works as a noun or as a verb now. |

## The identifier question

**No.** No identifier with "password", "secret", "token", or "key" sits beside a string literal
in the two test files. I read each match of those four words. Each one sits in a comment, in a
test name, or in an assertion message. One literal value still holds the text "password", at
`MongoAppReaderUnitTest.kt:425`, and its identifier is `markerSecondValue`. The gitleaks rule
`generic-api-key` reads the identifier, thus the `Secret scan` job passes.

## MINOR 5 (new). The sentinel reads no suppressed exception

`monitor/backend/src/test/kotlin/octometer/monitor/EndToEndContainerTest.kt:469` to `:474`.

The loop follows `cause` only. It skips `suppressedExceptions` of the throwable proxy. Logback
prints a suppressed exception in the console too. A marker in that text still passes the check
today. **The fix.** Walk `throwableProxy.suppressed` in the same loop. Add each entry to a work
list, beside the `cause`.

## What else I checked

- `./gradlew :monitor:backend:test --rerun` gave 428 tests, 0 failures, 0 errors, and 2 skipped.
  The two skips are `FrontendCspTest` and `WindowsStartScriptTest`, the known pair. I used
  `--rerun`, because the shared build cache first gave a cached pass.
- `EndToEndContainerTest`: 1 test, 0 skipped, 10.9 s in my run. The budget of 60 s holds.
- `.gitleaksignore` holds three fingerprints. Each one names a commit of this branch. Each value
  is a fake marker in a test file. No entry hides a real credential. The comment names issue
  #171 for the sweep after the merge.
- `CachedClient` is no longer a data class, and it prints one fixed text. The Mockito dependency
  is gone. The stub opens no socket at construction.
- No container and no reaper stayed after my runs. No temporary folder of the class stayed.
- The pull request text, the issue comment, and the diff hold no credential text and no name of
  the pilot app. Each CI check of the head is green.

## Summary

Both MAJOR findings are closed. I found one new MINOR finding, and MINOR 2 and MINOR 4 stay
open at the MINOR level. No open finding blocks this pull request. I approve it.

xamcross:
**Review (second review, Kotlin, Ktor, and MongoDB engineer): APPROVE**

I made a second clone. I ran `gh pr checkout 185` in it, at the head `36371f7`. I ran each
command below myself, with JDK 21 and Docker.

## My findings of round 1

|---|---|---|---|
| MINOR 1 | The comment names no issue for the `kind` gap | FIXED | `EndToEndContainerTest.kt:67` and `:73` name issue #110. Line `:124` names it beside `EXPECTED_FIRST_CLICKS`. |
| MINOR 3 | `CachedClient` is a data class | FIXED | `MongoAppReader.kt:116-118` is a plain class with `override fun toString() = "CachedClient"`. The test at `MongoAppReaderUnitTest.kt:565-577` asserts the fixed text, no hash, and no client text. I made the class a data class again, and that test failed. I restored the code. |
| MINOR 4 | The timer starts after the container start | FIXED | `classLoadNanos` is a companion object field (`EndToEndContainerTest.kt:101`), before the two `@Container` fields. `startNanos` reads it (`:166`). My run proves the effect: the JUnit XML gives 12.043 s, and the log line now gives 12.0 s. Round 1 gave 12.4 s against 7.2 s. |
| MINOR 6 | The test count was stale | FIXED | The pull request text says 428 tests. My full run gives 428. |
| MINOR 7 | Comment sentences longer than 25 words | PARTLY FIXED | See the note below. |
| MINOR 8 | Mockito is not necessary | FIXED | `build.gradle.kts` holds no `mockito` line. `MongoAppReaderUnitTest.kt:713-726` holds `StubMongoClient`, with Kotlin delegation (`by`) over the reactive-streams interface, and the public `MongoClient` constructor. My grep of `:monitor:backend:dependencies --configuration testRuntimeClasspath` counts 0 `mockito` entries. |
| MINOR 9 | `internal` on `clientFor` | NO CHANGE NEEDED | `MongoAppReader.kt:278` keeps `internal`. That matches my round 1 judgement. |

## The security MAJOR 1 fix in `clientFor`

`MongoAppReader.kt:278-285` now reads the cache, builds the fresh client, puts it, and closes the
old one last. A failed build therefore leaves the old, working client in the map.

I broke the cache five ways against this new code. Each break gave a failure, and I restored the
file after each one:

| The break | The failing test |
|---|---|
| Close the old client before the build (the old order) | `a failed client build leaves the old client open ...` |
| Key on the app id alone, with no hash compare | `a failed client build ...` and `clientFor closes the old client ...` |
| Always build a fresh client | the two above, plus `clientFor gives the same client back ...` |
| Keep the raw connection string in place of the hash | `no field of the reader, of a cached client entry, or of the client's own settings holds the connection string` |
| Make `CachedClient` a data class again | `toString of the cached client entry holds neither the hash nor the client text` |

The new test for a factory failure (`MongoAppReaderUnitTest.kt:529-556`) is sound. It proves four
facts: the failed build throws, the old client stays open, the next call with the old string
reuses that same client, and a later working string closes it. The `git status` of my clone stayed
clean after each restore.

## The test run

`OCTOMETER_REQUIRE_DOCKER=true ./gradlew :monitor:backend:test` gave **428 tests, 0 failures, 2
skipped**. The two skips are `FrontendCspTest` and `WindowsStartScriptTest`; each one needs the
`installDist` artifact, which my run did not build.

`EndToEndContainerTest` ran in **12.043 s** (the JUnit XML) and printed **12.0 s** (the log line).
Both numbers stay far under the 60 s budget. The two numbers now agree, so MINOR 4 is truly fixed.

## The one finding that stays open

**MINOR 7 (language) — PARTLY FIXED, MINOR, no block.** The response says "Each rewritten sentence
now holds 25 words or fewer". That claim is too strong. I split each added comment line of
`cf29561..36371f7` into sentences. Seven new sentences hold more than 25 words:

- `EndToEndContainerTest.kt` class comment: 27 words ("One HTTP click travels through ...").
- `EndToEndContainerTest.kt:163-166`: 28 words ("classLoadNanos, not a fresh ...").
- `EndToEndContainerTest.kt` marker comment: 27 words and 26 words.
- `MongoAppReaderUnitTest.kt:579-585`: 26 words.
- `MongoAppReaderUnitTest.kt:587-597`: 40 words ("The whole connection string never sits in one
  field, even there: ...").
- `MongoAppReaderUnitTest.kt` walk comment: 33 words ("It stops at the boundary of this module
  ...").

One older sentence also stays: `MongoAppReader.kt` "This method compares [connectionString] by its
SHA-256 hex only ..." holds 27 words. Fix: cut each sentence above at its colon or at its "so".
The `-ing` defect is gone; the file now writes "the assertion failed" and "the assertion passed".

## The parked findings

MINOR 2 (the D44 visit rule) and MINOR 10 (a PATCH keeps the old cursor) stay out of this pull
request, by the maintainer's decision. MINOR 5 (a per-class CI line for the Docker guard) stays
open by my own suggestion. I agree with each of the three.

## The other checks

- `.gitleaksignore` adds three fingerprint lines. It holds no credential text, only a commit hash,
  a path, a rule name, and a line number. The reason and the removal plan (issue #171) sit in the
  comment above each entry.
- The diff adds no database file and no build output.
- The rename to `markerSecondValue` (`36371f7`) follows the rule of pull request #160.

